PR #25, opened 2019-08-12
Port NFS client and server

dpausp:
* tested with 15.09 client and 19.03 server and vice versa

Changelog

* Port NFS client and server role.

### DIFF
--- a/nixos/lib/utils.nix
+++ b/nixos/lib/utils.nix
@@ -21,6 +21,26 @@ rec {
       config.networking.hostName
     ]);
 
+  # Returns service from /etc/nixos/services.json
+  # that matches the given name or null, if nothing matches.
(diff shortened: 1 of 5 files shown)


nixos/lib/utils.nix
```diff
@@ -21,6 +21,22 @@ rec {
       config.networking.hostName
     ]);
 
+  # Returns first service from /etc/nixos/services.json
+  # that matches the given name or null, if nothing matches.
+  # XXX: check for multiple matches here?
```
ctheune: Suggestion for the XXX: stop evaluation and break (causing no further config changes) which will cause no direct service interruption but will alert support personnel at an appropriate time.

ctheune (changes requested):

PR #34, opened 2019-10-07
Port statshost-relay role

dpausp:
Show error if someone tries to enable the statshostproxy role.
It should work but is untested. We don't need it right now.

Case 25409

## Release process

Impact:

Changelog:

* Port statshost-relay role (stats proxy for a resource group).

## Security implications

  - ported from 15.09 without new security implications 
  - Makes nginx listen on port 9090, but firewall blocks access by default. Access must be allowed by custom firewall rules for statshosts that want to fetch metrics.
  - Test checks if nginx listens only on the expected ports.
  - Manual check in dev with netstat.


nixos/roles/statshost/rg-relay.nix
```diff
@@ -1,27 +1,31 @@
...
-    #     access_log /var/log/nginx/loghost_access.log;
-    #     error_log /var/log/nginx/loghost_error.log;
+    services.nginx.enable = true;
+    systemd.tmpfiles.rules = [
+      "d /var/log/nginx 0755 nginx"
+    ];
+    services.nginx.appendHttpConfig = ''
+      server {
+        listen ${config.services.prometheus.listenAddress};
+        access_log /var/log/nginx/loghost_access.log;
```
ctheune: Rename to role!

nixos/roles/statshost/location-relay.nix
```diff
@@ -20,22 +30,20 @@ in
...
-    #     # XXX HTTPS!
-    #     ${fclib.nginxListenOn config "ethfe" port}
+    services.nginx.enable = true;
+    services.nginx.appendHttpConfig = ''
+      server {
+        ${fclib.nginxListenOn "ethfe" port}
 
-    #     access_log /var/log/nginx/statshost_access.log;
-    #     error_log /var/log/nginx/statshost_error.log;
+        access_log /var/log/nginx/statshost_access.log;
```
ctheune: Rename to role!

nixos/roles/statshost/location-relay.nix
```diff
@@ -1,16 +1,26 @@
 # Relay stats of a location via NGINX
```
ctheune: Rename file to statshost/location-proxy.nix

ctheune (changes requested):

PR #35, opened 2019-10-07
Port mysql / percona roles and additional packages

dpausp:
- Percona packages from fc-15.09, but with current percona versions
  for 5.6, 5.7 and 8.0
- mysql package for 5.5
- Use sensu plugin instead of nagios monitoring script
- use non-root user for monitoring
- xtrabackup: 8.0.5 -> 8.0.7
- systemd now always restarts mysql when it dies
- percona 8.0: bind to srv and lo addresses only
- remove stuff for versions before 5.5
- refactor role options / package selection
- remove maintenance timer
  - has never worked on NixOS because of a typo (onCalendar -> OnCalendar)

Case 25409

## Release process

Impact:

Changelog:

## Security implications

  - ported from 15.09 with some improvements
  - no root password in the Nix store
  - 8.0: mysql listens only on lo and srv addresses, older versions don't support that
  - monitoring is done with separate db users that can only access their own database with read permissions from localhost
  - Test checks that exactly the expected ports are opened by mysql and secret files have the correct permissions
  - Manual checks in dev that root pw is not in the nix store and that monitoring users have the correct permissions

### DIFF
--- a/nixos/rename.nix
+++ b/nixos/rename.nix
@@ -4,5 +4,7 @@
     # old -> new
     # redis and redis4 roles do the same
     (mkRenamedOptionModule [ "flyingcircus" "roles" "redis4" ] [ "flyingcircus" "roles" "redis" ])
+    (mkRemovedOptionModule [ "flyingcircus" "services" "percona" "rootPassword" ] "Change the root password via MySQL and modify secret files")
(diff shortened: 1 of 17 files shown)


ctheune (changes requested):
Sieht gut aus. Bitte noch security requirements durchgehen.

PR #37, opened 2019-10-09
Docker role

dpausp:

## Release process

Impact:

Changelog:

* Add docker role.

## Security implications

We just activate standard docker support on NixOS and don't change the settings.

  - Only users in the docker group are allowed to use docker commands.
  - We rely on upstream, we run a NixOS test that checks if docker works and if permissions are correct.


ctheune:
A little bit of testing would be nice :)

PR #38, opened 2019-10-22
Port antivirus role

dpausp:
* clamav package and service from upstream
* clamd listens on srv and lo addresses

Case 119631
Case 115501

## Release process

Impact:

Changelog:

* Port antivirus (clamav) role. Clamd listens on srv interfaces. (#119631, #115501).

## Security implications

- uses upstream service but disables private network because we want to listen to srv interfaces; disables private temp dir because we want to allow scanning files in /tmp

  - clamd listens on srv and lo interfaces
  - clamd find virus signatures
  - automatic test checks if correct listen addresses are set in the config
  - manual checks in dev if clamd listens to the correct interfaces and recognizes the EICAR virus test signature

### DIFF
--- a/nixos/roles/antivirus.nix
+++ b/nixos/roles/antivirus.nix
@@ -1,0 +1,66 @@
+{ config, lib, pkgs, ... }:
+
+with builtins;
+
+let
+  fclib = config.fclib;
+  listenAddresses =
+    fclib.listenAddresses "lo" ++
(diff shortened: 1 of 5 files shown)


ctheune (changes requested):
Fine so far, see #119631 for small changes.

dpausp:
small changes done...

PR #47, opened 2020-01-09
Port loghost, graylog and elasticsearch roles.

dpausp:
* graylog frontend uses SSL with automated letsencrypt certs by default
* logstash plugin not available anymore, use beats input instead
* uses upstream graylog 3 and elasticsearch 5 & 6
* upgrade from graylog 2.5 (NixOS 15.09 platform) works
* local JSON config can add/override graylog settings
* better fc-graylog that can be used to interact with the graylog API

bugs id: #119556

## Release process

Impact:

Changelog:

* Add loghost, graylog (3.x) and elasticsearch (5.x, 6.x) roles (#119556).

## Security implications

Improves security compared to the old platform by using a more recent graylog version and removing the admin pw from the nix store.

  - Nginx only listens on fe when public dashboard is enabled, other services listen only on srv/loopback.
  - no admin password in the nix store
  - secrets only accessible by graylog and service users
  - manual checks in dev that the admin pw is not in the nix store and that the services listen on the correct interfaces (compared with old platform)
  - automated tests check permissions of secrets and if elasticsearch and graylog listen on the correct interfaces

### DIFF
--- a/nixos/lib/utils.nix
+++ b/nixos/lib/utils.nix
@@ -9,10 +9,6 @@ in
 with lib;
 rec {
 
-  # get the DN of this node for LDAP logins.
-  getLdapNodeDN = config:
-    "cn=${config.networking.hostName},ou=Nodes,dc=gocept,dc=com";
-
   # Derives a password from host data and a custom prefix
   derivePasswordForHost = prefix:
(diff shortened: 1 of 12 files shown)


zagy:
Reviewed with @dpausp

PR #48, opened 2020-01-15
Reduce I/O contention during garbage collection

ckauhaus:
Switch all VMs to BFQ so that ionice does the right thing.

Use fc-userscan 0.4.4 which manages load:
- Use Rust toolchain from 19.09 since our standard version is too old.
- Add pkgs-19.09 import to our overlay.

Also:
- Decrease fc-userscan verboseness so that errors and warnings can be
  spotted easily.
- Rewrite fc-userscan control script with Python.

Case 119463
Case 115735

## Release process

Impact:

Changelog: Use BFQ for I/O scheduling to reduce system load during OS garbage collection.

## Security implications

Remove potential denial-of-service source.

Tried various I/O intensive operations during nix-collect-garbage: git checkout, Rust compilation, pkg installation.

### DIFF
--- a/nixos/infrastructure/flyingcircus.nix
+++ b/nixos/infrastructure/flyingcircus.nix
@@ -16,6 +16,7 @@ mkIf (cfg.infrastructureModule == "flyingcircus") {
     consoleLogLevel = mkDefault 7;
 
     initrd.kernelModules = [
+      "bfq"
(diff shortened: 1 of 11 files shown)


ctheune (approved):
Reviewed together at my computer. Desired effect is visible. Code looks fine.

PR #49, opened 2020-01-17
Update kernel to 4.19.94

dpausp:
bugs id: #122779

## Release process

Impact:

* [NixOS 19.03] VM will be rebooted.

Changelog:

* Update kernel to 4.19.94 (#122779).

## Security implications

- [x] [Security requirements]
This update is part of our patch effort. New kernels are expected to not increase the attack vector and thus don’t need to pass additional requirements.
  n/a, I have verified that it boots on a dev VM and that tests work


ctheune (approved):
passt

PR #50, opened 2020-01-17
Prometheus: fix missing data after upgrade from 15.09, cleanup

dpausp:
* Change prometheus data dir to /srv/prometheus/metrics.
  The upstream service hard-codes /var/lib as path prefix for the
  prometheus data dir. We use /srv/prometheus on 15.09 and want to keep
  that on 19.03.
* Add remote_write and remote_read to prometheus options, we use that to
  read / write to InfluxDB.
* Remove unused Prometheus 1 service code

bugs id: #115483

## Release process

Impact:

Changelog:

* Prometheus / statshost: fix missing data after upgrade from 15.09. (#115483);

## Security implications

- [x] [Security requirements]
  - same as before; new data dir should only be readable by prometheus service user
  - automatic test checks if permissions are set correctly, manual checks in dev

### DIFF
--- a/nixos/roles/statshost/default.nix
+++ b/nixos/roles/statshost/default.nix
@@ -55,17 +55,17 @@ let
     (proxy: { job_name = proxy.location;
               proxy_url = "http://${proxy.address}:9090"; })
     relayLocationProxies;
+
   relayLocationProxies =
     # We need the FE address, which is not published by directory. I'd think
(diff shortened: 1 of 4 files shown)


ckauhaus (approved):
Code looks good after visual inspection. Didn't try it out. Needs to be tested by the appops.

PR #51, opened 2020-01-21
Avoid logrotate errors on fc-manage runs

dpausp:
* try to restart nginx after log rotation if sending the signal fails
* remove logrotate from multi-user.target which avoids running logrotate
during fc-manage runs

Logrotate produced an error during a fc-manage run which caused the
maintenance task to fail. This happened because nginx was restarted at
the same time and logrotate tried to send a signal to the missing process.
Each one of the changes fixes the problem but it makes sense to include
both to avoid errors in similar situations.

bugs id: #119218

## Release process

Impact:

Changelog:

* Don't trigger logrotate on fc-manage runs. This avoids maintenance errors that are not really related to it. Also, make nginx log rotation more robust (#119218).

## Security implications

  - n/a
  - tested in dev that it rotates nginx log files and restarts nginx if neccessary

### DIFF
--- a/nixos/services/logrotate/default.nix
+++ b/nixos/services/logrotate/default.nix
@@ -72,28 +72,31 @@ in
       # We create one directory for each service user. I decided not to remove
       # old directories as this may be manually placed data that I don't want
       # to delete accidentally.
-      flyingcircus.localConfigDirs = let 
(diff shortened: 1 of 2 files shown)


ckauhaus (approved):
Looks good

PR #54, opened 2020-01-24
Use our pinned nixpkgs version when building releases on Hydra

dpausp:
It's ok to use the current revision of nixpkgs via <nixpkgs> for bootstrapping,
but we don't want to build with that version.
The release function has a nixpkgs arg which is by default our pinned
version, but it is overridden by Hydra because it has the same name.
Renaming fixes the problem.

bugs id: #122759

## Release process

Impact:

Changelog:

## Security implications

n/a, doesn't affect running VMs


ctheune (approved):
This looks fine - we still should consider pinning the outer <nixpkgs> input that we pass in from hydra. Our normal setup fixes this later ... but still.

PR #56, opened 2020-01-27
Statshost: remove unneeded PostgreSQL and InfluxDB labels

dpausp:
bugs id: #121869

## Release process

Impact:

Changelog:

* Statshost: remove unneeded PostgreSQL and InfluxDB labels (#121869).

## Security implications

n/a, just a manual test if the right metrics/labels are matched

### DIFF
--- a/nixos/roles/postgresql.nix
+++ b/nixos/roles/postgresql.nix
@@ -42,6 +42,20 @@ with builtins;
     })
 
     {
+      flyingcircus.roles.statshost.prometheusMetricRelabel = [
+        {
+          source_labels = [ "__name__" "datname" ];
+          regex = "postgresql_.+;(.+)-[a-f0-9]{12}";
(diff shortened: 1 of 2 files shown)


nixos/roles/postgresql.nix
```diff
@@ -42,6 +42,20 @@ with builtins;
     })
 
     {
+      flyingcircus.roles.statshost.prometheusMetricRelabel = [
+        {
+          source_labels = [ "__name__" "datname" ];
+          regex = "postgresql_.+;(.+)-[a-f0-9]{12}";
+          replacement = "$1";
+          target_label = "datname";
```
zagy, line 50: Ah, yes. That's a good idea. This way we fold it together to one label and do not remove the entire metric.

zagy (approved):

PR #61, opened 2020-02-10
Fix rabbitmq settings timer execution and warning

dpausp:
The service had RemainAfterExit=true set which means that it stays
active all the time. Timers don't reactivate active units, so it only
ran once and not hourly as intended.

The executed unit must be specified with extension, for example .service.
Systemd complained about that (no valid unit type) and ignored the line.
The default service name is the same as the timer name, so we don't have
to specify it at all.

bugs id: #123236
bugs id: #123185

## Release process

Impact:

Changelog:

* Fix RabbitMQ settings timer execution and warning (#123236).

## Security implications

  - timer must run to remove the insecure guest user and set up monitoring permissions
  - I have checked that the timer runs periodically

### DIFF
--- a/nixos/roles/rabbitmq.nix
+++ b/nixos/roles/rabbitmq.nix
@@ -101,7 +101,7 @@ with builtins;
         '';
 
       systemd.services.fc-rabbitmq-settings = {
-        description = "Prepare rabbitmq for operation in FC.";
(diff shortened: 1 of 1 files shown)


ckauhaus:
LGTM

PR #63, opened 2020-02-17
MySQL: set root password from password file on startup

dpausp:
We want to be able to change the mysql root password without writing it
to the Nix store. The password from the password file (default location
/etc/local/mysql/mysql.passwd) is set from the pre-start script now.
It also creates /root/.my.cnf. This ensures that the files containing
the password always match the actual password (after a restart).
If the password file is missing on startup, it's generated randomly.

bugs id: #123366

## Release process

Impact:

* [NixOS 19.03] MySQL will be restarted.

Changelog:

* MySQL: set root password on every service startup. The root password can be changed now by editing the password file in /etc/local and restarting mysql. It's also possible to set a password before enabling a MySQL role (#123366).

## Security implications

We add an init file with the root password in /run/mysqld which must be protected like the other secret files. It's not possible to delete it because the file is needed for service restarts done by systemd.

  - secret files only readable for root and service users
  - automated test checks secret file permissions, manual verification of permissions in dev

### DIFF
--- a/nixos/services/percona.nix
+++ b/nixos/services/percona.nix
@@ -8,10 +8,12 @@ let
 
   mysql = cfg.package;
 
+  initFile = "/run/mysqld/init_set_root_password.sql";
+
   pidFile = "${cfg.pidDir}/mysqld.pid";
 
   mysqldOptions =
-    "--user=${cfg.user} --datadir=${cfg.dataDir} --basedir=${mysql}";
+    "--user=${cfg.user} --datadir=${cfg.dataDir} --basedir=${mysql} --init-file=${initFile}";
 
   myCnf = pkgs.writeText "my.cnf"
   ''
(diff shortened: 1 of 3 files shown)


nixos/services/percona.nix
```diff
@@ -173,6 +182,36 @@ in
       unitConfig.RequiresMountsFor = "${cfg.dataDir}";
       preStart =
         ''
```
ckauhaus: I'd rather move preStart into a let expressions. This is rather bulky.

nixos/services/percona.nix
```diff
@@ -31,6 +33,13 @@ let
       "${mysql}/bin/mysqld --initialize-insecure ${mysqldOptions}"
     else
       "${pkgs.perl}/bin/perl ${mysql}/bin/mysql_install_db ${mysqldOptions}";
+
+  setPasswordSql =
+    if (lib.versionAtLeast mysql.mysqlVersion "5.7") then
+      "ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '$(< ${cfg.rootPasswordFile})';"
```
ckauhaus: Why not use `$pw`?

ckauhaus (changes requested):

PR #66, opened 2020-02-18
check_haproxy script, checkEnvPackages option, haproxy test

dpausp:
* add package for check_haproxy script
* new option flyingcircus.services.sensu-client.checkEnvPackages that is used
  here to add the check_haproxy script to the check PATH
  if haproxy is enabled
* add sensu-check-env script to show current bin dir that is included in
  the check PATH to make running checks manually easier
* fix haproxy test, use subtests and include it in default.nix

bugs id: #123360

## Release process

Impact:

Changelog:

* Add check_haproxy script to Sensu check env if haproxy is enabled. (#123360).
  * The command `sensu-check-env` can be used to show executables available in the PATH of Sensu checks.

## Security implications

n/a

### DIFF
--- a/pkgs/fc/check-haproxy/fc/check_haproxy/haproxy.py
+++ b/pkgs/fc/check-haproxy/fc/check_haproxy/haproxy.py
@@ -1,0 +1,178 @@
+import argparse
+import datetime
+import itertools
+import logging
+import nagiosplugin
+import numpy
+import re
+
+
+class HAProxyLog(nagiosplugin.Resource):
+
+    r_logline = re.compile(
+        r'haproxy.*: .* \d+/\d+/\d+/\d+/(\d+) (\d\d\d) .* "\w+ (/\S+) HTTP')
+
+    def __init__(self, logfile, statefile, percentiles, url_filters, exclude_patterns):
+        self.logfile = logfile
+        self.statefile = statefile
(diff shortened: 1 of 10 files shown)


pkgs/fc/check-haproxy/fc/check_haproxy/haproxy.py
```diff
@@ -0,0 +1,181 @@
+# Copyright (c) 2012 gocept gmbh & co. kg
+# See also LICENSE.txt
+
```
ckauhaus: This header can be deleted

pkgs/overlay-python.nix
```diff
@@ -35,4 +35,14 @@ python-self: python-super:
...
   });
 
+  nagiosplugin = python-super.buildPythonPackage rec {
+    pname = "nagiosplugin";
+    version = "1.2.4";
+    src = python-super.fetchPypi {
+      inherit pname version;
+      sha256 = "1fzq6mhwrlz1nbcp8f7sg3rnnaghhb9nd21p0anl6dkga750l0kb";
+    };
+    doCheck = false;  # "cannot determine number of users (who failed)"
```
ckauhaus: Better apply patch to remove the `who` test selectively.

ckauhaus (changes requested):

PR #73, opened 2020-03-17
Port Kibana role, fix es6 license problem and deprecation warning

dpausp:
* Uses an elasticsearch node that is running locally automatically now.
* ES6 uses an unfree license, we have to enable allowUnfree.
* Fix deprecation warning by using the new publish address format.
* Add show config scripts for es and kibana.

bugs id: #124025

## Release process

Impact:

* [NixOS 19.03] Elasticsearch 6 will be restarted.

Changelog:

* Fix Elasticsearch 6 role and avoid deprecation warning by using the new publish address format from ES7 (#123942).
* Port Kibana role (#124025).

## Security implications

  - Kibana server listens only on srv.
  - nonFree license allows us to use ES in the intended way
  - Automatic test confirms that kibana is listening on srv; manual checks on test VM for open ports.
  - https://github.com/elastic/elasticsearch/blob/6.8/licenses/ELASTIC-LICENSE.txt prohibits SAAS offerings where the _main_ motivation is ES. That's not the case here. ES is always only part of a customer application.

### DIFF
--- a/nixos/roles/kibana.nix
+++ b/nixos/roles/kibana.nix
@@ -1,0 +1,88 @@
+{ config, lib, pkgs, ... }:
+
+with builtins;
+
+let
+  cfg = config.flyingcircus.roles.kibana;
+  fclib = config.fclib;
+  esUrlFile = "/etc/local/kibana/elasticSearchUrl";
+  esCfg = config.services.elasticsearch;
+
+  # Determine the Elasticsearch URL, in order:
+  # 1. elasticSearchUrl option
+  # 2. URL from local config
+  # 3. Local Elasticsearch service
+  # 4. no URL, don't activate Kibana
+  elasticSearchUrl =
+    if cfg.elasticSearchUrl == null
+    then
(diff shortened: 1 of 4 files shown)


nixos/roles/kibana.nix
```diff
@@ -0,0 +1,87 @@
...
+    })
+
+    (lib.mkIf cfg.enable {
+      environment.etc."local/kibana/README.txt".text = ''
+        Kibana local configuration
+
+        If no URL has been set and elasticsearch is running on this machine it is used automatically.
+
+        To configure the ElasticSearch kibana connects to, add a file `elasticSearchUrl`
+        here, and put the URL in.
```
zagy, line 74: I suggest changing this to:

```
To configure the ElasticSearch Kibana connects to, add a file `elasticSearchUrl` here, and put the URL in.
 
 If no URL has been set and ElasticSearch is running on this machine the local ElasticSearch is used automatically.

```

zagy (changes requested):

PR #75, opened 2020-03-23
Remove wrong or redundant Unit setting in systemd timer units

dpausp:
The default is to activate the service with the same name as the timer.
Specifying the service name without ending causes a warning.

## Release process

Impact:

Changelog:

* Fix warning for ignored setting in fc-prometheus-update-relayed-nodes timer (#124132).

## Security implications

n/a


ctheune:
Sounds good!

PR #76, opened 2020-03-24
Add support for multi-node Kubernetes (experimental)

dpausp:
Uses Kubernetes 1.15 from NixOS 19.09.

Roles:

* kubernetes-master
  * no pods, exactly one per RG
* kubernetes-node
  * joins existing cluster in RG automatically
  * runs pods, at least one per RG

kubernetes-master and kubernetes-node can be combined on a host.
Activating kubernetes-node on an existing master-only VM works, but taints must
be disabled manually to run pods on it.

Policy routing is disabled on Kubernetes VMs because it interferes with
routing of Kubernetes service IPs.

Other VMs in a RG with a Kubernetes cluster can also access service IPs
via a route set to the first Kubernetes node.

Dashboard and API allow public access for authenticated users.

sudo-srv users can generate a kubeconfig usable for kubectl and the
dashboard by running `kubernetes-make-kubeconfig` on the master.

Case 122936

## Release process

Impact:

Changelog:

* Add experimental support for multi-node Kubernetes (#122936).

## Security implications

  - dashboard and API allow only authenticated access using HTTPS (certs: API self-signed, dashboard letsencrypt)
  - access to cluster-admin client cert keys only for sudo-srv users, every user gets a separate certificate
  - internal Kubernetes components listen only on srv / localhost
  - automated test checks if permissions for the admin cert key of a sudo-srv user are set properly.
  - I checked open ports and cert key permissions on a dev VM.
  - I checked authentication via kubeconfig for public dashboard and API access.
  - Otherwise, we rely on the implementation in NixOS (certificate generation for Kubernetes components, for example) and Kubernetes (dashboard and API authentication, for example).


nixos/roles/kubernetes/default.nix
```diff
@@ -0,0 +1,77 @@
...
+
+  modules-19_09 = <nixos-19.09/nixos/modules>;
+in
+{
+
+  disabledModules = kubernetesModules;
+
+  imports = [
+    ./frontend.nix
+    ./singlenode.nix
```
ctheune: Kann wohl weg :P

nixos/roles/kubernetes/default.nix
```diff
@@ -0,0 +1,77 @@
+{ config, lib, pkgs, ... }:
+let
+
+  # Kubernetes services depend on autogenerated certs.
+  # Certmgr needs some seconds to generate them if they don't exist.
+  # We have to wait for the certs or the services will fail.
+  mkUnitWaitForCerts = name: certNames:
```
ctheune, line 7: Kandidat für upstream!

nixos/roles/kubernetes/master.nix
```diff
@@ -0,0 +1,292 @@
...
+  kublib = config.services.kubernetes.lib;
+  master = fclib.findOneService "kubernetes-master-master";
+
+  domain = config.networking.domain;
+  location = lib.attrByPath [ "parameters" "location" ] "standalone" config.flyingcircus.enc;
+  feFQDN = "${config.networking.hostName}.fe.${location}.${domain}";
+  srvFQDN = "${config.networking.hostName}.fcio.net";
+
+  # We allow frontend access to the apiserver for external dashboards and kubectl.
+  apiserverAddresses = [
```
ctheune: Tobias is wondering whether this is wrong? At least using the same names twices seems weird.

nixos/roles/kubernetes/master.nix
```diff
@@ -0,0 +1,292 @@
...
+    feFQDN
+    "kube.${fclib.currentRG}.fcio.net"
```
ctheune: Specific name? 'kubernetes.xxx.fcio.net' ... shorter version would be OK if we need a separate name for the api and the dashboard. If both run on the same name then I'd prefer kubernetes.xxx.fcio.net

nixos/roles/kubernetes/master.nix
```diff
@@ -0,0 +1,292 @@
...
+    ];
+
+    # Policy routing interferes with virtual ClusterIPs handled by kube-proxy, disable it.
+    flyingcircus.network.policyRouting.enable = false;
+
+    flyingcircus.activationScripts.kubernetes-apitoken =
+      lib.stringAfter [ "users" ] ''
+        mkdir -p /var/lib/cfssl
+        umask 077
+        tok=/var/lib/cfssl/apitoken.secret
```
ctheune: 2 letters saved! \o/ ....... not, please :)

nixos/roles/kubernetes/master.nix
```diff
@@ -0,0 +1,292 @@
...
+
+    networking.firewall = {
+      allowedTCPPorts = [ apiserverPort ];
+    };
+
+    services.kubernetes = {
+      addons.dashboard.enable = true;
+      addonManager.addons.kubernetes-dashboard-svc = lib.mkForce dashboardSvc;
+      apiserver.extraSANs = apiserverAddresses;
+      # Changing the masterAddress is tricky and requires manual intervention.
```
ctheune, line 222: master is a precious thing and can't be moved around easily. please note this in the docs

pkgs/overlay.nix
```diff
@@ -17,6 +19,7 @@ in {
     });
   docsplit = super.callPackage ./docsplit { };
 
+
```
ctheune: oy!

ctheune (changes requested):

PR #77, opened 2020-03-26
Reload varnish on config changes instead of restarting

dpausp:
Restarting empties the cache, we want to avoid that.
Restart code is from 15.09, with added testing and a check if the config
has really changed.

Case 124192

## Release process

Impact:

* [NixOS 19.03] Varnish will be restarted.

Changelog:

* Reload varnish on config changes instead of restarting (#124192).

## Security implications

n/a

### DIFF
--- a/nixos/roles/webproxy.nix
+++ b/nixos/roles/webproxy.nix
@@ -1,15 +1,11 @@
 { config, lib, pkgs, ... }:
 
-with lib;
-
 let
   cfg = config.services.varnish;
   fccfg = config.flyingcircus.roles.webproxy;
   fclib = config.fclib;
 
-  varnishCfg = fclib.configFromFile /etc/local/varnish/default.vcl vcl_example;
-
-  vcl_example = ''
+  vclExample = ''
     vcl 4.0;
     backend test {
       .host = "127.0.0.1";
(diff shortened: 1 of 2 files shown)


nixos/roles/webproxy.nix
```diff
@@ -88,15 +97,38 @@ in
...
+          path = with pkgs; [ varnish procps gawk ];
+          reloadIfChanged = true;
+          restartTriggers = [ configFile ];
+          reload = ''
+            if pgrep -a varnish | grep  -Fq '${varnishCmd}'
+            then
+              config=$(readlink -e /etc/current-config/varnish.vcl)
+              # Varnish doesn't like slashes and numbers in config names.
+              name=$(tr -dc 'a-z' <<< $config)
+              varnishadm vcl.list | grep $name && echo "Config unchanged." && exit
```
ckauhaus: `grep -q` ?

nixos/roles/webproxy.nix
```diff
@@ -88,15 +97,38 @@ in
...
+              varnishadm vcl.load $name $config && varnishadm vcl.use $name
+
+              for vcl in $(varnishadm vcl.list | grep ^available | awk {'print $4'});
```
ckauhaus, line 117: `grep -q`

ckauhaus (changes requested):

nixos/roles/webproxy.nix
dpausp: we need the grep output here

PR #83, opened 2020-05-12
Make global statshost and location proxy production-ready

dpausp:
* Move globally allowed metrics definitions from roles to a single place in the
  statshost config. This makes it easier to see what is allowed without
  grepping all roles. Moreover, we don't have the same roles on all plattform
  versions and need a way to add their metrics.
but must support all of them on the global statshost.
* Add missing apache (15.09 only) and postfix metrics.
* Rename statshostproxy to statshostproxy-location and statshost to
  statshost-global.
* Test global statshost in connection with statshostproxy-location/collectdproxy.
* Location proxy support SSL now on an alternative port 9443.
* Add missing telegraf host metrics that were lost when porting to 19.03.

Case 123876

## Release process

Impact:

Changelog:

* Upgrade central statshost to NixOS 19.03 (#123876).

## Security implications

  - nginx only opens allowed ports 9090 and 9443 on location proxy
  - firewall rules allow only statshost-global to access 9090 and 9443 on location-proxy 
  - otherwise, firewall rules and opened ports should be the same as on services12, the global statshost running on 15.09 currently
  - automatic tests check that nginx only opens allowed ports on proxy; manual port and firewall rule checks in the interactive test runner


nixos/roles/statshost/default.nix
```diff
@@ -627,11 +623,11 @@ in
...
 
     # outgoing collectd proxy for this location
-    (mkIf (cfgProxyGlobal.enable && statshostService != null) {
+    (mkIf (cfgProxyLocation.enable && statshostService != null) {
       flyingcircus.services.collectdproxy.location = {
         enable = true;
-        statshost = cfgStatsGlobal.hostName;
-        listenAddr = config.networking.hostName;
+        statshost = cfgStats.hostName;
+        listenAddr = (head config.networking.interfaces.ethsrv.ipv6.addresses).address;
```
ckauhaus: why not `::`?

ckauhaus: Ok, wildcard seems a not so good idea because the firewall is open. Next best thing: use helper functions from fclib to get the first IPv6 address directly from ENC data. See nixos/roles/mailserver.nix for example.

dpausp: I added a specific ip6tables rule for ethsrv instead of opening the port to everywhere

PR #86, opened 2020-05-19
Retry nixos-rebuild switch if the first run fails

ckauhaus:
Case 126172

## Release process

Impact:

Changelog: Retry system update on failure if update in maintenance is enabled. This is expected to avoid failed maintenance jobs (#126172).

## Security implications

* unchanged

### DIFF
--- a/pkgs/fc/agent/fc/manage/manage.py
+++ b/pkgs/fc/agent/fc/manage/manage.py
@@ -29,9 +29,11 @@ ACTIVATE = """\
 set -e
 nix-channel --add {url} nixos
 nix-channel --update nixos
-nixos-rebuild switch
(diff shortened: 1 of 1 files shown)


ctheune (approved):
Initially I thought this is too simplistic. However, simplistic is actually sufficient here.

I wondered whether we'll need some kind of sleep in between and whether this has impact of overall load during releases. However: this is already in the 'switch' phase, so the load is already distributed by the scheduler. Running it back to back also has the benefit that it's much more likely that the nixpkgs information is still in the VFS cache. 

I'm not sure whether all instances would be fixed by running it back to back but we don't have specific evidence for or against it either, so I'm happy to try this first.

PR #88, opened 2020-05-28
Extra channels: 19.09 -> 20.03, drop 18.09

dpausp:
Replace packages and modules from 19.09 with the versions from 20.03.

Updates Kubernetes from 1.15 to 1.17 and uses the 20.03 rust toolchain for building our tools.

Our pinned nixpkgs needed an update to remove the Kubernetes options from rename.nix
that are already defined in the 20.03 modules.

18.09 is dropped. We only used libpcap 1.8 which is now included in the
overlay.

Case 124530

## Release process

Impact:

Changelog:

* Replace 19.09 channel with 20.03. Lowers RAM consumption of our periodic management task which should reduce swapping on VMs with only 1GB RAM. Updates Kubernetes to 1.17 (#124530).

## Security implications

  - No changes from our side.
  - Automatic tests have been run

### DIFF
--- a/nixos/platform/garbagecollect/default.nix
+++ b/nixos/platform/garbagecollect/default.nix
@@ -76,7 +76,7 @@ in {
 
       # keep in sync with pkgs/overlay.nix
       environment.etc."nixos/garbagecollect-protect-references".text = ''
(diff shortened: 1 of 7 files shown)


ckauhaus:
LGTM

PR #91, opened 2020-06-10
Update fc-userscan to 0.4.5

ckauhaus:
This makes fc-collect-garbage more forgiving to problematic file permissions/file ownership, i.e. when a user has non-readable files in its home directory.

Also improve test.

Case 124165

## Release process

Impact:

Changelog:

## Security implications

  - Improved reliability of fc-userscan
  - But: 3rd-party install procedures may dump code owned by root or other privileged accounts into service users' homes. These may contain references into the Nix store. fc-userscan now skips these files with warnings. This may lead to unwanted cleanups in the Nix store and may affect application reliability.

  - NixOS test provide
  - The effect outlined above must be watched and adjusted if this presents problems in practice.


ckauhaus:
PR needs to be reworked

PR #93, opened 2020-06-12
Add RabbitMQ 3.8 role

dpausp:
Works with the existing service rabbitmq37 which is named just rabbitmq now.
Removed a test line that doesn't work with 3.8 and doesn't add real
value.

Case 126083

## Release process

Impact:

Changelog:

* Add RabbitMQ 3.8 role.

## Security implications

- minimise attack surface area: no additional ports are opened
- use a version that still gets security updates
- check that the new version doesn't introduce features that will harm security
- reviewed changelog and checked manually on test VM if no additional ports are opened
- 3.8 is the current version of RabbitMQ. It's from NixOS 20.03 which still gets security fixes.

### DIFF
--- a/nixos/services/rabbitmq.nix
+++ b/nixos/services/rabbitmq.nix
@@ -3,7 +3,7 @@
 with lib;
 
 let
-  cfg = config.flyingcircus.services.rabbitmq37;
+  cfg = config.flyingcircus.services.rabbitmq;
 
   inherit (builtins) concatStringsSep;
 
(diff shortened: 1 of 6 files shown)


nixos/services/rabbitmq.nix
```diff
@@ -15,21 +15,21 @@ let
...
+          Queuing Protocol (AMQP) broker.
         '';
       };
 
       package = mkOption {
-        default = pkgs.rabbitmq_server_3_7;
+        default = pkgs.rabbitmq_server_3_8;
         type = types.package;
-        defaultText = "pkgs.rabbitmq_server_3_7";
+        defaultText = "pkgs.rabbitmq_server_3_8";
```
ckauhaus, line 30: `defaultText` really necessary? I don't think so

ckauhaus (commented):
Also tested upgrade from 3.7 to 3.8. Works

ckauhaus:
LGTM, but would suggest to remove defaultText (see above)

PR #102, opened 2020-06-30
firewall: Accept empty lines and POSTROUTING rules

ckauhaus:
Not allowing empty lines is a nuisance at least and does not work well
with rule files generated from templates.

We need to accept POSTROUTING rules to express masquerading.

Case 126819
Case 126822

## Release process

Impact:

Changelog: Accept empty lines and POSTROUTING rules in `/etc/local/firewall` (#126819, #126822).

## Security implications

- The empty lines change is functionally transparent
- POSTROUTING is necessary to prevent breakage of Docker installation which need to set up masquerading. Being able to install own POSTROUTING rules includes the possibility to create outgoing packets with non-local IP sender addresses. But the additional risk is low since this is already possible by other means. This issue needs to be addressed in the new overlay/underlay based network architecture.

### DIFF
--- a/nixos/platform/filter-rules.py
+++ b/nixos/platform/filter-rules.py
@@ -3,7 +3,6 @@ import fileinput
 import re
 import shlex
 import sys
-import os.path as p
 
 R_ALLOWED = re.compile(r'^(ip(6|46)?tables .*)?''$')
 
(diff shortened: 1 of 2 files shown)


dpausp (approved):
We have to discuss how to avoid builtin chains. This fixes the current problem and can be merged.

ckauhaus:
Follow-up case https://fb.flyingcircus.io/f/cases/126879/

PR #110, opened 2020-07-13
Provide coturn role

dpausp:
Uses current coturn 4.5.13 from unstable.

Case 126406

## Release process

Impact:

Changelog:

* Provide coturn (STUN/TURN server) role (#126406).

## Security implications

Using coturn can have very different security implications depending on the configuration. We provide a base config that is quite locked-down. Users must configure coturn for their needs and think about security implications for their use case.

- use current version
- secure defaults: firewalled by default, should not open unexpected ports
  - automated test checks basic functionality and that no unexpected ports are opened
  - manual checks on dev VM, checked open ports
  - checked default settings of the NixOS service
  - had a look at the configuration docs for coturn: https://github.com/coturn/coturn/blob/master/README.turnserver

### DIFF
--- a/nixos/lib/utils.nix
+++ b/nixos/lib/utils.nix
@@ -115,4 +115,9 @@ rec {
         config.flyingcircus.users.userData
       );
 
+  writePrettyJSON = name: x:
+  let json = pkgs.writeText "write-pretty-json-input" (toJSON x);
+  in pkgs.runCommand name { preferLocalBuild = true; } ''
+      ${pkgs.jq}/bin/jq . < ${json} > $out
+    '';
 }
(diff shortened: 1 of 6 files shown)


nixos/lib/utils.nix
```diff
@@ -115,4 +115,8 @@ rec {
         config.flyingcircus.users.userData
       );
 
+  writePrettyJSON = name: x:
+    pkgs.runCommand name { preferLocalBuild = true; } ''
+      echo '${toJSON x}' | ${pkgs.jq}/bin/jq . > $out
+    '';
```
ckauhaus, line 122: Quoting correct? JSON strings may contain `'` IIRC

ckauhaus (changes requested):

nixos/lib/utils.nix
dpausp: I write raw json to a file now and pass that to jq. Avoids quoting problems and makes debugging jq errors easier.

PR #119, opened 2020-07-23
126611 create lamp role

ctheune:

## Release process

Impact:

-

Changelog:

- Add a new role "LAMP" to provide an easy-to-use and managed version of an Apache server 
  (not intended to run on the frontend network) with PHP that only needs to be pointed at
  a service-user's docroot.

  Also includes support for the Tideways APM service, only the API key needs to be added.

## Security implications

We open ports 8000 freely but we don't expose them on ethfe, only on
localhost and ethsrv (relying on the firewall). The apache user is
limited to 'nobody:service' and access to docroot and upload directories
 et.al. is granted through the application deployment.

Tideways is bound to localhost so PHP can talk to it, it talks to the
internet in turn, when activated by the application deployment.

Tested by manual inspection.

### DIFF
--- a/nixos/roles/default.nix
+++ b/nixos/roles/default.nix
@@ -24,7 +24,6 @@ in {
     ./memcached.nix
     ./mongodb
     ./mysql.nix
-    ./lamp.nix
     ./nfs.nix
     ./nginx.nix
     ./postgresql.nix
--- a/nixos/roles/default.nix
+++ b/nixos/roles/default.nix
@@ -17,12 +17,14 @@ in {
     ./graylog.nix
     ./kibana.nix
     ./kubernetes
+    ./lamp.nix
     ./loghost.nix
     ./mailout.nix
     ./mailserver.nix
     ./memcached.nix
     ./mongodb
     ./mysql.nix
+    ./lamp.nix
     ./nfs.nix
     ./nginx.nix
     ./postgresql.nix
--- a/nixos/roles/lamp.nix
+++ b/nixos/roles/lamp.nix
@@ -24,7 +24,7 @@ in {
           output_buffering = On
           short_open_tag = On
           curl.cainfo = /etc/ssl/certs/ca-certificates.crt
-          sendmail_path = /var/setuid-wrappers/sendmail -t -i
+          sendmail_path = /run/wrappers/bin/sendmail -t -i
 
           ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
           ; Tideways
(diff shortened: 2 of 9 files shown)


nixos/roles/default.nix
```diff
@@ -17,12 +17,14 @@ in {
...
     ./kibana.nix
     ./kubernetes
+    ./lamp.nix
     ./loghost.nix
     ./mailout.nix
     ./mailserver.nix
     ./memcached.nix
     ./mongodb
     ./mysql.nix
+    ./lamp.nix
```
dpausp: duplicate

nixos/roles/lamp.nix
```diff
@@ -0,0 +1,157 @@
...
+    flyingcircus.roles.lamp.enable =
+      lib.mkEnableOption "Flying Circus LAMP stack";
+  };
+
+  config = lib.mkIf role.enable (
+    let
+      apacheConfFile = /etc/local/lamp/apache.conf;
+      phpiniConfFile = /etc/local/lamp/php.ini;
+
+      services.httpd.user = "s-vmonitor";
```
dpausp: services.httpd.user doesn't make sense here in the let. It's set to nobody later.

nixos/roles/lamp.nix
```diff
@@ -0,0 +1,157 @@
...
+          extension=${pkgs.php73Packages.imagick}/lib/php/extensions/imagick.so
+          extension=${pkgs.php73Packages.redis}/lib/php/extensions/redis.so
+
+          ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+          ; General settings
+
+          output_buffering = On
+          short_open_tag = On
+          curl.cainfo = /etc/ssl/certs/ca-certificates.crt
+          sendmail_path = /var/setuid-wrappers/sendmail -t -i
```
dpausp: This only works on 15.09. It's /run/wrappers/bin now.

dpausp (changes requested):

nixos/roles/default.nix
ctheune: thanks, copy-pasta!

nixos/roles/lamp.nix
ctheune: good catch!

nixos/roles/lamp.nix
ctheune: merci. also that reminded me off adding the mailstub to the RGs ...

PR #122, opened 2020-07-27
MongoDB 3.6 and 4.0 roles

dpausp:
* uses current versions 3.6.19 and 4.0.19

Case 126656

## Release process

Impact:

Changelog:

* Add MongoDB roles for 3.6 and 4.0. We support upgrades from 3.2 to 4.0 which must include every major version in between (#126656).

## Security implications

  - listens only on srv and localhost
  - use recent versions that include all security updates
  - new versions should not introduce features that harm security
  - automatic test checks basic functionality
  - manually checked open ports on a dev VM and ran some test inserts in the cli
  - checked release notes for security-related changes: https://docs.mongodb.com/manual/release-notes
  - 3.6.19 and 4.0.19 are the newest minor versions

### DIFF
--- a/pkgs/overlay.nix
+++ b/pkgs/overlay.nix
@@ -79,12 +79,34 @@ in {
 
   mailx = super.callPackage ./mailx.nix { };
   mc = super.callPackage ./mc.nix { };
-  mongodb_3_2 = super.callPackage ./mongodb/3.2.nix {
+
+  mongodb-3_2 = super.callPackage ./mongodb/3.2.nix {
     sasl = super.cyrus_sasl;
     boost = super.boost160;
     # 3.2 is too old for the current libpcap version 1.9, use 1.8.1
     libpcap = self.libpcap_1_8;
   };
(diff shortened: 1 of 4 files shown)


pkgs/overlay.nix
```diff
@@ -79,12 +79,34 @@ in {
...
-  mongodb_3_2 = super.callPackage ./mongodb/3.2.nix {
+
+  mongodb-3_2 = super.callPackage ./mongodb/3.2.nix {
     sasl = super.cyrus_sasl;
     boost = super.boost160;
     # 3.2 is too old for the current libpcap version 1.9, use 1.8.1
     libpcap = self.libpcap_1_8;
   };
+  mongodb_3_2 = self.mongodb-3_2;
+  mongodb-3_4 = super.mongodb;
```
ckauhaus, line 90: There is a chance that this breaks during future updates of our nixpkgs fork and mongodb-3_4 will contain something different than 3.4.x. Better guard this assumption in the test.

ckauhaus (changes requested):

PR #125, opened 2020-07-29
Allow service users to write the mysql password for root

dpausp:
We need that for automated deployments.

Case 126845

## Release process

Impact:

* MySQL will be restarted.
Changelog:

* MySQL root password can now be set by service users before MySQL starts by changing the password file. That is helpful for automated deployments (#126845). 

## Security implications

Service users can already read and change the root password at runtime. Allowing them to change to root password before mysql starts doesn't create new security risks.

  - mysql root password only readable and writable by service users and root
  - automated test checks root pw file permissions

### DIFF
--- a/nixos/services/percona.nix
+++ b/nixos/services/percona.nix
@@ -53,7 +53,7 @@ let
       echo -n "''${pw}" > ${cfg.rootPasswordFile}
     fi
     chown root:service ${cfg.rootPasswordFile}
-    chmod 640 ${cfg.rootPasswordFile}
+    chmod 660 ${cfg.rootPasswordFile}
 
     pw=$(<${cfg.rootPasswordFile})
     cat > /root/.my.cnf <<__EOT__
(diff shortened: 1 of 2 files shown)


ctheune:
Ok, there's a bit confusion here. We need more tests, I guess ... 

* why is mysql restarted to change the mode of the local file?
* can the root password be updated without restarting mysql?

MySQL not restarting should be part of the unit test.

ckauhaus:
1) systemd unit file changed
2) I don't expect that - but having a test for it would be sensible

ctheune:
Well. MySQL as a product doesn't require a restart to change the root password, so this is definitely something that I think people would consider a bug in our platform. They change the root passwort and run fc-manage and get an immediate restart of MySQL. Not fun.

ctheune:
Also: I know that it restarts because of the systemd unit file. The question was why do we do that to change the password?

ctheune:
We made a new issue for tracking how to update passwords better (or rather that we currently don't). This is done for now.

PR #127, opened 2020-08-04
Use new tideways daemon that fixes a crash in our test suite and when…

ctheune:
… running with the wrong user.

## Release process

Impact:

* Tideways daemons will be restarted.

Changelog:

* Update tideways to 1.6.6 so it doesn't crash in test suite now and can be run as a separate user (instead of nobody).

## Security implications

* restricting daemons to specific users reduces attack surface

* tideways daemon now tested in hydra tests

### DIFF
--- a/nixos/roles/lamp.nix
+++ b/nixos/roles/lamp.nix
@@ -149,10 +149,7 @@ Listen localhost:8001
           '';
           Restart = "always";
           RestartSec = "60s";
-          # The daemon currently crashes when run with the tideways user.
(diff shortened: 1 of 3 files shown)


nixos/roles/lamp.nix
```diff
@@ -152,7 +152,7 @@ Listen localhost:8001
           # The daemon currently crashes when run with the tideways user.
```
dpausp: comment is outdated

dpausp (changes requested):

PR #133, opened 2020-08-12
Fix fc-userscan permission problems

ckauhaus:
- Be more forgiving w.r.t. problematic access permissions, i.e. when a
  user has non-readable files in its home directory.

- Use setuid() instead of seteuid()

Got the setup of setuid execution wrong in detail: The real uid stays
with the invoking user but the effective uid is set to 0. In our case,
when forking into a user context from root, we need to set the ruid not
the euid.

Case 124165

## Release process

Impact:

Changelog: Improve robustness of fc-userscan, the garbage collection scanner. It doesn't stop on unreadable files now. Also fix setuid-root invocation: GC roots and cache files are opened with dropped privileges.

## Security implications

Principle of least privilege: fc-userscan makes use of its elevated privileges only when reading files, not when writing GC roots or cache files

Checking this fact explicitly via NixOS test.

### DIFF
--- a/nixos/platform/garbagecollect/default.nix
+++ b/nixos/platform/garbagecollect/default.nix
@@ -10,8 +10,6 @@ let
   log = "/var/log/fc-collect-garbage.log";
 
   garbagecollect = pkgs.writeScript "fc-collect-garbage.py" ''
-    #! ${pkgs.python3.interpreter}
-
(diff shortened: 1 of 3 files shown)


ckauhaus:
Also fixes case 126341

PR #146, opened 2020-09-10
Add Jitsi role

dpausp:
* supports fallback to TURN if UDP doesn't work
* Jibri integration for recording and livestreaming
  * Jibri itself must be provided separately
* optional authentication for room creation

Auth support and Jibri integration should be integrated upstream.

Case 127649

## Release process

Impact:

Changelog:

* Add Jitsi Meet role (#127649).

## Security implications

Jitsi is essentially a public service and anyone who knows a room link can join. Authentication is only relevant when someone wants to create a new room. Most of the implementation is from NixOS upstream which was reviewed there.

  - use recent versions
  - transmit data securely:  coturn and nginx only use TLS, video and audio traffic is encrypted by jitsi videobridge
  - only open required ports: firewall additional coturn ports, videobridge UDP port can be firewalled optionally to route everything via TLS-encrypted coturn, internal Jitsi components shouldn't be exposed to public internet
  - we use the current stable versions of Jitsi components except Jitsi Meet (just the frontend) which uses a recent dev version
  - manually tested on VM test46, checked open ports

### DIFF
--- a/nixos/roles/coturn.nix
+++ b/nixos/roles/coturn.nix
@@ -18,7 +18,6 @@ let
 
   domain = config.networking.domain;
   location = lib.attrByPath [ "parameters" "location" ] "standalone" config.flyingcircus.enc;
-  feFQDN = "${config.networking.hostName}.fe.${location}.${domain}";
   cfg = config.flyingcircus.roles.coturn;
 
   ourSettings = [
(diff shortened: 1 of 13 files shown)


ckauhaus:
Did not find anything suspicious on visual inspection. Didn't try it out though.

PR #155, opened 2020-09-30
128712 gitlab mailin

ckauhaus:

## Release process

Impact: Gitlab instances will be restarted.

Changelog: Gitlab: add processing of incoming mails via the mail_room component. Settings must be configured via the new `services.gitlab.incomingEmail` attribute set (#128712).

## Security implications

Security concerns set forth in https://docs.gitlab.com/ee/administration/incoming_email.html#security-concerns apply. Misconfigured mail processing can be a security issue so it should be done with care. Giblab's documentation, however, is quite explicit on this.

Instance running on cusyce00 does not show problems as outlined above because it uses a dedicated subdomain (ce.cusy.io).

### DIFF
--- a/nixos/services/gitlab/default.nix
+++ b/nixos/services/gitlab/default.nix
@@ -596,8 +596,8 @@ in {
 
         logPath = mkOption {
           type = types.path;
-          default = "${cfg.statePath}/log/mail_room.log";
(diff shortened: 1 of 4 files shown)


dpausp (commented):
mail room path has uneeded entries, log to /var/log/gitlab

PR #158, opened 2020-10-05
Agent: use spread time in channel-with-maintenance mode, check every hour

dpausp:
Spread time was already in use for the channel update mode which we use
on non-prod VMs. Production VMs, however, first build the new channel
and schedule the switch in a maintenance window. Spread wasn't used in
that case. That meant that all production VMs built the new channel directly
after the release in a short time window. Building is quite
resource-intensive and can cause overload on our storage systems.
Building the new channel now also uses the spread time.
It's now 60min instead of 120min to roll out channel updates faster.

Also removes global spread variable and deprecated dev mode.

## Release process

Impact:

Changelog:

* Agent: Spread build start time for updates on production VMs over a longer time period. This should decrease load on our systems and improve availability during release rollouts (#127919).

## Security implications

  - should improve availablity by decreasing load on our systems
  - manually tested on test35

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -46,7 +46,7 @@ in {
 
       interval = mkOption {
         type = types.int;
-        default = 120;
+        default = 60;
         description = "Run channel updates every N minutes.";
       };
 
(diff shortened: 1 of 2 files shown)


ckauhaus:
TS will coordinate a real-world test before release.

PR #166, opened 2020-10-22
Use hashedPassword for user vagrant on vagrant box

frlan:
This PR uses hashedPassword instead of password for creation of the user vagrant as well as adds the user vagrant to group login. Both is used for generating /etc/local/nginx/htpasswd_fcio_users

bugs id: #129041

## Release process

Impact:

Changelog:
* /etc/local/nginx/htpasswd_fcio_users can now be used on Vagrant with the vagrant user

## Security implications

* Password still working for the user vagrant
* /etc/local/nginx/htpasswd_fcio_users with a entry for user vagrant

### DIFF
--- a/nixos/infrastructure/vagrant.nix
+++ b/nixos/infrastructure/vagrant.nix
@@ -39,8 +39,8 @@ in {
     # These groups are created through the ENC normally which
     # doesn't exist in vagrant or at least here, yet.
     users.groups = {
-      login = {};
-      service = {};
+      login = { members = ["vagrant"]; };
+      service = { members = ["vagrant"]; };
       sudo-srv = {};
       admins = {};
     };
@@ -48,8 +48,9 @@ in {
     users.users.vagrant = {
       description = "Vagrant user";
       group = "users";
-      extraGroups = [ "login" "service" "docker" ];
-      password = "vagrant";
+      extraGroups = [ "docker" ];
+      # password: vagrant
+      hashedPassword = "$5$xS9kX8R5VNC0g$ZS7QkUYTk/61dUyUgq9r0jLAX1NbiScBT5v1PODz4UC";
       home = "/home/vagrant";
       isNormalUser = true;
       openssh.authorizedKeys.keys = [


nixos/infrastructure/vagrant.nix
```diff
@@ -39,7 +39,7 @@ in {
     # These groups are created through the ENC normally which
     # doesn't exist in vagrant or at least here, yet.
     users.groups = {
-      login = {};
+      login = { members = ["vagrant"]; };
```
ctheune, line 42: Hmm. This seems to be a duplicate. Line 51 already assigns login as an extra group to the vagrant user.

nixos/infrastructure/vagrant.nix
```diff
@@ -49,7 +49,7 @@ in {
       description = "Vagrant user";
       group = "users";
       extraGroups = [ "login" "service" "docker" ];
-      password = "vagrant";
+      hashedPassword = "$5$xS9kX8R5VNC0g$ZS7QkUYTk/61dUyUgq9r0jLAX1NbiScBT5v1PODz4UC";
```
ctheune, line 53: Can you put a comment with the plaintext password here so that the uninitiated know what it is? It's a public password anyway, so ...

ctheune (changes requested):

nixos/infrastructure/vagrant.nix
ctheune: Actually I meant to the opposite of what happened now: why are you extracting those here? Is the code on line 51 buggy somehow?

frlan: It was. The logic behind generating the htpasswd-file actually did not respect the settings done with `users.users.vagrant.extraGroups` -- in an update I moved login and service out of the extraGroups, kept docker.

nixos/infrastructure/vagrant.nix
frlan: done

nixos/infrastructure/vagrant.nix
ctheune: Ugh. "interesting". Alright then. Actually that means we should be writing a test for that htpasswd generation code and maybe fix that bug ...

ctheune (approved):

PR #177, opened 2020-12-01
Nginx: start master process as root, worker as nginx

dpausp:
This avoids problems with certificates that are not readable by the
nginx user, for example.
Upstream changed the service to start as nginx but we need
compatibility with older installations.
In the future, SSL certificates should be auto-generated by the
letsencrypt automation NixOS provides.

PL-129520

## Release process

Impact:

Changelog:

* Nginx: start as root and allow the master process to read protected files. The workers run as nginx user and drop extended capabilities. This ensures compatibility with our 19.03 platform (PL-129520).

## Security implications

  - worker processes should run as unprivileged user nginx
  - cache directories should only be readable by nginx user
  - manually checked permissions and effective UID of worker processes on VM test45, automated test checks basic functionality, added check for cache dir permissions

### DIFF
--- a/nixos/services/nginx/base-module.nix
+++ b/nixos/services/nginx/base-module.nix
@@ -64,6 +64,8 @@ let
   '';
 
   configFile = pkgs.writers.writeNginxConfig "nginx.conf" ''
+    user ${cfg.user} ${cfg.group};
+
     pid /run/nginx/nginx.pid;
     error_log ${cfg.logError};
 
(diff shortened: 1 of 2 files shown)


dpausp:
This didn't actually fix the problem as promised, it's fixed in https://github.com/flyingcircusio/fc-nixos/commit/c6c58ad22bd50f8fb2c0f22c40f6bd49315828b0

PR #189, opened 2020-12-17
Own FQDN should resolve to SRV address, not 127.0.0.1

dpausp:
This was changed upstream in NixOS 20.09 which causes too much trouble
for us when upgrading. Fix is from the NixOS manual. Maybe we can go back
to the upstream default in the future when everything is on 20.09.

 #PL-129549

## Release process

Impact:

Changelog:

* Own FQDN (like test99.fcio.net) should resolve to SRV address, not 127.0.0.1. This change restores the behaviour of our older platform versions (#PL-129549).

## Security implications

  - own FQDN should resolve to SRV address
  - manually checked entries in /etc/hosts on test45

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -121,6 +121,13 @@ in
     '';
 
     networking = {
+
+      # FQDN and host name should resolve to the SRV address
+      # (set by hostsFromEncAddresses) and not 127.0.0.1.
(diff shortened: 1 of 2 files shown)


ctheune (changes requested):
I'd really like to see an automated test for this...

PR #198, opened 2021-01-18
Agent: remove result link when channel switch succeeded

dpausp:
Use a meaningful location for the result link created by nixos-rebuild
dry-run when preparing the channel switch and delete it
when nixos-rebuild switch suceeded in the maintenance run.

 #PL-129591

## Release process

Impact:

Changelog:

* Use /run/next-system/result instead of /result for the link to the prepared system update and delete it after a successful maintenance run. (#PL-129591).

## Security implications

  - nothing 
  - manually checked if it works on a test VM

### DIFF
--- a/pkgs/fc/agent/fc/manage/manage.py
+++ b/pkgs/fc/agent/fc/manage/manage.py
@@ -24,13 +24,22 @@ import tempfile
 _log = logging.getLogger()
 enc = {}
 
-ACTIVATE = """\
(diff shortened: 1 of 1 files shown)


ctheune (changes requested):
As discussed putting stuff in / is an anti-pattern and the case for the optimisation to avoid additional builds after a reboot isn't strong enough.

Please move to /run/next-system.

PR #203, opened 2021-01-26
Graylog: create separate telegraf users to avoid conflict

dpausp:
Add an hostname suffix to the telegraf user name to make it unique in
the cluster so each node can have its own telegraf password.
Before, each node overwrote user "telegraf" with a different password
which meant that only one node had the correct password and could gather
metrics.

In a graylog cluster, we want metrics from each graylog node.

 #PL-129455

## Release process

Impact:

Changelog:

* Graylog: make telegraf metrics collection work for all nodes in a cluster (#PL-129455).

## Security implications

  -  metrics collection should work on all nodes without sharing passwords between nodes
  - automated test checks if telegraf user is created
  - manual check if metrics collection works on a test VM

### DIFF
--- a/nixos/services/graylog/default.nix
+++ b/nixos/services/graylog/default.nix
@@ -104,6 +104,7 @@ let
 
   graylogConfPath = "/run/graylog/graylog.conf";
 
+  telegrafUsername = "telegraf-${config.networking.hostName}";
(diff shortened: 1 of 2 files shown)


ctheune:
Note that this leaves behind a on old telegraf user that will be there forever. Can we delete that?

ctheune:
As discussed personally @ts will review the existing instances manually.

PR #237, opened 2021-03-26
Agent: only switch to new system if config changes (lazy activation)

dpausp:
fc-manage now has a -l flag that skips the switch if the build
result is the same as the running system.

The flag is used by default for automatic agent runs and can be
disabled with the NixOS option flyingcircus.agent.lazySwitch.

This saves some CPU time and avoids triggering targets on every agent
run which is a problem for the ACME renewal targets, at least.

May have some side effects because tmpfile rules are not executed every
10 minutes and inactive units that should run will not be activated
anymore. If something like that is really needed, an explicit timer
or unit settings like Restart=always should be used to trigger that.

Also adds an option flyingcircus.agent.verbose to enable debugging
output.

 #PL-129745

## Release process

Impact:

Changelog:

* fc-agent: the automatic management task that runs every ten minutes now only switches to the new system configuration if it actually has changed with the last build. We already used that behaviour on the 15.09 platform. This saves some CPU time and avoids frequent ACME certificate renewal checks which also trigger unneccessary Nginx reloads (#PL-129745).

## Security implications

  - agent should properly pick up config changes, build it and switch to the new system if something changed 
  - errors in agent should be reported properly so we find and diagnose problems  
  - checked on a test VM that building/switching works with local dev checkouts and branch channels 
  - checked on a test VM that errors in all stages (build, switch, register) are properly logged/printed

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -49,7 +49,7 @@ in {
         default = "";
         description = ''
           Additional commands to execute within an agent run
-          after the main NixOS configuration/build has been 
+          after the main NixOS configuration/build has been
           activated.
         '';
       };
@@ -60,6 +60,20 @@ in {
         description = "Run channel updates every N minutes.";
       };
 
+      lazySwitch = mkOption {
+        default = true;
+        description = ''
+          Only switch to the new configuration if the system has actually
+          changed.
+        '';
+        type = types.bool;
+      };
+
+      verbose = mkOption {
+        default = false;
+        description = "Enable additional logging for agent debugging.";
+        type = types.bool;
+      };
     };
   };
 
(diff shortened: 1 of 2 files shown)


nixos/platform/agent.nix
```diff
@@ -107,12 +121,15 @@ in {
...
+          let
+            interval = toString cfg.agent.interval;
+            lazy = lib.optionalString cfg.agent.lazySwitch "-l";
+            verbose = lib.optionalString cfg.agent.verbose "-v";
           in ''
             rc=0
             timeout 14400 ionice -c3 \
               fc-manage -E ${cfg.encPath} -i ${interval} \
-              ${cfg.agent.steps} || rc=$?
+              ${cfg.agent.steps} ${lazy} ${verbose} || rc=$?
```
ctheune: I'd recommend putting the free form steps last.

pkgs/fc/agent/fc/manage/manage.py
```diff
@@ -118,15 +136,93 @@ def check_local_channel(self):
...
+            raise BuildFailed()
+
+        result_path = proc.stdout.read().strip()
+        _log.debug("Built channel, result is: %s", result_path)
+        assert result_path.startswith("/nix/store/"), \
+            f"Output doesn't look like a Nix store path: {result_path}"
+        self.result_path = result_path
+
+    def switch_to_configuration(self, lazy):
+
```
ctheune: superfluous empty line

ctheune (changes requested):

PR #270, opened 2021-05-03
Stabilise monodb. Now it does not fail anymore just for timing reasons

nichmoe:
Now it does not fail anymore just for timing reasons.
    #PL-129796

## Release process

Impact:

NONE

Changelog:

NONE

## Security implications

    - Test results should be consisten with same input. 

    - Test was run multiple times without failing

### DIFF
--- a/tests/mongodb.nix
+++ b/tests/mongodb.nix
@@ -72,6 +72,7 @@ in {
         machine.succeed("killall -11 mongod")
         import time
         time.sleep(5)
+        machine.wait_for_unit("mongodb.service")
         machine.succeed("systemctl show mongodb | grep ActiveState=active")
         _, out = machine.execute('pgrep mongod')
         new_pid = int(out.strip())


tests/mongodb.nix
```diff
@@ -44,6 +44,7 @@ in {
       machine.wait_for_unit("mongodb.service")
       machine.wait_for_open_port(27017)
       machine.wait_until_succeeds('mongo --eval db')
+      machine.wait_for_unit("mongodb.service")
```
dpausp: redundant: we already wait for the same service in line 44

tests/mongodb.nix
```diff
@@ -72,6 +73,7 @@ in {
         machine.succeed("killall -11 mongod")
         import time
```
dpausp, line 73: I think that the sleep is not needed anymore when there's a wait_for_unit later

dpausp (changes requested):

tests/mongodb.nix
nichmoe: I tested without sleep, but unfortunetly it skips the machine.wait_for_unit line otherwise.

PR #282, opened 2021-06-01
redis: add maxmemory option, add memoryPercentage option (fixes PL-115928)

mkg20001:
Fixes PL-115928

cc @dpausp 

Currently memoryPercentage has a default value (50%). Is that ok like that or should nothing be set by default?

## Release process

Impact:

Changelog:

* Redis: add options `flyingcircus.services.redis.maxmemory` and `flyingcircus.services.maxmemory-policy`. Memory used by Redis is limited to 80% by default (#PL-115928).

## Security implications

  - limit memory consumption of Redis to avoid problems with availability/performance when memory fills up  (by dpausp)
  - checked on a test VM that memory limits are added to the config by default and that it can be changed (by dpausp)

### DIFF
--- a/nixos/services/redis.nix
+++ b/nixos/services/redis.nix
@@ -43,6 +43,40 @@ in {
         description = "The precise Redis package to use";
         example = "pkgs.redis";
       };
+
+      maxmemory = mkOption {
+        type = types.str;
+        default = "${toString ((fclib.currentMemory 1024) * cfg.memoryPercentage / 100)}mb";
+        description = "Maximum memory redis is allowed to use for a dataset";
+        example = "100mb";
+      };
+
+      maxmemory-policy = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = ''
+          The exact behavior Redis follows when the maxmemory limit is reached is configured using the maxmemory-policy configuration directive.
+
+          The following policies are available:
+
+          - noeviction: return errors when the memory limit was reached and the client is trying to execute commands that could result in more memory to be used (most write commands, but DEL and a few more exceptions).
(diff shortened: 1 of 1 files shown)


nixos/services/redis.nix
```diff
@@ -67,6 +101,14 @@ in {
         package = cfg.package;
         requirePass = password;
         vmOverCommit = true;
+        settings = lib.mkMerge [
+          (lib.mkIf (config.flyingcircus.services.redis.maxmemory-policy != null) {
+            maxmemory-policy = config.flyingcircus.services.redis.maxmemory-policy;
+          })
+          ({
+            maxmemory = config.flyingcircus.services.redis.maxmemory;
```
dpausp: better use `memory = cfg.maxmemory;` (cfg defined at line 6) or even `inherit (cfg) maxmemory`;

dpausp (changes requested):
50% as a default is ok, just some code-style changes requested :)

nixos/services/redis.nix
```diff
@@ -43,6 +43,40 @@ in {
...
+          - volatile-ttl: evict keys with an expire set, and try to evict keys with a shorter time to live (TTL) first, in order to make space for the new data added.
+
+          Read more at https://redis.io/topics/lru-cache
+        '';
+        example = "noeviction";
+      };
+
+      memoryPercentage = mkOption {
+        type = types.int;
+        default = 50;
```
ctheune: Talked to @zagy and looking at the existing instances we should go with an 80% default. @mkg20001

ctheune (changes requested):

PR #296, opened 2021-06-15
fix: php loading config twice and throwing warning

mkg20001:
```
Cannot load Zend OPcache - it was already loaded
```
This is an ugly hack because overrideAttrs doesn't seem to apply if 
other modules override aswell

@dpausp 

NOTE: I haven't yet tested running something that uses php via apache with it, but I assume apache gives php the config explicitly so it should work that way.

## Release process

Impact:

Changelog:
- Fix php loading config twice and throwing warning

## Security implications

  - another warning that is to be ignored, making the user blind to actual problems
  - local vagrant machine

### DIFF
--- a/pkgs/overlay.nix
+++ b/pkgs/overlay.nix
@@ -7,6 +7,33 @@ let
   nixpkgs_18_03 = import versions.nixpkgs_18_03 {};
   inherit (super) lib;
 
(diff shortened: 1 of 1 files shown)


mkg20001:
Superseeded by #308

PR #308, opened 2021-06-28
LAMP: don't load PHP extensions twice in PHP CLI & add php56 from fossar/nix-phps

mkg20001:
This includes php56 from github.com/fossar/nix-phps
~~(currently pointing to a PR branch, see fossar/nix-phps#28)~~
now merged, pointing to upstream

NOTE: This adds a requirement on unstable,
since the code requires some exposed functions

Also added @dpausp's fix for the php config which broke due to the php56 config being different, which is now fixed due to php56 being built like all the other phps

Plus the 18.03 dependency can finally be removed

## Release process

Impact:

Changelog:
- php56 now being pulled from fossar/nix-phps and built with current nixpkgs

## Security implications

  - security: remove outdated 18.03 libraries, rebuild with current nixpkgs instead
  - security: fix warning about opcache being loaded twice to prevent warning blindness
  - no longer any 18.03 dependencies present
  - warning no longer present, tested in test vm

### DIFF
--- a/nixos/roles/lamp.nix
+++ b/nixos/roles/lamp.nix
@@ -120,10 +120,10 @@ in {
 
           services.httpd.enable = true;
           services.httpd.adminAddr = "admin@flyingcircus.io";
-          environment.shellInit = ''
-            export PHPRC='${config.systemd.services.httpd.environment.PHPRC}'
-          '';
-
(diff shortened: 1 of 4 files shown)


ctheune (changes requested):
Looks good to me. Tobias requested a squash before merge.

PR #310, opened 2021-06-30
Pressure stalling

elikoga:

## Release process

Impact:

Changelog:

* Add PSI metrics that will replace the traditional "load" metric in the future. (PL-129904)

## Security implications

* Just adding metrics (and currently unused options for the sensu check). Exposing this data to Telegraf is fine.

nothing to test, functionality is tested via nixos test


pkgs/fc/sensuplugins/fc/sensuplugins/pressure_stall_information.py
```diff
@@ -0,0 +1,89 @@
...
+        for line in dev:
+            entries = re.match(PSI_PATTERN, line)
+            if not entries:
+                continue
+            entries = entries.groupdict()
+            actual[entries['extent']] = (entries['avg10'], entries['avg60'], entries['avg300'])
+
+    exit_code = 0
+
+    if not actual:
```
ctheune: I'd suggest using set(actual) != set(critical). Also in this case we loose access to the data that was actually measured.

elikoga:
Rewriting with nagiosplugin in order to have syntax for metrics managed

ctheune:
Just to make sure we're talking about the same thing: you're talking about the command line parsing and infrastructure for the check. We do not use the Nagios metrics API to forward the metrics data elsewhere but we'll need to extract the PSI data through telegraf. I think there is some code around somewhere already in a branch that needs to be finalized (I can see we have _some_ PSI data in prometheus, but not for all current VMs). Using this data we can then determine thresholds we want to alert on. Doing this (and adjusting the Grafana dashboards for the 3 PSI metrics) before finishing the check would reduce latency that would be necessary to wait for at least 1-2 weeks of data to flow in.

ctheune:
Also, one more thing: please stop force pushing in single commits. I'd like to be able to look together (with team mates) on what you're working on. Figuring out what changed after you force pushed is unnecessarily complicated. We can clean up before a merge (and then use a force push) later.

elikoga:
> Also, one more thing: please stop force pushing in single commits. I'd like to be able to look together (with team mates) on what you're working on. Figuring out what changed after you force pushed is unnecessarily complicated. We can clean up before a merge (and then use a force push) later.

Got it, I force pushed the previous commit from the reflog so you can see the history.

> Just to make sure we're talking about the same thing: you're talking about the command line parsing and infrastructure for the check. [...]

Yup, I had the idea to use https://github.com/influxdata/telegraf/tree/master/plugins/parsers/nagios to extract the data from the sensu check itself.

I had the impression that the sensu checks had to abide to the nagios api but that was pretty wrong.

Now that I read more I saw that the Sensu check does not have to abide to the Nagios api but only have correct exit code so one could output the data for telegraf consumption in some other way

ctheune:
> Now that I read more I saw that the Sensu check does not have to abide to the Nagios api but only have correct exit code so one could output the data for telegraf consumption in some other way

@nichmoe you had started something there somewhere already, care to point us to any code that you already had? Maybe I'm misremembering things right now. I'm starting to remember that telegraf doesn't have a built-in sensor so maybe @elikoga might even have been right, but in any case I would not use the Nagios format to pass things through to Telegraf and rather use something like json that telegraf can consume directly.

elikoga:
Some of the work is at de7c62d PL-129904-Replace-check_load-with-check_psi https://github.com/flyingcircusio/fc-nixos/compare/PL-129904-Replace-check_load-with-check_psi which apparently is not findable on the web interface at and afaik it does not go much much further than what is here from what I can see

IMO json or nagios does not really make a difference since they are at they same level of "directly" imo
https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md

ctheune:
Well, the Nagios format has a lot of deficiencies and is quite out of fashion. I see that telegraf supports that format, but I'd still rather use the json format as it will be easier to use in other services later. Also @nichmoe did already have the telegraf config available (I remember piecing it together) somewhere, maybe he could throw it over.

nichmoe:
[This should be the config you mentioned.](https://github.com/flyingcircusio/fc-nixos/blob/6806663b0a202082527a4e7e64ec05473045a661/nixos/services/telegraf/psi.conf)

elikoga:
Pressure data is now delivered to Telegraf, I created a package in the pkgs folder to hold the script

ctheune:
Please add test coverage to show that PSI is included in the telegraf prometheus output.

PR #316, opened 2021-07-08
PL-129922 rebase nginx to fix gzip error

mkg20001:

## Release process

Impact:

Changelog:
- Rebase nginx mainline and stable with https://github.com/flyingcircusio/nginx/commit/2ad7b63de0391df4c49c887f2929a72658bce329 ontop
- Also use modsecurity-nginx from upstream as they have the same version now

Additionally I've seen that the patch for remote_addd_anon just strips the IPv6 to /64. Not sure how good that is considering handing out /48s to a single subscriber is a common practice in some places plus a /64 usually is one network, so while we don't get the device MACs (64-128) from the address, we still get a /64 that is tied to one particular internet subscriber

Should I make it zero everything after /48 instead of /64?

## Security implications

  - data protection (GDPR): IP addresses in log files must be anonymized.
  - checked proper anonymization of IPv4 and IPv6 on a test VM
  - automated test still runs

### DIFF
--- a/pkgs/remote_addr_anon.patch
+++ b/pkgs/remote_addr_anon.patch
@@ -1,0 +1,136 @@
+From 222c9f32c7be62e8789b71e70395fddef8f0fa76 Mon Sep 17 00:00:00 2001
+From: Christian Theune <ct@flyingcircus.io>
+Date: Wed, 16 Sep 2020 17:15:14 +0200
+Subject: [PATCH] Anonymize logs by default, provide remote_addr_anon as a
+ builtin variable.
+
+---
+ src/http/modules/ngx_http_log_module.c |  2 +-
+ src/http/ngx_http_variables.c          | 84 ++++++++++++++++++++++++++
+ 2 files changed, 85 insertions(+), 1 deletion(-)
+
+diff --git a/src/http/modules/ngx_http_log_module.c b/src/http/modules/ngx_http_log_module.c
+index f7c4bd2f..615b2c36 100644
+--- a/src/http/modules/ngx_http_log_module.c
++++ b/src/http/modules/ngx_http_log_module.c
+@@ -225,7 +225,7 @@ static ngx_str_t  ngx_http_access_log = ngx_string(NGX_HTTP_LOG_PATH);
+ 
+ 
+ static ngx_str_t  ngx_http_combined_fmt =
+-    ngx_string("$remote_addr - $remote_user [$time_local] "
++    ngx_string("$remote_addr_anon - $remote_user [$time_local] "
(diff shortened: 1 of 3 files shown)


ctheune:
The IPv6/48 is a good catch. I reviewed a bit of documentation. Google does /48's too. I think this should be sufficient. Some people talk about cutting of at /32 but that seems excessive. Individual subscribers usually get /56 so a /48 would already give us a good cut. Company accounts usually might /48s and larger assignments are typically for LIRs. 

So yes, please turn this into a /48. We also need to adjust our data protection documentation (public and internally) about this.

mkg20001:
Another idea instead of the custom repo would be to just apply the patch as a regular patch via patches = [];

I think that's better as that saves us from rebasing for every minor and should the patch break tests can catch it

dpausp (changes requested):
Mainline nginx works with remote_addr_anon so the hack in tests/nginx.nix lines 189-191 should be removed.

mkg20001:
@dpausp Done

pkgs/remote_addr_anon.patch
```diff
@@ -0,0 +1,136 @@
...
++        v->valid = 1;
++        v->no_cacheable = 0;
++        v->not_found = 0;
++
++        v->data = ngx_pnalloc(r->pool, len);
++        if (v->data == NULL) {
++            return NGX_ERROR;
++        }
++
++        for (i=6; i<16; i++) {
```
ctheune, line 84: Can I get a PR for the /48 change to my repo please?

tests/nginx.nix
```diff
@@ -186,9 +186,6 @@ in {
       # Prepare change to mainline nginx. We are not interested in testing mainline itself here.
       # We only need it as a different version so we can test binary reloading.
       server.execute("ln -sfT ${pkgs.nginxMainline} /etc/nginx/running-package")
-      # Mainline doesn't know about our custom remote_addr_anon variable, patch our config.
```
ctheune, line 189: Can we please also add explicit tests that our default config is in fact anonymizing the IPs?

ctheune (changes requested):
Looks good in general, see the two comments.

tests/nginx.nix
mkg20001: Added

pkgs/remote_addr_anon.patch
mkg20001: It's rebased against another branch of nginx, so a PR won't work as usual afaik

You can pull in nginx-1.21.1-fc-patched and nginx-1.20.1-fc-patched from git@github.com:mkg20001/nginx into the flyingcircus/nginx repo

PR #284, opened 2021-06-03
PL-129868 structured haproxy config

elikoga:
@flyingcircusio/release-managers

## Release process

Impact:

- Haproxy service will be reloaded/restarted

Changelog:

- Adds nix options to configure haproxy under the key `flyingcircus.services.haproxy`
- Migrate internal services to haproxy configured with nix

## Security implications

- [x] [Security requirements](https://wiki.flyingcircus.io/System_Development_Guideline#Security_requirement_principles_and_testing) defined? (WHERE)
  - Haproxy config can be read from nix store - has to be unconfidential
- [x] Security requirements tested? (EVIDENCE)
  - Haproxy can be configured through nix options
  - System still builds
  - haproxy.nix test passes

PR #318, opened 2021-07-13
HAproxy: structured Nix config, use it for Graylog and Kubernetes

dpausp:
The old way of putting plain config in /etc/local/haproxy
still works as expected, structured config must be activated
explicitly with `flyingcircus.services.haproxy.enableStructuredConfig`
Both can be used at the same time.

Graylog and Kubernetes haproxy config have been rewritten with the
structured approach. This allows combining these roles with roles
that also want to set haproxy config without overwriting each other.

See nixos/services/haproxy/README.md for details.

 #PL-129868

## Release process

See original PR #284


dpausp:
I reviewed the original PR and made some small changes that don't affect the functionality.

PR #327, opened 2021-07-20
mail: upgrade mailserver fork

mkg20001:
Note: Currently referencing my own fork, re-hosted on git.mkg since
github doesn't like the 100mb files of upstream and import failed last
week when I tried it

I've rebased all commits from christian kauhaus, you can review the list
here
https://git.mkg20001.io/mkg20001/fc-nixos-mailserver/-/commits/fc-rebase

## Release process

Impact:
- maybe mailservices will be restarted

Changelog:
- upgrade nixos-mailserver to latest master

## Security implications

  - upgrade nixos-mailserver
  - tested using tests and testvm

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -6,9 +6,11 @@
     "sha256": "0avbsx5chbwr0y55shndkzf0ixx3bznbzq526p5nj8llryxa10af"
   },
   "nixos-mailserver": {
-    "owner": "flyingcircusio",
-    "repo": "nixos-mailserver",
-    "rev": "d4566383b01c081bed866d397909a9e4fbefc7e2",
-    "sha256": "0hl31fjn0jf6j2rhf8g799q9vgmixyyy5j9srfa1njhy1npb04v3"
+    "url": "https://gitlab.flyingcircus.io/flyingcircus/nixos-mailserver.git/",
+    "rev": "32c90985be7ae42736716044f00c28435492a462",
+    "sha256": "0cvpvq13ballwfi0qcq82dc977j5qmgqmwvhrja950wsfkh68dwm",
+    "fetchSubmodules": false,
+    "deepClone": false,
+    "leaveDotGit": false
   }
 }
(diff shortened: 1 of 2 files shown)


versions.json
```diff
@@ -6,9 +6,11 @@
     "sha256": "1056r3383aaf5zhf7rbvka76gqxb8b7rwqxnmar29vxhs9h56m5k"
   },
   "nixos-mailserver": {
-    "owner": "flyingcircusio",
-    "repo": "nixos-mailserver",
-    "rev": "d4566383b01c081bed866d397909a9e4fbefc7e2",
-    "sha256": "0hl31fjn0jf6j2rhf8g799q9vgmixyyy5j9srfa1njhy1npb04v3"
+    "url": "https://git.mkg20001.io/mkg20001/fc-nixos-mailserver",
```
dpausp: change this to point to the fcio gitlab

nixos/services/mail/default.nix
```diff
@@ -156,7 +156,7 @@ in {
         certificateScheme = 3;
         enableImapSsl = true;
         enableManageSieve = true;
-        lmtpSaveToDetailMailbox = "no";
+        # lmtpSaveToDetailMailbox = "no";
```
dpausp: can we just remove the line?

dpausp (changes requested):

nixos/services/mail/default.nix
mkg20001: Option is removed by one of the nixos-mailserver fc commits, should we drop those we'd need to re-enable it, I think

But need to check again

mkg20001: Ah so option wasn't removed, this was just back when I pulled some random revision and had things break. We can leave it set to no.

dpausp:
@mkg20001 Change is ok, but the commit message is misleading now and needs a change

PR #330, opened 2021-07-21
mail: allow finegrained control of mail domains

mkg20001:
This changes several things
    
    - mailserver.domains is an attrset with a submodule
    - mailserver.domains = ["example.tld"] is transparently mapped to
    mailserver.domains."example.tld" = { enable = true; }
    - deperate domains as list, make attrset the default in docs
    - the user can now disable (enabled by default) autoconfig
    mailserver.domains."example.tld".autoconfig = false;
    - update docs to reflect the changes
    
    Note: this is fully backwards compatible, but using a list is now discouraged
    and deperacted
    
    fixes PL-130003

## Release process

Impact:
- none

Changelog:
- mailserver.domains is now an attrset
- mailserver.domains."domain".autoconfig/primary can now be used to change whether autoconfig is enabled/which domain is primary
- map mailserver.domains (list) to mailserver.domains (attrset) to be backwards compatible
- add message about mailserver.domains (list) being deperacted

## Security implications

  -  fix cert renewal caused by domain where autoconfig went unused, but was still enabled
  - tested on test vm

### DIFF
--- a/nixos/roles/mailserver.nix
+++ b/nixos/roles/mailserver.nix
@@ -1,6 +1,7 @@
 { config, lib, pkgs, ... }:
 
 with builtins;
+with lib;
 
 let
   params = lib.attrByPath [ "parameters" ] {} config.flyingcircus.enc;
@@ -17,7 +18,7 @@ let
     lib.hasAttrByPath [ "interfaces" "fe" ] params &&
     listenFe4 != [] && listenFe6 != []);
 
-  stdOptions = with lib; {
+  stdOptions = {
 
     mailHost = mkOption {
       type = types.str;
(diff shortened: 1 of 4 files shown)


nixos/roles/mailserver.nix
```diff
@@ -91,6 +92,7 @@ in
         Mailout on all nodes in this RG/location.
       '';
 
+      # shorthand for directly creating domains in
```
dpausp: comment not finished?

nixos/roles/mailserver.nix
```diff
@@ -99,9 +101,41 @@ in
...
+            autoconfig = mkOption {
+              description = "Enable autoconfig host for this domain";
+              default = true;
+              type = types.bool;
+            };
+          };
+        });
+
+        default = (listToAttrs (map (domain: (nameValuePair domain { enable = true; })) roles.mailserver.domains))
+          // (if roles.mailserver.domains != [] then { ${elemAt roles.mailserver.domains 0} = { enable = true; primary = true; }; } else {});
```
dpausp: use `optionalAttrs` here instead of if...else

nixos/roles/mailserver.nix
```diff
@@ -99,9 +101,41 @@ in
...
           various places.
+
+          Finegrained control of settings is possible by setting
+          <literal>flyingcircus.roles.mailserver.domain."domain.tld".autoconfig = false;</literal>
         '';
       };
 
+      # this allows finegrained control over each domain
+      # for example domain."test.fcio.net".autoconfig = false;
+      domain = mkOption {
```
dpausp: can this be set in JSON config?

nixos/services/mail/default.nix
```diff
@@ -16,7 +17,12 @@ let
   svc = config.flyingcircus.services.mail;
   fclib = config.fclib;
   primaryDomain =
-    if role.domains != [] then elemAt role.domains 0 else "example.org";
+    let firstPrimary =
+      elemAt (attrNames (filterAttrs (domain: config: config.enable && config.primary) role.domain)) 0;
```
dpausp: head instead of elemAt ... 0

nixos/roles/mailserver.nix
dpausp: head instead of elemAt ... 0

dpausp (changes requested):
We also need some documentation for the new options in doc/src/mailserver.rst.

nixos/roles/mailserver.nix
mkg20001:
```
{"domain":{"domain.tld":{"autoconfig":false}}}
```

Does it automatically take all options from JSON or anything specific necesarry?

mkg20001: Tested, seems so it does it automagically

dpausp:
@mkg20001 conflicts with your previous PR, please rebase

PR #331, opened 2021-07-21
mail: allow user to redirect FQDN hostname to imprint

mkg20001:
fixes PL-129598

## Release process

Impact:

Changelog:
- adds imprintUrl and imprintText to allow the user to specify either a bunch of HTML or a link to redirect the FQDN to so that t-online has an imprint to look at

## Security implications

  - imprint is served over tls
  - addSSL is set, tested on testvm

### DIFF
--- a/doc/src/mailserver.rst
+++ b/doc/src/mailserver.rst
@@ -48,6 +48,14 @@ which will be advertised as MX name on your mail domain. This host name (called
 **mailHost** from here on) must resolve to the FE addresses with both forward
 and reverse lookups.
 
+Additionally, some mail providers (namely [Telekom/T-Online](https://postmaster.t-online.de/#t4.1))
+may require that your mailserver has an imprint served at its hostname.
+
+For this you can either set imprintUrl to the location of your existing imprint,
+or use imprintText to specify an imprint in HTML format
+
+Note that it is not possible to set both imprintUrl and imprintText and imprint cannot be used if you serve webmail under the mailHost (meaning mailHost and webmailHost cannot be the same)
+
 .. warning::
 
   Incorrect DNS setup is the most frequent source of delivery problems. Let our
@@ -68,7 +76,8 @@ and *test2.fcio.net*::
       "test2.fcio.net": {
         "autoconfig": false
       }
-    }
+    },
+    "imprintUrl": "your-company.tld/imprint"
   }
 
 .. note::
--- a/nixos/roles/mailserver.nix
+++ b/nixos/roles/mailserver.nix
@@ -151,6 +151,24 @@ in
           else v);
       };
 
+      imprintUrl = mkOption {
+        type = with types; nullOr str;
+        description = ''
+          Webaddress of your imprint to be in compilance with T-Online Postmaster Rules
+          You can instead provide your imprint in text form using the imprintText option
+        '';
+        default = null;
+      };
+
+      imprintText = mkOption {
+        type = with types; nullOr str;
+        description = ''
+          Imprint as HTML text to be in compilance with T-Online Postmaster Rules
+          You can instead provide the webaddress to your imprint using the imprintUrl option
+        '';
+        default = null;
+      };
+
       webmailHost = mkOption {
         type = with types; nullOr str;
         description = "(Virtual) host name of the webmail service.";
--- a/nixos/services/mail/default.nix
+++ b/nixos/services/mail/default.nix
@@ -113,6 +113,23 @@ in {
     let
       fqdn = with config.networking; "${hostName}.${domain}";
     in {
+      assertions = [
+        # at least one needs to be null
+        { assertion = role.imprintUrl == null || role.imprintText == null;
+          message = ''
+            The options imprintUrl and imprintText are mutually exclusive
+            Remove one to fix this issue
+          '';
+        }
+        # either not equal or no imprint options used
+        { assertion = role.mailHost != role.webmailHost || (role.imprintUrl == null && role.imprintText == null);
+          message = ''
+            imprintUrl/imprintText cannot be set when webmail is already being served under ${role.mailHost}
+            Change webmailHost or remove imprint
+          '';
+        }
+      ];
+
       environment = {
         etc = {
           "local/mail/dns.zone".text =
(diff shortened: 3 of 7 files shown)


nixos/roles/mailserver.nix
```diff
@@ -102,6 +102,24 @@ in
...
+          Webaddress of your imprint to be in compilance with T-Online Postmaster Rules
+          You can instead provide your imprint in text form using the imprintText option
+        '';
+        default = null;
+      };
+
+      imprintText = mkOption {
+        type = with types; nullOr str;
+        description = ''
+          Imprint as HTML text to be in compilance with T-Online Postmaster Rules
```
ctheune, line 166: Verweis/Link auf postmaster.t-online.de und noch zum TMG wo die Anforderungen an die  Anbieterkennzeichnung beschrieben sind.

nixos/roles/mailserver.nix
```diff
@@ -102,6 +102,24 @@ in
         '';
       };
 
+      imprintUrl = mkOption {
```
ctheune, line 154: Wir sollten das zur Mindestanforderung machen: es muss eins gesetzt sein (aber es dürfen nicht beide gesetzt sein).

ctheune (changes requested):

nixos/roles/mailserver.nix
mkg20001: Hat mir wer ein guten link den ich da verlinken soll zum TMG?

nixos/roles/mailserver.nix
mkg20001:
```
error: 
Failed assertions:
- The imprintUrl or imprintText option needs to be set

This is mandated by:
- Telekom Postmaster Rules §4.1 (see https://postmaster.t-online.de/#t4.1), for sending mail to Telekom customers
- Telemediengesetz Anbieterkennzeichnung (see #todo#), for compilance with German law

(use '--show-trace' to show detailed location information)
ERROR: Building channel failed!
```

nixos/roles/mailserver.nix
dpausp: https://www.gesetze-im-internet.de/tmg/__5.html ?

nixos/services/mail/default.nix
```diff
@@ -86,6 +86,21 @@ in {
     let
       fqdn = with config.networking; "${hostName}.${domain}";
     in {
+      warnings = optional (role.imprintUrl != null || role.imprintText != null) ''
```
dpausp: optional is not defined here

nixos/services/mail/default.nix
```diff
@@ -86,6 +86,21 @@ in {
...
+        The imprintUrl or imprintText option needs to be set
+
+        This is mandated by:
+        - Telekom Postmaster Rules §4.1 (see https://postmaster.t-online.de/#t4.1), for sending mail to Telekom customers
+        - Telemediengesetz Anbieterkennzeichnung (see https://www.gesetze-im-internet.de/tmg/__5.html), for compilance with German law
+
+        Please do it as soon as possible
+      '';
+      assertions = [
+        { assertion = role.imprintUrl != null && role.imprintText != null;
```
dpausp: the assertion must be turned around: at the moment it fails when none or one of the options is set.

nixos/services/mail/default.nix
```diff
@@ -200,7 +215,27 @@ in {
               inherit (role) mailHost webmailHost;
             });
           } config.fclib.network.fe.dualstack.addressesQuoted);
-        in listToAttrs (map cfgForDomain role.domains);
+        in listToAttrs ((map cfgForDomain role.domains) ++
+          [ (lib.nameValuePair fqdn
```
dpausp: This must point to the `role.mailHost`, not `fqdn` here. (it's a bit misleading in the code because `mailserver.fqdn = role.mailHost` but there's a `fqdn` in a let that points to the internal hostname like test33.fcio.net).

After fixing this, there will be a problem with nixos/services/mail/roundcube.nix, line 51 which redirects `role.mailHost` to the webmailer. This has to be removed. Also, we must assert that `role.webmailHost != role.mailHost` to avoid a conflict between imprint and webmailer...

Still an open question: we already have some servers that use webmailHost == mailHost, what to do here?

nixos/services/mail/default.nix
```diff
@@ -200,7 +215,27 @@ in {
...
+              (lib.mkIf (role.imprintUrl != null) {
+                globalRedirect = role.imprintUrl;
+              })
+              (lib.mkIf (role.imprintText != null) {
+                root = with pkgs; writeTextDir "index.html" role.imprintText;
+              })
+              (lib.mkIf (role.imprintUrl == null && role.imprintText == null) {
+                root = with pkgs; writeTextDir "index.html" ''
+                  <h1>Imprint missing</h1>
+                  <h2>Imprint can be configured using <tt>flyingcircus.roles.mailserver.imprintText</tt> or <tt>flyingcircus.roles.mailserver.imprintUrl</tt></h2>
```
dpausp: A link to the documentation at https://doc.flyingcircus.io/roles/fc-21.05-production/mailserver.html#how-do-i-perform-a-basic-setup would be helpful.

nixos/services/mail/default.nix
```diff
@@ -86,6 +86,21 @@ in {
     let
       fqdn = with config.networking; "${hostName}.${domain}";
     in {
+      warnings = optional (role.imprintUrl != null || role.imprintText != null) ''
+        The imprintUrl or imprintText option needs to be set
+
+        This is mandated by:
+        - Telekom Postmaster Rules §4.1 (see https://postmaster.t-online.de/#t4.1), for sending mail to Telekom customers
```
dpausp: This should in the documentation, too.

dpausp (changes requested):
See comments. Also needs a rebase, I changed the documentation in the meantime.

nixos/services/mail/default.nix
dpausp: New plan: keep the redirect to webmailHost as platform default, but it can be overridden by setting imprintText or imprintUrl. Both options should be optional and not generate a warning. If mailHost == webmailHost, both options should produce an error.
We have to document this properly in the mailserver role docs.

nixos/services/mail/default.nix
```diff
@@ -104,6 +104,21 @@ in {
     let
       fqdn = with config.networking; "${hostName}.${domain}";
     in {
+      assertions = [
+        { assertion = !(role.imprintUrl != null && role.imprintText != null);
```
dpausp: The double negation is hard to read, this looks better with an `||`

doc/src/mailserver.rst
```diff
@@ -48,6 +48,12 @@ which will be advertised as MX name on your mail domain. This host name (called
 **mailHost** from here on) must resolve to the FE addresses with both forward
 and reverse lookups.
 
+Additionally, some mail providers (namely [Telekom/T-Online](https://postmaster.t-online.de/#t4.1))
+require that your mailserver has an imprint served at it's hostname.
+
+For this you can either set imprintUrl to the location of your existing imprint,
+or use imprintText to specify an imprint in HTML format
+
```
dpausp, line 56: Tell the users about the webmailHost behaviour: what's the default and what will lead to an error (assertions)?

doc/src/mailserver.rst
```diff
@@ -48,6 +48,12 @@ which will be advertised as MX name on your mail domain. This host name (called
 **mailHost** from here on) must resolve to the FE addresses with both forward
 and reverse lookups.
 
+Additionally, some mail providers (namely [Telekom/T-Online](https://postmaster.t-online.de/#t4.1))
+require that your mailserver has an imprint served at it's hostname.
```
dpausp: we are not really sure what they want, I'd say `may require`

`it's` -> `its`

nixos/services/mail/default.nix
```diff
@@ -104,6 +104,21 @@ in {
...
+          message = ''
+            The options imprintUrl and imprintText are mutually exclusive
+            Remove one to fix this issue
+          '';
+        }
+        { assertion = !((role.imprintUrl != null && role.imprintText != null) && (role.mailHost == role.webmailHost));
```
dpausp: The first part is true if both options are set which isn't possible. Avoid double negation here as well.

nixos/services/mail/default.nix
```diff
@@ -104,6 +104,21 @@ in {
...
+          message = ''
+            imprintUrl/imprintText cannot be set when webmail ${role.webmailHost} is already being served under ${role.mailHost}
```
dpausp: I'd remove the `${role.webmailHost}` because it's the same as `${role.mailHost}` which looks strange in the final output.

nixos/services/mail/default.nix
```diff
@@ -218,7 +233,27 @@ in {
               inherit (role) mailHost webmailHost;
             });
           } config.fclib.network.fe.dualstack.addressesQuoted);
-        in listToAttrs (map cfgForDomain (attrNames (filterAttrs (domain: config: config.enable && config.autoconfig) role.domains)));
+        in listToAttrs ((map cfgForDomain (attrNames (filterAttrs (domain: config: config.enable && config.autoconfig) role.domains))) ++
+          (optional (role.imprintUrl != null || role.imprintText != null ) (lib.nameValuePair role.mailHost
+            (lib.mkMerge [
+              {
```
dpausp, line 251: listen addresses must be set the the frontend IPs here or Nginx will prefer the autoconfig vhosts (they are already listening on fe only)

dpausp (changes requested):

mkg20001:
Following problem:

module system gives us the values without priorities, just the clean, resolved values

we need the values with their priorities so when doing virtualHosts = fc.virtualHosts the module system knows which were defaults and which were user-set values from fc.virtualHosts

I have no idea how to do that

Alternative is to use options.flyingcircus.services.nginx.virtualHosts.definitions (raw user-definied keys), use modules.mergeDefinitions and go through each of the nameValuePairs, applying the FC changes like with the JSON previously.

also noticed other stuff:
- currently listenAddress and listenAddress6 default to 0.0.0.0 and [::] respectivly, meaning that the default that should use frontend addresses for listenAddresses will never trigger. I think that's wrong and all fc.virtualHosts should use FE addresses by default - correct?

PR #334, opened 2021-07-26
Remove `cipher` option from openvpn

nichmoe:
This option is deprecated for server-client mode (see openvpn man page).
   #PL-129937

## Release process

Impact: May require restart of openvpn.

Changelog: Remove --cipher option in openvpn server and client config.

## Security implications

     - Deprecated options should not be used. 

    - Openvpn test exited without error.

### DIFF
--- a/nixos/roles/external_net/openvpn.nix
+++ b/nixos/roles/external_net/openvpn.nix
@@ -179,9 +179,6 @@ let
     plugin ${openvpn}/lib/openvpn/plugins/openvpn-plugin-auth-pam.so openvpn
     management localhost ${mgmPort} ${pkgs.writeText "openvpn-mgm-psk" mgmPsk}
 
-    # Since 21.05 cipher must be set. Previously it defaulted to BF-CBC as fallback
-    cipher AES-256-GCM
-
     user nobody
     group nogroup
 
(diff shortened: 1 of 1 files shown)


nixos/roles/external_net/openvpn.nix
```diff
@@ -179,9 +179,6 @@ let
     plugin ${openvpn}/lib/openvpn/plugins/openvpn-plugin-auth-pam.so openvpn
     management localhost ${mgmPort} ${pkgs.writeText "openvpn-mgm-psk" mgmPsk}
 
-    # Since 21.05 cipher must be set. Previously it defaulted to BF-CBC as fallback
-    cipher AES-256-GCM
```
dpausp, line 183: I think we should use `data-ciphers-fallback alg` with the first value from our allowedCiphers list instead to have a secure default

nichmoe: `data-cipher-fallback alg` seems to be a [deprecated option](https://community.openvpn.net/openvpn/wiki/CipherNegotiation?__cf_chl_jschl_tk__=pmd_cb4e1091d9b1079d188efd5486043fa932a0f3a9-1627625809-0-gqNtZGzNAfijcnBszQbi#Specialrequirement:OpenVPNbuiltwith--enable-small). A secure default should already come from `allowedCyphers`.

PR #337, opened 2021-08-05
Mailserver: fix legacy domain config, Roundcube listen on FE only

dpausp:
The legacy domains options broke with more than one list items because
primary wasn't set explicitly for the secondary domains.

Roundcube (webmailHost) now listens on FE only to avoid that Nginx
prefers the autoconfig virtualhosts which use FE and are thus more
specific. Nginx matches by IP first and not by server name!

 #PL-130021

## Release process

Impact:

Changelog:

* Mailserver: fix domains option when multiple mail domains are given in a list (#PL-130021).

## Security implications

  - no new requirements here, fixes broken domain config and unreachable webmailer  
  - checked on test mail server that nginx config is generated properly, webmail is reachable and more than one domain works with legacy config
  - automated test still runs

### DIFF
--- a/nixos/roles/mailserver.nix
+++ b/nixos/roles/mailserver.nix
@@ -143,7 +143,10 @@ in
               }
           ''
           recursiveUpdate
-            (listToAttrs (map (domain: (nameValuePair domain { enable = true; autoconfig = true;})) v))
+            (listToAttrs (map (domain: (nameValuePair domain {
+              enable = true;
+              autoconfig = true;
+              primary = false; })) v))
             (optionalAttrs (v != []) { "${head v}" = { primary = true; }; })
           else v);
       };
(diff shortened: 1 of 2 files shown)


mkg20001 (approved):
LGTM

mkg20001:
Note that this is not allowing the user to change the setting. Another way that also works is to specify `services.nginx.virtualHosts.${...` where currently the attrset with { forceSSL is and set forceSSL on the virtualHost

dpausp:
> Note that this is not allowing the user to change the setting. Another way that also works is to specify `services.nginx.virtualHosts.${...` where currently the attrset with { forceSSL is and set forceSSL on the virtualHost

I think it's ok for the use case here as it is.

PR #339, opened 2021-08-10
PL-130029 fc agent structlog and refactoring

dpausp:
* Python dev env, Python 3.9
  * Package dependencies that only work on Linux are added conditionally so it can be built on MacOS now.
* fc-agent: format Python code with yapf, Yapf settings were taken from the backy project.
* factor out reusable system management code and clean up manage.py a bit
* adapt structlog logging code for console/file from fc-qemu
* structlog to systemd journal renderer/logger
* use structlog in manage.py and nixos.py
* output from nix commands/switch-to-configuration is written line-by-line
  to separate files in /var/log/fc-agent:
  * interactive uses the same file /var/log/fc-agent/build-output.log
    and overwrites it with each run
  * when called from a systemd unit, files with unique names (date +
    systemd invocation id) are kept if something has changed for 180 days
* /var/log/fc-agent.log logs everything (mainly for agent debugging
  purposes)
* test important functionality from nixos.py and Channel.prepare_maintenance
* current fc environment name is now written to /etc/fcio_environment_name

 #PL-130029

## Release process

Impact:

Changelog:

* fc-agent: add structured logging output for better error reporting and more insights into what automated updates are doing. This also changes the output of the fc-manage command when run manually to give better user feedback. Structured logging output can be explored with journalctl or graylog (loghost) (#PL-130029).

## Security implications

  - have proper logging and error reporting to find problems early
  - tested on multiple test VMs with and without production flag
  - automatic NixOS tests still run (except ceph test where Python tests break with an fc.util import error)

### DIFF
--- a/pkgs/fc/agent/setup.py
+++ b/pkgs/fc/agent/setup.py
@@ -10,14 +10,12 @@ here = path.abspath(path.dirname(__file__))
 with open(path.join(here, 'README.rst'), encoding='utf-8') as f:
     long_description = f.read()
 
-
 test_deps = [
     'freezegun>=0.3',
     'responses',
     'pytest>=3',
     'pytest-cov',
 ]
-
 
 setup(
     name='fc.agent',
--- a/pkgs/fc/agent/setup.py
+++ b/pkgs/fc/agent/setup.py
@@ -14,6 +14,8 @@ test_deps = [
     'freezegun>=0.3',
     'responses',
     'pytest>=3',
+    'pytest-mock',
+    'pytest-structlog',
     'pytest-cov',
 ]
 
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ share/
 target/
 tmp/
 /*.sublime-*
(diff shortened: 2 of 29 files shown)


dpausp:
ceph test breaks at the moment because it cannot import fc.util from the agent (Python 3.8/3.9 problem?)

pkgs/fc/agent/setup.py
```diff
@@ -10,14 +10,14 @@
...
     'freezegun>=0.3',
     'responses',
     'pytest>=3',
-    'pytest-cov',
-]
-
+    'pytest-mock',
+    'pytest-structlog',
+    'pytest-subprocess',
+    'pytest-cov',]
```
mkg20001: such formatting defeats the purpose of the trailing comma, which is to prevent merge conflicts
the ] should be on another line imo

dpausp:
Agent uses now Python 3.8 to make it easier. We have to upgrade all our Python packages at once to avoid problems.

PR #340, opened 2021-08-13
PL 129843 Nginx should be fully configurable via Nix code

mkg20001:
**upstream sync moved to PR #349**

## Release process

Impact:

Changelog:
- flyingcircus nginx configurations can now be written in pure nix language now

## Security implications

  - no security related changes

### DIFF
--- a/nixos/services/nginx/default.nix
+++ b/nixos/services/nginx/default.nix
@@ -4,7 +4,7 @@
 # This may also affect other services that use reloading.
 { lib, config, pkgs, ... }:
 
-with builtins;
+with builtins; with lib;
 
 let
   cfg = config.flyingcircus.services.nginx;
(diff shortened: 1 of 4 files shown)


nixos/services/nginx/default.nix
```diff
@@ -101,7 +102,7 @@ let
 
   baseHttpConfig = ''
     # === Defaults ===
-    default_type application/octet-stream;
+    # default_type application/octet-stream;
```
dpausp: Do we need this?

dpausp (changes requested):
In the JSON config, setting an *SSL option automatically enables enableACME. We want to keep that. Doesn't work with the new NixOS options so the example config doesn't build anymore.

nixos/services/nginx/default.nix
mkg20001: Not really, removing

mkg20001:
Indeed strange, as https://github.com/flyingcircusio/fc-nixos/pull/340/files#diff-f5fb9002fd859a9923715a59f5f83653e6baf9675b48013b8dc5bd567661ec75L73 should've enabled it.

mkg20001:
image

Trying the example config from the README.txt results in acme jobs for the domain - couldn't reproduce

mkg20001:
Tried with the nix version, that one indeed causes the problem. The issue is that `//` doesn't properly work with things from the module system, using lib.mkMerge causes other trouble (we have to override using mkForce, which means that if the user specifies enableACME themselves to false it will be overwritten, too)

mkg20001:
Fixed. Now using the module system properly and overriding defaults for both enableACME and listenAddresses with what the fc function previously did.

PR #361, opened 2021-09-07
fc-agent, garbage-collect: limit IO and CPU usage via unit settings

dpausp:
We experienced some load alerts for short peaks caused by garbage
collection or the agent. This is often caused by iowait and it seems like
that VMs that are mostly idle are affected more than others.

Restrict the number of IOPS on cgroup level depending on the limit for the
VM to avoid requesting much more IOPS for maintenance tasks than the VM can deliver.
The agent gets more because it shouldn't take forever. For the garbage
collection, there's a more restrictive setting because time doesn't
matter here and the timeout is set to infinity now.

We also set other IO and CPU-related unit settings to make sure that regular user
processes have higher priority than our maintenance.

Drop the limitation on how much fc-collect-garbage can collect in one
run because the other limits should prevent problems.

 #PL-130083

## Release process

Impact:

Changelog:

* fc-agent and fc-garbage-collect: apply CPU/IO limitations and low priority settings via SystemD units to avoid load peaks and resulting alerts. This is often caused by iowait on HDD VMs. We tried to limit the impact before by using ionice and by limiting the amount of collected garbage but we still had problems on mostly idle machines or when agent and garbage collection ran at the same time (#PL-130083).  

## Security implications

  - avoid alarm noise
  - prioritize regular user processes over maintenance tasks 
  - manually checked that the settings are generated and effective on a test VM
  - (doesn't apply to automated tests)

### DIFF
--- a/nixos/infrastructure/flyingcircus-virtual.nix
+++ b/nixos/infrastructure/flyingcircus-virtual.nix
@@ -7,9 +7,10 @@ let
   ioScheduler =  if (cfg.infrastructure.preferNoneSchedulerOnSsd && (cfg.enc.parameters.rbd_pool == "rbd.ssd"))
                  then "none"
                  else "bfq";
+  maxIops = attrByPath [ "parameters" "iops" ] 250 cfg.enc;
 in
 mkIf (cfg.infrastructureModule == "flyingcircus") {
- 
+
   boot = {
     initrd.kernelModules = [
       "virtio_blk"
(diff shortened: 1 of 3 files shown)


nixos/infrastructure/flyingcircus-virtual.nix
```diff
@@ -77,10 +78,34 @@ mkIf (cfg.infrastructureModule == "flyingcircus") {
...
+        in {
+          IOReadIOPSMax = vdaIopsMax;
+          IOWriteIOPSMax = vdaIopsMax;
+        };
+
+      fc-collect-garbage.serviceConfig =
+        let
+          # Must not consume more than 12.5% of available IOPS.
+          # We don't have a problem with this taking really long.
+          vdaIopsMax = "/dev/vda ${toString (maxIops / 8)}";
```
ctheune, line 104: can we make the comment and code a bit more aligned? I'd prefer `maxIops * 0.125` here ... ?

nixos/infrastructure/flyingcircus-virtual.nix
```diff
@@ -77,10 +78,34 @@ mkIf (cfg.infrastructureModule == "flyingcircus") {
...
+        script = "echo \"$(date) -- SERIAL CONSOLE IS LIVE --\" > /dev/ttyS0";
+      };
+
+      fc-agent.serviceConfig =
+        let
+          # Must not consume more than 75% of available IOPS.
+          # The systemd setting is split between read and write but
+          # we only have one IOPS limit for the VM so combined we
+          # could get requests that are over the VM limit.
+          vdaIopsMax = "/dev/vda ${toString (maxIops - maxIops / 4)}";
```
ctheune, line 94: How about maxIops * 0.75?

ctheune (changes requested):

nixos/infrastructure/flyingcircus-virtual.nix
dpausp: Couldn't find a simple way to do multiplication and get an integer for the config string. Division yields an integer so that was the easiest way to do it.

nixos/infrastructure/flyingcircus-virtual.nix
dpausp: Couldn't find a simple way to do multiplication and get an integer for the config string. Division yields an integer so that was the easiest way to do it.

ctheune (approved):

PR #373, opened 2021-09-20
htop: don't show userland threads by default

mkg20001:

## Release process

Impact: none

Changelog:
- make htop hide userland threads by default

## Security implications

  - not touching anything critical, no reqs

### DIFF
--- a/pkgs/hide-userland.patch
+++ b/pkgs/hide-userland.patch
@@ -1,0 +1,24 @@
+From 27b970cd23610bfbdacb00873e15caa758052b09 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Maciej=20Kr=C3=BCger?= <mkg20001@gmail.com>
+Date: Thu, 15 Jul 2021 09:41:07 +0200
+Subject: [PATCH] Hide userland threads by default
+
+---
+ Settings.c | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/Settings.c b/Settings.c
+index ca8253e..a8cdfa7 100644
+--- a/Settings.c
++++ b/Settings.c
+@@ -461,6 +461,7 @@ Settings* Settings_new(unsigned int initialCpuCount) {
+    if (!ok) {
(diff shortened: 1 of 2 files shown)


ctheune:
Thanks for splitting this of. I took some more time to review this and this actually does touch a requirement: see our Systems Development Guideline, section 1.1:

> Changes to existing (external) software packages need special consideration.A.14.2.4 The decision should be backed by a senior developer or Management. Changes should also be feed back upstream so they become part of a regular release.

Also, there was extensive discussion about this in the upstream project and I'm not sure we want to go against their own decisions about their tools' usability:

https://github.com/htop-dev/htop/issues/99

Do you still think hiding them by default is a good idea? I can't remember right now whether I was asking to remove them but I feel cured now and would be more than happy to just leave them as is.

mkg20001:
The initial motivation to do this on my own systems is the fact that toggling this option every time on every machine takes some time

Really depends on who is using htop. If it's mostly the customers than it should be on (as is now), but if it's mostly the admins and they know about the userland setting and what it does _and_ they frequently find themselves toggling it off then it makes more sense to just keep it on this patch should be the way to go

ctheune:
Additionally, I think we should be able to use a simple config file instead of patch ... no?

> If no user configuration exists htop
       tries to read the system-wide configuration from
       ${prefix}/etc/htoprc and as a last resort, falls back to its hard
       coded defaults.

nichmoe:
My personal opinion as someone who uses htop on occasion is that the change is not really necessary. Someone could make the change to their config and that's it. 
But regardless of if we should do it at all, I think setting some system-wide configuration as theune suggested is cleaner, because the patch just changes the default config anyway. Also I'm not quite sure, but doesn't it mean that we now would have to compile htop our self because we added a patch to the package definition?

frlan:
I never had any idea to change this setting for daily use at all.

dpausp:
we decided not to do that but I learned about `Shift-H` to hide the threads ;)

PR #406, opened 2021-10-28
agent: add nopasswd rule to allow service users to use sudo fc-manage properly again

mkg20001:
fixes PL-130171

## Release process

Impact:

Changelog:
- set nopasswd for fc-manage sudo rule

## Security implications

  - unsure allowing fc-manage with nopasswd, but without it doesn't really make sense as service users don't really have passwords

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -83,7 +83,8 @@ in {
 
       flyingcircus.passwordlessSudoRules = [
         {
-          commands = [ "${pkgs.fc.agent}/bin/fc-manage" ];
+          # we need NOPASSWD here as service users have no password
+          commands = [ { command = "${pkgs.fc.agent}/bin/fc-manage"; options = [ "NOPASSWD" ]; } ];
           groups = [ "sudo-srv" "service" ];
         }
       ];


nixos/platform/agent.nix
```diff
@@ -83,7 +83,8 @@ in {
 
       flyingcircus.passwordlessSudoRules = [
         {
-          commands = [ "${pkgs.fc.agent}/bin/fc-manage" ];
+          # we need NOPASSWD here as service users have no password
+          commands = [ { command = "${pkgs.fc.agent}/bin/fc-manage"; options = [ "NOPASSWD" ]; } ];
           groups = [ "sudo-srv" "service" ];
```
mkg20001, line 88: not sure if service should get NOPASSWD and sudo-srv should get PASSWD

dpausp:
The option adds passwordless sudo rules automatically, so this change should have any effect. I can already see on an unchanged test system:

`%service        ALL=(ALL:ALL)   NOPASSWD: /nix/store/79rm3z0j91mpf9n4b2kagsa3yskzqycd-python3.8-fc-agent-1.0/bin/fc-manage`

mkg20001:
Seems like the problem is postgres not being in service group, not broken sudo roles

PR #407, opened 2021-10-30
mysql: enable slow log

mkg20001:

## Release process

Impact:
- restarts mysql

Changelog:
- enable slow log in mysql

fixes PL-100075

## Security implications

  - log queries that may impact performance, must only be readable by mysql and service users as it may contain sensitive data
  - automated test still runs, checked that log files are created with the correct permissions

### DIFF
--- a/nixos/roles/mysql.nix
+++ b/nixos/roles/mysql.nix
@@ -116,6 +116,20 @@ with builtins;
       }
     ];
 
+    systemd.tmpfiles.rules = [
+      "d /var/log/mysql 0755 mysql service -"
+      "f /var/log/mysql/mysql.slow 0640 mysql service -"
+    ];
+
+    services.logrotate.paths.mysql-slowlog = {
+      path = "/var/log/mysql/mysql.slow";
+      extraConfig = ''
+        rotate 2
+        create 0640 mysql service
+        missingok
+      '';
+    };
+
     services.percona = {
       enable = true;
       inherit package rootPasswordFile;
@@ -142,8 +156,16 @@ with builtins;
         default_storage_engine     = InnoDB
         table_definition_cache     = 512
         open_files_limit           = 65535
-        sysdate-is-now             = 1
+        sysdate-is-now             = ON
         sql_mode                   = NO_ENGINE_SUBSTITUTION
+
+        log_slow_verbosity = 'full'
+        slow_query_log = ON
+        long_query_time = 0.1
+        log_slow_slave_statements = ON
+        slow_query_log_file = '/var/log/mysql/mysql.slow'
+        log_slow_admin_statements = ON
+        log_queries_not_using_indexes = ON
 
         init-connect               = 'SET NAMES ${charset} COLLATE ${collation}'
         character-set-server       = ${charset}


nixos/roles/mysql.nix
```diff
@@ -145,6 +159,12 @@ with builtins;
         sysdate-is-now             = 1
         sql_mode                   = NO_ENGINE_SUBSTITUTION
 
+        log_slow_verbosity = 'full'
```
mkg20001, line 162: this is copied as-is from the wiki, not sure if this much loging is desirable

nixos/roles/mysql.nix
```diff
@@ -116,6 +116,20 @@ with builtins;
       }
     ];
 
+    systemd.tmpfiles.rules = [
+      "d /var/log/mysql 0755 mysql mysql -"
+      "f /var/log/mysql/mysql.slow 0644 mysql mysql -"
```
dpausp: The file shouldn't be readable by everybody, just service users, I suggest 0640 mysql service

nixos/roles/mysql.nix
```diff
@@ -116,6 +116,20 @@ with builtins;
...
+    ];
+
+    services.logrotate.extraConfig = ''
+      /var/log/mysql/mysql.slow
+      {
+          rotate 2
+          create 0644 mysql mysql
```
dpausp: change permissions here, too

nixos/roles/mysql.nix
```diff
@@ -145,6 +159,12 @@ with builtins;
         sysdate-is-now             = 1
         sql_mode                   = NO_ENGINE_SUBSTITUTION
 
+        log_slow_verbosity = 'full'
+        slow_query_log = ON
+        long_query_time = 0
```
dpausp: We should use 100ms like for Postgresql to avoid logging too many small queries.

Is it easy to override these values via Nix code? Otherwise we should add options to change the query time and verbosity.

dpausp (changes requested):

nixos/roles/mysql.nix
mkg20001: percona would need a patch to use rfc42-style options then any of those values could be easily changed. I'm going to add that, long-term that's better than inventing our own options

mkg20001: Is percona just mysql but slightly different? We could just use upstream, it has rfc42 already

mkg20001: I've looked through things and it seems like the upstream module has some of the fc changes missing, most notably rootPasswordFile

Not sure but inventing an option versus going upstream and not being able to modify it for now both don't seem like good choices, which one then..

ctheune: Yeah, we should aim for unifying the percona and the upstream MySQL stuff, getting the rootPasswordFile upstream would be welcome too, but let's do that in a separate issue.

nixos/roles/mysql.nix
```diff
@@ -145,6 +159,12 @@ with builtins;
         sysdate-is-now             = 1
         sql_mode                   = NO_ENGINE_SUBSTITUTION
 
+        log_slow_verbosity = 'full'
+        slow_query_log = ON
+        long_query_time = 100
```
ctheune: This means 100 seconds.

Also, we should `log_slow_admin_statements` and `log_queries_not_using_indexes`.

ctheune (changes requested):

nixos/roles/mysql.nix
dpausp: We can override default settings per flyingcircus.roles.mysql.extraConfig or /etc/local/mysql. They appear later in the generated config file and have precedence.

PR #413, opened 2021-11-09
kernel: experimental backport for XFS CIL bug, enable all sysrq keys

ctheune:

## Release process

Impact:

* VMs will reboot due to kernel update

Changelog:

## Security implications

sysrq keys are potentially security sensitive, but I've been expecting them to be enabled already. Only root and hypervisor users can access this (through /proc/sys/kernel/sysrq and the localhost-bound qemu monitor interface).

nothing to test really as the sysrq already were there but not all of them were enabled


ctheune:
Waiting for the actual backport that should make it to 5.10 mainline.

PR #433, opened 2021-12-03
mysql: rotate slow log by size, compress, fix log reopen

dpausp:
* Rotates when file size is greater than 2GB, at least weekly.
* Added postrotate command to make mysql reopen log properly

 #PL-130221

## Release process

Impact:

Changelog:

* MySQL: rotate slow log when file size is greater than 2GB and compress, at least once a week (#PL-130221).

## Security implications

  - limit size of mysql slow log to avoid running out of disk space  
  - checked on a test VM that logrotate is configured correctly (`logrotate -v`), rotates files larger than maxsize and that mysql writes to the new file

### DIFF
--- a/nixos/roles/mysql.nix
+++ b/nixos/roles/mysql.nix
@@ -131,15 +131,26 @@ in
     ];
 
     systemd.tmpfiles.rules = [
-      "d /var/log/mysql 0755 mysql service -"
+      "d /var/log/mysql 0755 mysql service 90d"
       "f /var/log/mysql/mysql.slow 0640 mysql service -"
     ];
 
+    # Note that this does not use our platform defauls for logrotate.
+    # Configuration via logrotate.paths is added before our defaults so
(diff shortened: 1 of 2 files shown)


nixos/roles/mysql.nix
```diff
@@ -131,15 +131,26 @@ in
     ];
 
     systemd.tmpfiles.rules = [
-      "d /var/log/mysql 0755 mysql service -"
+      "d /var/log/mysql 0755 mysql service - 90d"
```
ctheune: This doesn't look right. I think the '-' needs to go away as that's the field for the cleanup age?

man tmpfiles.d says:

```
       #Type Path                                     Mode User Group Age         Argument
       d     /directory/to/create-and-cleanup         mode user group cleanup-age -
```

ctheune (changes requested):

PR #436, opened 2021-12-07
Ceph: wrap up NixOS integration, lots of fixes and add test coverage

ctheune:

## Release process

Impact:

none

Changelog:

* Our Ceph infrastructure has been fully ported to NixOS now. (PL-127635)
 
## Security implications

No specific new requirements have been found as this is pretty straight 1:1 port of the Gentoo-based management features with a number of simplifications.

I have improved the key management that now includes purging old keys, which we didn't do before. It's not perfect yet, but there's a ticket in youtrack that follows up on the remaining ideas.

I have also automated the creation of firewall rules to grant access to the RGW servers from within the SRV network.

There is also new functionality for running explicit enter/exit scripts before and after maintenances which is a generic feature for the whole platform. service users can register those scripts through the existing system config mechanism and run them as root, however, that's a known and intentional vector anyways.

All of the functions have been tested manually and a good part is now also exercised by Hydra tests.

### DIFF
--- a/nixos/infrastructure/flyingcircus-physical.nix
+++ b/nixos/infrastructure/flyingcircus-physical.nix
@@ -27,6 +27,10 @@ mkIf (cfg.infrastructureModule == "flyingcircus-physical") {
         fsIdentifier = "provided";
         gfxmodeBios = "text";
       };
+
+      # Wanted by backy and Ceph servers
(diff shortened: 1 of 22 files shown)


dpausp:
looks good to me, reviewed with theuni. Can be merged when tests pass on Hydra.

PR #441, opened 2021-12-14
[20.09] elasticsearch: set -Dlog4j2.formatMsgNoLookups=true

dpausp:
Elasticsearch is not susceptible to remote code execution
via the log4j2 CVE-2021-44228 but we should close this attack vector anyways.

 #PL-130251

## Release process

Impact:

* Elasticsearch will be restarted.

Changelog:

* Elasticsearch: fix log4j2 CVE-2021-44228 by setting `-Dlog4j2.formatMsgNoLookups=true`. Without this, Elasticsearch is susceptible to a minor information leak about the system environment. Remote code execution was never possible via Elasticsearch (#PL-130251).

## Security implications

  - prevent information leaks about the system environment
  - we use the measures recommended by Elasticsearch to fix the issue 
  - checked manually on test VM if ES works with that and the flag is present
  - automated tests still run

### DIFF
--- a/nixos/roles/elasticsearch.nix
+++ b/nixos/roles/elasticsearch.nix
@@ -149,6 +149,7 @@ in
         # Appending the next two lines overrides the former.
         "-Xms${toString esHeap}m"
         "-Xmx${toString esHeap}m"
+        "-Dlog4j2.formatMsgNoLookups=true"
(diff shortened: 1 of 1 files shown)


dpausp:
We already set the flag on some VMs. Having it twice in the command line doesn't hurt, we can remove the manual setting later.

PR #449, opened 2022-01-04
Matomo 4.6.2, geoip2 fix, fix URL for archive processing

dpausp:

## Release process

Impact:

Changelog:

* Matomo: update to 4.6.2 and various fixes (#PL-130234):
  * The GeoIP2 plugin works now (it's active by default). It failed previously because it couldn't write the database file.
  * Fix the periodic archive processing task. Setting `services.matomo.hostname` also changes the URL used for archive processing. 
  * Don't require a local mysql database in the systemD unit as Matomo can be configured to use databases on other machines. 

## Security implications

  - use a recent version of matomo
  - checked if updating from 4.5.0 works on a test VM, also tested starting from scratch
  - checked Matomo release notes for critical changes

### DIFF
--- a/nixos/services/default.nix
+++ b/nixos/services/default.nix
@@ -8,6 +8,7 @@ let
     "services/networking/jitsi-videobridge.nix"
     "services/networking/prosody.nix"
     "services/web-apps/jitsi-meet.nix"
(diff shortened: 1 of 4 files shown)


frlan:
LGTM

PR #483, opened 2022-02-16
devhost: make container cleanup optional, make nginx more robust

ctheune:

## Release process

Impact:

none

Changelog:

* Improve our container `devhost` role: automatic cleanup is optional now and nginx configs are more robust regarding container DNS lookups.

## Security implications

no specific requirements afaict

tested manually on our internal devhost

### DIFF
--- a/nixos/roles/devhost/default.nix
+++ b/nixos/roles/devhost/default.nix
@@ -8,21 +8,21 @@ let
   testedRoles = attrNames (lib.filterAttrs (n: v: v.supportsContainers or true) config.flyingcircus.roles);
   excludedRoles = attrNames (lib.filterAttrs (n: v: !(v.supportsContainers or true)) config.flyingcircus.roles);
 
-  enabledContainers = with lib;
-    (filter (container: container.enabled or true)
-           (map
-            (filename: fromJSON (readFile "/etc/devhost/${filename}"))
-            (filter
-              (filename: hasSuffix ".json" filename)
-              (attrNames (readDirMaybe "/etc/devhost/")))))
-    # BBB can be removed after a grace period
-    ++ (filter (container: container.enabled or true)
-           (map
-            (filename: fromJSON (readFile "/etc/devserver/${filename}"))
-            (filter
-              (filename: hasSuffix ".json" filename)
-              (attrNames (readDirMaybe "/etc/devserver/")))));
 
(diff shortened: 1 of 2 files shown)


nixos/roles/devhost/default.nix
```diff
@@ -8,13 +8,12 @@ let
...
-              (filename: hasSuffix ".json" filename)
-              (attrNames (readDirMaybe "/etc/devhost/")))))
+  allContainers = with lib;
+     (map
+      (filename: fromJSON (readFile "/etc/devhost/${filename}"))
+      (filter
+        (filename: hasSuffix ".json" filename)
+        (attrNames (readDirMaybe "/etc/devhost/"))))
     # BBB can be removed after a grace period
     ++ (filter (container: container.enabled or true)
```
dpausp: Does it make sense to filter for enabled containers here? The result is called `allContainers`

nixos/roles/devhost/default.nix
```diff
@@ -191,6 +212,16 @@ in
       };
       networking.nat.internalInterfaces = ["ve-+"];
 
+      networking.extraHosts = ''
+        # static entries for devhost containers to avoid nginx issues
+        # if containers are not running.
+        #
+        # This could be a problem with duplicate entries, but I think
+        # we'll fine.
```
dpausp: "we'll be fine"

nixos/roles/devhost/default.nix
```diff
@@ -181,6 +190,18 @@ in
         fi
       '';
 
+      # BBB can be removed after a grace period
+      flyingcircus.activationScripts.devhost-migrate-ip-config = ''
+        set -x
+        cd /etc/devhost
```
dpausp, line 195: Is it ok to ignore the old `/etc/devserver` here?

dpausp (changes requested):

nixos/roles/devhost/default.nix
ctheune: Yeah, this has been rolled out already and I'm assuming the other script runs first in any case. Also, even if it doesn't it's working in a converging manner, so ...

nixos/roles/devhost/default.nix
ctheune: Stupid copy and paste mistake. Refactored.

PR #498, opened 2022-03-04
[21.05] ceph/radosgw: bind to port 80 and provide iptables port redirection

ctheune:
This fixes pre-signed URLs where Ceph does not understand that
the process is running on 7480 but virtual hosted on 80.

PL-130368 Fix S3 presigned URLs

## Release process

Impact:

* Our internal Rados Gateways will be restarted but frontends will fail over between multiple gateways transparently.

Changelog:

* This fixes pre-signed URLs where Ceph does not understand that the process is running on 7480 but virtual hosted on 80.

## Security implications

no relevant changes

functionality tested manually in dev with zagy

### DIFF
--- a/nixos/roles/ceph/rgw.nix
+++ b/nixos/roles/ceph/rgw.nix
@@ -41,6 +41,7 @@ in
           admin socket = /run/ceph/radosgw.asok
           rgw data = /srv/ceph/radosgw/ceph-$id
           rgw enable ops log = false
+          rgw frontends = "civetweb port=80"
           debug rgw = 0 5
           debug civetweb = 1 5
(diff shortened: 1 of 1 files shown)


nixos/roles/ceph/rgw.nix
```diff
@@ -86,18 +87,38 @@ in
         };
       };
 
+      # use mkAfter to ensure we run after setting up the nat chains.
+      networking.firewall.extraStopCommands = ''
```
dpausp, line 90: is`mkAfter` missing here or is the comment wrong?

dpausp (changes requested):

PR #508, opened 2022-03-21
fc-agent command refactoring, various smaller improvements

dpausp:
* Commands are now: `fc-manage`, `fc-maintenance` and `fc-resize-disk`, all
  built using the typer CLI library.  All other executable script
  functionality moved to one of these.
* `fc-manage` has new sub commands `update-enc` and `switch`. They cover the
  same functionality as the old (`-b -c -e`) commands which are still
  present.
* `fc-manage dry-activate` now allows to pre-build and predict unit
  changes for a given nixexprs (channel) URL.
* `fc-maintenance`
  * Add show command for more verbose info about a single maintenance
    request. Also supports dumping the raw YAML serialization with
    `--dump-yaml`.
  * Running the script without arguments does not schedule/run anymore,
    use the explicit `schedule` and `run` commands for that.
  * The `request` command bundles everything to request new maintenance
    activities: `reboot`, `update`, `script` and `system-properties` (mem, cpu,
    qemu, kernel) .
  * Service users now may sudo run `list`/`show`/`delete`.
  * Commands now use rich for nicer output.
* `fc-resize` is replaced by `fc-resize-disk` and `fc-maintenance request
  system-properties`. 
  `fc-resize-disk` now only does what the name says and
  runs before everything else in the fc-agent to make space for the
  agent run.
* Unified logging setup to make sure that all entry points have proper
  structlog config and log command output to separate files under
  /var/log/fc-agent. Before, command output from maintenance activities
  was lost and is now kept properly, for example (#PL-130532).
* Channel updates are now done by a separate unit `fc-update-channel`
  which runs every hour, with randomized start time.  The agent lock
  prevents the channel update from running at the same time as the
  management commands from the fc-agent service.
* Commands now show when they are blocked waiting for the lock file.
* Disabling the agent by setting `enable=false` only disables the timer.
  The service unit will be always present and can be started manually.
  The same applies to the new fc-update-channel units.
* Removed `KillMode=none` from the fc-agent unit. The setting is
  deprecated and may go away in the future. systemd can now kill agent
  processes so we can use TimeoutStartSec now.

## Release process

Impact:

Changelog:

* fc-agent: restructured fc-manage and fc-maintenance commands, along with improvements to command output and logging: `fc-manage` has new sub commands `update-enc` and `switch`. They cover the same functionality as the old (`-b -c -e`) commands which are still present. Service users now may sudo run `fc-maintenance` `list`/`show`/`delete` to look at and delete maintenance requests (#PL-130441).

## Security implications

  - the way how system management via the fc-agent works is still the same, mostly refactoring and usability improvements
  - service users may run various tasks needed for system management and informational purposes (fc-manage, fc-maintenance list/show/delete); other, more dangerous management commands should still be limited to root  
  - tested extensively on a test VM 
    - all commands with error cases
    - channel updates with and without updateInMaintenance flag (prod/non-prod)
    - basic NixOS test which checks prod/non-prod switch for channel updates still runs


pkgs/fc/agent/fc/manage/manage.py
```diff
@@ -59,17 +59,20 @@ class Channel:
...
         self.name = name
         self.environment = environment
         self.system_path = None
 
-        if url.startswith("file://"):
-            self.is_local = True
-            self.resolved_url = url.replace("file://", "")
+        if resolve_url:
+            if url.startswith("file://"):
+                self.is_local = True
```
ctheune: This smells funny. Apparently this is never called in this way, however, from an isolation perspective the logic for resolving is only relevant for non-file URLs and should be turned upside down.

pkgs/fc/agent/fc/maintenance/request.py
```diff
@@ -126,6 +128,15 @@ def load(cls, dir, log):
             instance = yaml.load(f, Loader=yaml.FullLoader)
             if instance.next_due and not instance.next_due.tzinfo:
                 instance.next_due = pytz.UTC.localize(instance.next_due)
+            if instance.added_at and not instance.added_at.tzinfo:
```
ctheune: Please refactor and extract the if/else/update logic into a single statement:

```
a = ensure_timezone_present(a)
```

pkgs/fc/agent/fc/maintenance/reqmanager.py
```diff
@@ -342,6 +372,52 @@ def archive(self):
...
+                    f"request with prefix '{request_id}'![/bold]"
+                )
+                return
+
+        if len(requests) > 1:
+            rich.print(
+                "[bold blue]Notice:[/bold blue] [bold]Multiple requests "
+                "found, showing the newest:[/bold]\n"
+            )
+            req = sorted(
```
ctheune: Small improvement possibility

* sort first
* check for > 1 and print note
* always choose requests[-1]

pkgs/fc/agent/fc/maintenance/tests/test_system_properties.py
```diff
@@ -0,0 +1,2 @@
+import pytest
```
ctheune: hmmm?

pkgs/fc/agent/fc/util/disk.py
```diff
@@ -0,0 +1,123 @@
...
+        ).decode()
+        free = self.r_free.search(sgdisk_out)
+        if not free:
+            raise RuntimeError(
+                "unable to determine number of free sectors", sgdisk_out
+            )
+        return int(free.group(1))
+
+    def grow_partition(self):
+        log.info(
```
ctheune, line 52: This is a bit from the outside, but please review the current state in the Gentoo-based fc.agent as we have a lot of instrumentation code to help us diagnose problems here.

pkgs/fc/agent/fc/util/lock.py
```diff
@@ -1,22 +1,57 @@
...
+
+    if not lockfile.exists():
+        lockfile.touch()
+
     with open(lockfile, "r+", buffering=1) as f:
-        fcntl.flock(f, fcntl.LOCK_EX)
+        try:
+            fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except OSError:
+            file_locked = False
```
ctheune: Suggestion: remove the superfluous if/else handler

pkgs/fc/agent/fc/util/lock.py
```diff
@@ -1,22 +1,57 @@
...
+            lockfile=lockfile,
+        )
         yield
+        fcntl.flock(f, fcntl.LOCK_UN)
+        log.debug(
+            "lock-released",
+            _replace_msg="Released {lockfile} from PID {pid}",
+            lockfile=lockfile,
+        )
         f.truncate(0)
```
ctheune: Danger! You unlocked it and now the truncate() comes in too late.

pkgs/fc/agent/fc/util/lock.py
```diff
@@ -1,22 +1,57 @@
...
+            other_pid = f.read().strip()
+            log.info(
+                "lock-try",
+                _replace_msg="Looks like another management command is "
+                "running, waiting for {lockfile} which is "
+                "locked by PID {other_pid} "
+                "(no timeout) ...",
+                lockfile=str(lockfile),
+                other_pid=other_pid or "<unknown>",
+            )
```
ctheune, line 34: Formatting suggestion:

```
            log.info(
                 "lock-try",
                 # fmt: off
                 _replace_msg="Looks like another management command is "
                              "running, waiting for {lockfile} which is "
                              "locked by PID {other_pid} (no timeout) ...",
                 # fmt: on
                 lockfile=str(lockfile),
                 other_pid=other_pid or "<unknown>",
             )
```

pkgs/fc/agent/fc/util/lock.py
```diff
@@ -1,22 +1,57 @@
...
+        if not file_locked:
+            other_pid = f.read().strip()
+            log.info(
+                "lock-try",
+                _replace_msg="Looks like another management command is "
+                "running, waiting for {lockfile} which is "
+                "locked by PID {other_pid} "
+                "(no timeout) ...",
+                lockfile=str(lockfile),
+                other_pid=other_pid or "<unknown>",
```
ctheune: Maybe move the 'or' to the f.read() line as that case is kind of important from a logical perspective here and I already didn't see it once ;)

ctheune (changes requested):

PR #514, opened 2022-03-25
[21.05] Disable createHome for users

dpausp:
(Same as #512 for 21.11) 

This avoids setting the permissions for home dirs twice in the
activation script. This led to situations in the past where the
activation script died after the users script and home dir permissions
were set to 700.
Let our activation snippet handle home dir creation and
setting permissions to 755.

 #PL-130524

## Release process

Impact:

Changelog:

* Make sure that home dir permissions are always set correctly when activating a new system (#PL-130524). 

## Security implications

  - we n must set home dir permissions reliably to the value we want
  - manually checked on a test VM that home dirs are created with the right permissions and that wrong permissions are fixed when running the activation script

### DIFF
--- a/nixos/platform/users.nix
+++ b/nixos/platform/users.nix
@@ -36,6 +36,7 @@ let
           value = {
             description = user.name;
             group = primaryGroup user;
+            createHome = false;
             extraGroups =
               lib.optionals (user.class == "service") serviceUserExtraGroups;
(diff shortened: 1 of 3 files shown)


dpausp:
second try of backporting #512 to 21.05. #513 was also merged to 21.11 because it had the wrong base branch.

PR #545, opened 2022-05-13
Agent: fix bootstrapping for new VMs

dpausp:
Because of an inverted file existence check, initial ENC data was never
loaded on bootstrap, rendering new VMs unusable. The failure to load ENC
data also uncovered a problem with the fallback channel URL which is
used when there's no channel URL in the ENC data. Redirects for that
URL are resolved now before passing it to nix-build which doesn't work
with redirects.

 #PL-130631

## Release process

Impact:

Changelog:

(not relevant for running customer VMs)

## Security implications

  - nothing new here, just fixing the VM bootstrap process. No impact on running systems. 
  - automated test checks that initializing ENC data now works and error cases are reported properly.
  - manually checked on a test VM that fc-manage switch --update-enc works as expected.

### DIFF
--- a/pkgs/fc/agent/fc/manage/manage.py
+++ b/pkgs/fc/agent/fc/manage/manage.py
@@ -88,15 +88,19 @@ class Channel:
         return NotImplemented
 
     @classmethod
-    def current(cls, log, channel_name):
-        """Looks up existing channel by name."""
+    def current(cls, log, channel_name, resolve_url=False):
+        """Looks up existing channel by name.
+        The URL found is usually already resolved (no redirects)
+        so we don't do it again here. It can still be enabled with
+        `resolve_url`, when needed.
+        """
         if not p.exists("/root/.nix-channels"):
             return
         with open("/root/.nix-channels") as f:
             for line in f.readlines():
                 url, name = line.strip().split(" ", 1)
                 if name == channel_name:
-                    return Channel(log, url, name, resolve_url=False)
+                    return Channel(log, url, name, resolve_url=resolve_url)
 
     def load_nixos(self):
         self.log_with_context.debug("channel-load-nixos")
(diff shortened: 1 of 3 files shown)


pkgs/fc/agent/fc/manage/manage.py
```diff
@@ -88,15 +88,19 @@ def __eq__(self, other):
         return NotImplemented
 
     @classmethod
-    def current(cls, log, channel_name):
-        """Looks up existing channel by name."""
+    def current(cls, log, channel_name, resolve_url=False):
+        """Looks up existing channel by name.
+        The URL found is usually already resolved (no redirects)
```
osnyx, line 93: What is the reason for **not** allowing redirects in the default case and only allowing them for the fallback channel? Security considerations?
Because otherwise just *always* allowing redirects is more robust.

pkgs/fc/agent/fc/util/enc.py
```diff
@@ -30,13 +30,43 @@ def load_enc(log, enc_path):
...
 
 
-def seed_enc(tmpdir, enc_path):
-    tmp_enc_path = tmpdir / "fc-data/enc.json"
-    if not enc_path.exists():
-        return
-    if not tmp_enc_path.exists():
+def seed_enc(log, tmpdir, enc_path):
+    """Populates ENC data during bootstrapping. Our automation puts an
+    initial ENC file in /tmp which provides the password to allow the agent to
```
osnyx, line 35: phrasing: This initial file is the "seeding" referred to later in the logging messages, correct? The same terminology should be used in the docstring as well to link these together.

```suggestion
    initial ENC file in /tmp for seeding which provides the password to allow the agent to
```

pkgs/fc/agent/fc/util/enc.py
```diff
@@ -30,13 +30,43 @@ def load_enc(log, enc_path):
...
+            ),
+            enc_path=str(enc_path),
+            initial_enc_path=str(initial_enc_path),
+        )
+        shutil.move(initial_enc_path, enc_path)
+    else:
+        log.info(
+            "seed-enc-initial-data-not-found",
+            _replace_msg=(
+                "ENC file not found at {enc_path} and initial data from "
```
osnyx, line 64:
```suggestion
                "ENC file not found at {enc_path} and initial seed data from "
```

see terminology remark above

osnyx (approved):

pkgs/fc/agent/fc/manage/manage.py
dpausp: resolving it always caused a lot of unneccessary requests to our hydra so we disabled it.

osnyx (commented):
LGTM

PR #554, opened 2022-06-02
[21.05] elasticsearch: add dummy option single_node

dpausp:
Add a dummy option that does nothing on 21.05 to make upgrades to 21.11
easier.  Set it to false before upgrading multi-node clusters! On
21.11, the option is required and has `true` as default which splits
multi-node clusters.  This behaviour breaks indices that have replicas
and they cannot be recovered.

 #PL-130608

## Release process

Impact:

Changelog:

* Elasticsearch: add dummy option `services.elasticsearch.single_node` which does nothing on 21.05 but is required on 21.11 to make upgrades easier. On 21.11, the option has `true` as default which breaks multi-node clusters on upgrade. Make sure to set the option to `false` on 21.05 before upgrading such clusters!

## Security implications

  - option does nothing here, only relevant on 21.11 
  - checked on a test VM that defining the option works

### DIFF
--- a/nixos/roles/elasticsearch.nix
+++ b/nixos/roles/elasticsearch.nix
@@ -130,6 +130,16 @@ in
       enable = mkEnableOption "Enable the Flying Circus elasticsearch7 role.";
       supportsContainers = fclib.mkEnableContainerSupport;
     };
+
+    # Dummy option that does nothing on 21.05 to make upgrades to 21.11
+    # easier.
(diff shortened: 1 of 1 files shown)


osnyx (approved):
While in theory it'd be nice to add an actual description to the dummy option, we currently do not use such meta documentation infos. Thus this simple empty option is sufficient.

PR #562, opened 2022-06-20
PL-130673 add ldapMemberOf option, set to "crew" for global statshost

elikoga:

make the RG name that is checked configurable and set it to "crew" for the central statshost. "crew" is a documented platform role/permission that is already used for access control.

PL-130673

## Release process

Impact:

none

Changelog:

* Make resource-group-based access to statshost configurable instead of always using the statshost's own RG. (PL-130673)

## Security implications

  - make the RG name that is checked configurable and set it to "crew" for the central statshost. "crew" is a documented platform role/permission that is already used for access control.
  - [x] tested by running nixos test for global statshost

### DIFF
--- a/nixos/roles/statshost/default.nix
+++ b/nixos/roles/statshost/default.nix
@@ -132,7 +132,7 @@ in
 {
 
   imports = [
-    ./global-statshost.nix
+    ./global-metrics.nix
     ./location-proxy.nix
     ./relabel.nix
     ./rg-relay.nix
@@ -330,6 +330,8 @@ in
               }];
           })
         relayLocationProxies);
+
+      flyingcircus.roles.statshost.ldapMemberOf = "crew";
     })
 
     (mkIf (cfgStatsRG.enable || cfgProxyRG.enable) {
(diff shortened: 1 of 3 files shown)


nixos/roles/statshost/default.nix
```diff
@@ -132,7 +132,7 @@ in
 {
 
   imports = [
-    ./global-metrics.nix
+    ./global-statshost.nix
```
dpausp: We should keep the global-metrics.nix file, I think this change is too confusing if someone is looking for the place to change the globally allowed metrics. Other config that is specific to the central statshost is currently in `default.nix` guarded by `mkIf cfgStatsGlobal.enable`, line 253. Please put changes there. We have to refactor the whole thing later when Gentoo is gone.

nixos/roles/statshost/global-statshost.nix
```diff
@@ -59,4 +59,6 @@ in mkIf config.flyingcircus.roles.statshost-global.enable
 
   flyingcircus.roles.statshost.prometheusMetricRelabel =
     lib.mkAfter (markAllowedMetrics ++ dropUnmarkedMetrics);
+
+  flyingcircus.roles.statshost.ldapMemberOf = "crew";
```
dpausp: move to default.nix

dpausp (changes requested):

PR #577, opened 2022-07-25
mailserver: fix password change via roundcube

dpausp:
Users couldn't change their passwords on 21.11 via the roundcube UI.
The wrapper for roundcube-chpasswd didn't work because 21.11
doesn't set setuid by default anymore for security.wrappers.
We now use sudo instead to give roundcube the permission to run
roundcube-chpasswd as vmail user who owns the mail passwd file.

 #PL-130674

## Release process

Impact:

Changelog:

* mailserver: fix changing the password via web mailer (roundcube) settings (#PL-130674).

## Security implications

  -  logged-in mail users should be able to change their own password via the UI
  - the roundcube-chpasswd command must be restricted by the roundcube user
  - tested changing the password via roundcube manually
  - automated test checks that changing the password works and sudo rules are correct
  - checked sudo rules on a test VM

### DIFF
--- a/nixos/services/mail/roundcube.nix
+++ b/nixos/services/mail/roundcube.nix
@@ -2,20 +2,18 @@
 
 let
   role = config.flyingcircus.roles.mailserver;
-  chpasswd = "/run/wrappers/bin/roundcube-chpasswd";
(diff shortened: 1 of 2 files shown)


osnyx (approved):
LGTM

PR #578, opened 2022-07-26
Update jitsi, add watchdogs, systemd notify

dpausp:

Update packages to latest stable

Jicofo and jitsi-videobridge2 services now have "watchdogs" in their
start scripts that restart the service when timing out after 40 seconds
by using WatchdogSignal=SIGTERM and Restart=Always.

Both use the Jicofo health API but they check for different things: The
videobridge watchdog checks if the videobridge is unusable for Jicofo
while Jicofo itself checks for everything else (like 5xx codes from the
API or connection timeout).

Both services now use the "notify" service type, signalling their
readyness only when the Jicofo API health check is ok (from their
perspective).

Because of that, the separate service for delaying the videobridge
start should not be needed anymore and is removed.

The videobridge also uses "Wants" for the relation to Jicofo now
so it doesn't die immediately when Jicofo exits but still starts
Jicofo if it's not active when videobridge wants to start.

Other changes:

* don't stop Jicofo when the unit changes
* fix shellcheck warnings for unquoted vars and `export var=$(...)`

## Release process

Impact:

* [NixOS 21.11] Jitsi services will be restarted. Conferences will be interrupted for a short amount of time.

Changelog:

* Jitsi: update packages to latest stable and add watchdog checks to restart Jitsi services if they are running but don't work properly. This usually happens after a reboot or large system update, making Jitsi conferences unavailable until a manual service restart (#PL-130720). 

## Security implications

  - use up-to-date Jitsi versions  
  - have proper availabilty checks for Jitsi conferencing components
  -checked if automatic restarts work and basic conferencing features are ok on a test VM  
  - looked at [Jitsi changelog](https://github.com/jitsi/jitsi-meet-release-notes/blob/master/CHANGELOG-WEB.md)

### DIFF
--- a/nixos/services/jitsi/jicofo.nix
+++ b/nixos/services/jitsi/jicofo.nix
@@ -99,23 +99,66 @@ in
       ];
       environment.JAVA_SYS_PROPS = concatStringsSep " " (mapAttrsToList (k: v: "${k}=${toString v}") jicofoProps);
 
+      stopIfChanged = false;
+
       script = ''
+        watchdog() {
+          for count in {1..300}; do
+            sleep 1
+            ${pkgs.curl}/bin/curl -s http://localhost:8888/about/health && break
+            echo "Watchdog: waiting for Jicofo startup, try: $count"
+          done
+
+          # Wait before notifying systemd because Jicofo may take a bit longer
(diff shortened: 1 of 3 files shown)


osnyx (approved):
LGTM.

PR #580, opened 2022-07-28
mailserver: handle missing FE properly, fix on devhost

dpausp:
mailserver: handle missing FE properly, fix on devhost

The role had multiple issues which prevented it from being used in
devhost containers because they don't have a ethfe interface.

- Add a new option `flyingcircus.roles.policydSPFExtraSkipAddresses` which
  works similar to `smtpBind4` which already had a proper fallback to SRV or
  none at all.
- Fix type for `smtpBind*` settings and missing mail addresses for default
  users in container.nix.
- Use flyingcircus.nginx.virtualHosts which is already correctly set up
  for the devhost environment and remove explicit FE listen addresses.

 #PL-130785

## Release process

Impact:

Changelog:

- `mailserver` role now works in devhost environments and in other situations where the machine has no FE interface (#PL-130785).

## Security implications

  - doesn't affect existing mailserver instances which all have a FE interface; role should also work when no FE interface is present for example in a devhost environment
  - deployed a container with mailserver role on devhost and checked that postfix and nginx services are running, checked on test mailserver that policyd config is still the same

### DIFF
--- a/nixos/services/mail/default.nix
+++ b/nixos/services/mail/default.nix
@@ -28,9 +28,9 @@ let
     abuse@${primaryDomain} root@${role.mailHost}
   '';
   userAliases =
-    concatStringsSep "\n" (
-      (map (e: "${e.uid}: ${concatStringsSep ", " e.email_addresses}"))
-      config.flyingcircus.users.userData);
+    concatMapStringsSep "\n"
+      (e: "${e.uid}: ${concatStringsSep ", " e.email_addresses}")
+      config.flyingcircus.users.userData;
   checkMailq = "${pkgs.fc.check-postfix}/bin/check_mailq";
 
 in {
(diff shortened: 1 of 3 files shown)


nixos/services/mail/default.nix
```diff
@@ -28,10 +28,12 @@ let
     abuse@${primaryDomain} root@${role.mailHost}
   '';
   userAliases =
-    concatStringsSep "\n" (
-      (map (e: "${e.uid}: ${concatStringsSep ", " e.email_addresses}"))
-      config.flyingcircus.users.userData);
+    concatMapStringsSep "\n"
```
ctheune, line 31: I would rather recommend to set xyz@example.com for the synthetic human and service user in containers.

nixos/services/mail/default.nix
```diff
@@ -28,10 +28,12 @@ let
...
+      (e: "${e.uid}: ${concatStringsSep ", " e.email_addresses}")
+      (filter (hasAttr "email_addresses") config.flyingcircus.users.userData);
   checkMailq = "${pkgs.fc.check-postfix}/bin/check_mailq";
+  feAddresses =
```
ctheune: Hmmhmnhmnhm. I'd prefer the build to fail if running without FE. Unsure how to properly fake this in our container environment.

ctheune: One option: make this an option in our role, use FE as default and override via container.nix

ctheune (changes requested):

nixos/services/mail/default.nix
dpausp: All other parts of the role/service check if FE is available and use SRV or nothing instead if there is no frontend. The new option is the better approach here.

PR #581, opened 2022-07-29
[21.05] Pl 130798: add host utilities `fc-ledtool` and `fc-vm-migration-watch`

osnyx:
This adds two small useful utilities to our physical hosts:

- All hosts get `fc-ledtool`, a small helper script for finding and controlling the enclosure locate LEDs of physical disks.
- KVM hosts get `fc-vm-migration-watch`, a watch command displaying  both running VMs and their migration log at the same time

## Release process

Impact: just 2 comfort tools for internal use

Changelog: There is no public documentation on KVM and other physical hosts, no changelog necessary.
It might make sense to present on this internally. 

## Security implications

  These are just 2 added administration helper utilities. One of them only combines existing tools for reading system state, the other exclusively writes to `/sys/class/enclosure/`. Manual functionality test is sufficient.

  Manually tested on test KVM host kyle09. Both new commands can be executed. Further functionality of `fc-ledtool` has **not** been checked as kyle09's hardware does not support enclosure LEDs. But @ctheune has successfully tested the tool on several other machines.

### DIFF
--- a/pkgs/fc/default.nix
+++ b/pkgs/fc/default.nix
@@ -17,6 +17,8 @@ rec {
   collectdproxy = callPackage ./collectdproxy {};
   roundcube-chpasswd = callPackage ./roundcube-chpasswd {};
   fix-so-rpath = callPackage ./fix-so-rpath {};
+  ledtool = pkgs.writers.writePython3Bin "fc-ledtool"
+    {} (builtins.readFile ./ledtool/led.py);
   logcheckhelper = callPackage ./logcheckhelper { };
   megacli = callPackage ./megacli { };
   multiping = callPackage ./multiping.nix {};
--- a/pkgs/fc/ledtool/led.py
+++ b/pkgs/fc/ledtool/led.py
@@ -1,0 +1,63 @@
+import argparse
+import glob
+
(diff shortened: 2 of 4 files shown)


pkgs/fc/default.nix
```diff
@@ -17,6 +17,10 @@ rec {
   collectdproxy = callPackage ./collectdproxy {};
   roundcube-chpasswd = callPackage ./roundcube-chpasswd {};
   fix-so-rpath = callPackage ./fix-so-rpath {};
+  ledtool = pkgs.writeScriptBin "fc-ledtool"
```
dpausp: Using `pkgs.writers.writePython3Bin` instead should be a bit cleaner here, it takes care of the interpreter shebang.

pkgs/fc/ledtool/led.py
```diff
@@ -0,0 +1,64 @@
+import argparse
+import glob
+import sys
```
dpausp: unused import `sys`

dpausp (changes requested):

pkgs/fc/default.nix
osnyx: TIL

PR #596, opened 2022-09-01
Clean up special group handling, improve journal ACL setup

dpausp:
- Remove option `flyingcircus.users.adminGroup`/`adminGroupPath` and
  fetching of admins.json from the directory which defines the
  adminGroup option. We haven't had a use case for this option in years
  and changing the name of the admins group or gid is a bad idea anyway
  which will break things.The technical contacts which were also part of
  admins.json are unused in platform code.
- Add option `flyingcircus.users.requiredPlatformGroups` for group names
  we rely upon in the platform code.
- Fix journal-acl activation script error in tests by adding the admins
  group to the set of required groups. This also removes the need for
  tests to explicitly add the admins group.
- Make activation script tolerant to missing groups and print them.  The
  error message of setfacl is not helpful so we have to catch it
  ourselves.
- Test that activation script runs without errors in journal.nix and
  users.nix.

 #PL-130885

## Release process

Impact:

Changelog:

## Security implications

  - proper error reporting 
  - automated test checks that activation script can be executed without errors
  - looked at the activation script on a test VM

### DIFF
--- a/nixos/platform/systemd.nix
+++ b/nixos/platform/systemd.nix
@@ -5,9 +5,27 @@ with lib;
 
 let
   fclib = config.fclib;
+  cfg = config.flyingcircus.systemd;
 in
 {
+  options = with lib; {
+    flyingcircus.systemd = {
+
+      journalReadGroups = mkOption {
+        description = "Groups that are allowed to read the system journal.";
+        type = types.listOf types.string;
+      };
+
+    };
+  };
+
   config = {
+
+    flyingcircus.systemd.journalReadGroups = [
+      "sudo-srv"
+      "service"
+      "admins"
+    ];
 
     services.journald.extraConfig = ''
       SystemMaxUse=2G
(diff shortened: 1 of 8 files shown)


nixos/platform/systemd.nix
```diff
@@ -25,18 +25,22 @@ in
...
-      in ''
+      mkSetfaclCmd = group: ''
+        if [ $(getent group ${group}) ]; then
+          ${pkgs.acl}/bin/setfacl -R -m g:${group}:rX -m d:g:${group}:rX /var/log/journal
+        else
+          echo "Warning: expected group '${group}' not found, skipping ACL."
+        fi
+      '';
+
+      in lib.stringAfter [ "groups" ] (''
```
osnyx: I really like the alternative notation of activation scripts as an attrset with

```nix
{
  text = ''the script'';
  deps = [ "some" "other" "activation" "scripts" "like" "groups" ];
}
```

That makes it more clear what happens there IMHO. What do you think about using this for new or modified activation scripts from now on?

PR #599, opened 2022-09-15
fc-agent: rework initial run, manage state version, fix broken units related to NFS

dpausp:
- Rework initial agent run for new VMs, VMs are now reachable via SSH earlier and init should be more reliable.
- Set state version and read it from a file, default is still 22.05 if the file doesn't exist
- Fixes an issue with new 22.05 VMs with production flag: they used the floating dev channel forever because preparing the maintenance for the first channel updated failed indefinitely.
- fc-manage check looks for problems that could break agent runs, executed via sensu or manually.
- Also fixes a problem with broken unit files when httpd and/or nginx are not enabled (#PL-130509) because it blocked the new init process
 
 #PL-130893

## Release process

Impact:

Changelog:

- Improve init process for new machines which fixes an issue with unusable machines bootstrapped on 22.05 in production. The changes also increase reliability of the process and shortens the time until a VM reaches desired states. SSH access to new machines now works earlier (using an IP address, DNS may still take some time for new machines) (#PL-130893).
- Write state version (currently equal to the platform version: 22.05) to `/etc/local/nixos/state_version`. This has no effect for now but allows us to manage platform upgrades better in the future (#PL-130893).
- Fix broken `httpd.service` and `nginx.service` unit files when the services are not enabled. This caused confusing ("has a bad unit file setting") error messages (#PL-130503).

## Security implications

  -  no security impact expected, most changes affect only the bootstrapping phase of VMs before they are in productive use. 
  - manually tested code on existing VMs and bootstrapped multiple VMs with and without production flag 
  - automated channel test checks basic functionality

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -95,6 +95,10 @@ in
           ];
           groups = [ "sudo-srv" "service" ];
         }
+
+        { commands = [ "${pkgs.fc.agent}/bin/fc-manage check" ];
+          groups = [ "sensuclient" ];
+        }
       ];
 
       environment.etc."fc-agent.conf".text = ''
@@ -247,5 +251,17 @@ in
         };
       };
     })
+
+    {
+      flyingcircus.services.sensu-client = {
+        checks = {
+          fc-agent = {
+            notification = "fc-manage check failed. System may not build and update correctly.";
+            command = "sudo ${pkgs.fc.agent}/bin/fc-manage check";
+            interval = 300;
+          };
+        };
+      };
+    }
   ];
 }
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -229,6 +229,15 @@ in
       systemd.timers.fc-update-channel = {
         description = "Update system channel";
         wantedBy = [ "timers.target" ];
+
+        unitConfig = {
+          # This prevents the timer from activating before the first
+          # successful agent run finishes. The initial run takes care of
+          # updating the channel initially, without scheduling maintenance.
+          # The condition has no effect if the timer is already active.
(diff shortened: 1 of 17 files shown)


nixos/platform/agent.nix
```diff
@@ -192,10 +192,12 @@ in
         "r! /reboot"
         "f /etc/nixos/local.nix 644"
         "d /root 0711"
-        "d /var/lib/fc-manage"
-        "r /var/lib/fc-manage/stamp-channel-update"
         "d /var/log/fc-agent - - - 180d"
         "d /var/spool/maintenance/archive - - - 180d"
+        # Remove various obsolete files and directories
```
osnyx, line 201: These rules are a somewhat hacky workaround to get rid of old state files from previous platform versions. At some point, we might want to drop them, so it makes sense to keep track of for which platform releases the individual removals are designed. This allows us to drop a whole group after phasing out a certain platform release.

pkgs/fc/agent/fc/manage/manage.py
```diff
@@ -264,6 +274,95 @@ def notify(category):
...
+
+    if production_flag and uses_local_checkout:
+        warnings.append("production VM uses local dev checkout.")
+
+    # nixos channel checks (missing/malformed)
+    nixos_channel = nixos.current_nixos_channel_url(log)
+    if nixos_channel is None:
+        errors.append("`nixos` channel not set.")
+    elif not (
+        nixos.get_fc_channel_build(nixos_channel, log)
```
osnyx: Please add a comment explaining how these conditions signal a state where the channel is still allowed to be misformed. This combination of conditions is not necessarily obvious.

osnyx (changes requested):

osnyx (approved):

PR #605, opened 2022-09-23
Add fc-postgresql command and autoupgrade options

dpausp:
Add fc-postgresql command and autoupgrade options

The new command `fc-postgresql` (to be run as postgres user) can show
the current state of data directories for the available major versions,
prepare and run upgrade migrations.

Documentation now shows how to prepare and run manual upgrades using
`fc-postgresql upgrade`.

This also adds the `flyingcircus.services.postgresql.autoUpgrade`
submodule which is at the moment intended to be used internally in
platform code for roles that need postgresql and want to manage upgrades
without user intervention. The options are not in the user documentation
but we may add them later when we want to recommend using auto-upgrades
to users.

Documented fc-postgresql commands are:

- list-versions
- upgrade

Undocumented in customer docs:

- check-autoupgrade-unexpected-dbs (to be used as Sensu check if
  autoupgrade is enabled)
- prepare-autoupgrade (to be used manually to prepare automated
  upgrades)

 #PL-130197

## Release process

Impact:

Changelog:

- postgresql: Add `fc-postgresql` with various subcommands for listing data directories for all major versions of PostgreSQL and upgrading database clusters (a wrapper around `pg_upgrade`). The command has to be run as `postgres` user. We also add options for automated database upgrades which are experimental and are meant for internal use only in platform code at the moment. (#PL-130197).

## Security implications

  - No new permissions are required: we automate steps that would be done manually with the same permissions (as postgres user, limited to sudo-srv and service users) otherwise. Activating autoupgrade will reduce the need to use elevated privileges a bit.
  - new automated tests which tests various upgrade scenarios, with manual commands, mixed and fully automatic

### DIFF
--- a/nixos/services/postgresql.nix
+++ b/nixos/services/postgresql.nix
@@ -4,6 +4,7 @@ with builtins;
 
 let
   cfg = config.flyingcircus.services.postgresql;
+  upstreamCfg = config.services.postgresql;
   fclib = config.fclib;
   packages = {
     "10" = pkgs.postgresql_10;
@@ -12,6 +13,8 @@ let
     "13" = pkgs.postgresql_13;
     "14" = pkgs.postgresql_14;
   };
+
+  oldestMajorVersion = head (lib.attrNames packages);
 
   listenAddresses =
     fclib.network.lo.dualstack.addresses ++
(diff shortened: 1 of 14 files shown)


nixos/services/postgresql.nix
```diff
@@ -63,205 +97,304 @@ in {
...
+        if ! ${psql} -c '\du' template1 | grep -q '^ *nagios *|'; then
+          ${psql} -c 'CREATE ROLE nagios NOSUPERUSER NOCREATEDB NOCREATEROLE NOINHERIT LOGIN' template1
+        fi
+        if ! ${psql} -l | grep -q '^ *nagios *|'; then
+          ${postgresqlPkg}/bin/createdb --port ${toString upstreamCfg.port} nagios
+        fi
+        ${psql} -q -d nagios -c 'REVOKE ALL ON SCHEMA public FROM PUBLIC CASCADE;'
+
+        ln -sfT ${postgresqlPkg} ${upstreamCfg.dataDir}/package
+        ln -sfT ${postgresqlPkg}/package /nix/var/nix/gcroots/per-user/postgres/package_${cfg.majorVersion}
```
osnyx: ???

osnyx: Also, are these links ever going to be cleared up?

pkgs/fc/agent/fc/conftest.py
```diff
@@ -8,7 +8,7 @@
 from fc.maintenance.activity import Activity
 from fc.maintenance.reqmanager import ReqManager
 from fc.maintenance.request import Request
-from pytest import fixture
+from pytest import fixture, mark
```
osnyx: You say this can go away.

osnyx (changes requested):

nixos/services/postgresql.nix
dpausp: link in 160 is fixed now and points to the package link in the data directory. Both are cleaned up by the upgrade command at the end of the process.

osnyx (approved):

PR #606, opened 2022-09-29
[21.05] fc-ceph/osd: safety checks when destroying OSDs

osnyx:
When operating on Luminous OSDs, all operations involving the destruction of an OSD now do a safety check by default whether the operation would cause reduced data redundancy beyond the desired level. The check can be overriden if necessary.
Tests have been adapted to test both safe as well as unsafe destruction and rebuild.

Jewel OSD operations remain unchanged.

closes #PL-130928

## Release process

Impact: internal tooling only

Changelog:

## Security implications

- [x] [Security requirements](https://wiki.flyingcircus.io/System_Development_Guideline#Security_requirement_principles_and_testing) defined? This implements the *Establish secure defaults* OWASP rule for destructive OSD operations. Apart from that, operability shall remain unchanged.
- [x] Security requirements tested? Extended automated VM tests with both safe and unsafe destroy and rebuild operations; manually tested on a dev machine

### DIFF
--- a/pkgs/fc/ceph/src/fc/ceph/main.py
+++ b/pkgs/fc/ceph/src/fc/ceph/main.py
@@ -47,6 +47,12 @@ def main(args=sys.argv[1:]):
         help="Use the destruction process for the specified objectstore type, "
         "instead of autodetecting it.",
     )
+    parser_destroy.add_argument(
+        "--unsafe-destroy",
+        action="store_true",
+        help="Skip the check (Luminous only) whether an OSD is safe to destroy without "
+        "affecting data redundancy. This can result in data loss or cluster failure!!",
+    )
     parser_destroy.set_defaults(action="destroy")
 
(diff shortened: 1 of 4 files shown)


pkgs/fc/ceph/src/fc/ceph/main.py
```diff
@@ -334,6 +346,7 @@ def main(args=sys.argv[1:]):
 
     # extract parsed arguments from object into a dict
     args = vars(parser.parse_args(args))
+    print(args)
```
ctheune: This looks like debugging code?

ctheune (changes requested):

pkgs/fc/ceph/src/fc/ceph/main.py
osnyx: Indeed, I somehow missed to commit the removal.

PR #607, opened 2022-09-29
k3s frontend: adapt for coredns 1.9

dpausp:
CoreDNS removed support for wildcard queries (like `*.myapp.mynamespace.svc.cluster.local`) so we cannot get the pod IPs anymore from non-headless services to load-balance in haproxy.

There are 3 different port options now which are mutually exclusive and have to match the used service type. This also means that we cannot have the old fallback concept anymore where we talked to pods directly or used the cluster IP as backup. The preferred way is now to have an headless service which still allows us to get the Pod IPs but with a normal DNS query, which corresponds to the `podPort` option.

Forwarding to cluster IPs (`servicePort`) and to the k3s load balancer (`lbServicePort`) is also possible.

Documentation also got a significant update, mainly to describe the networking scenarios more clearly.

## Release process

Impact:

Changelog:

* Kubernetes/k3s: fix frontend setup issues that occured on NixOS 22.05 in some networking scenarios. Load-balancing between pods in HAProxy didn't work anymore with the old setup because CoreDNS doesn't allow wildcard DNS queries anymore. To see how the frontend has to be configured now for load-balancing, look at the first networking scenario and the configuration reference in the improved Kubernetes/k3s role documentation (#PL-130916).

## Security implications

  - nothing new, we still require outside traffic to the cluster to go through the frontend VM, nodes don't have public IPs. 
- manually checked in our test cluster that generated haproxy config makes sense, assertions for mutually exclusive pod options work and services in Kubernetes are accessible as expected


doc/src/kubernetes.md
```diff
@@ -173,120 +173,182 @@ VMs and all services managed by our platform can be accessed as usual by
...
 # /etc/local/nixos/myappfrontend.nix
 { ... }:
 {
-
-  flyingcircus.kubernetes.frontend.myapp = {
-    # haproxy now listens on a public IP address
-    bind = [ "192.0.124.96:443" ];
+  flyingcircus.kubernetes.frontend.traefik-ingress-https = {
+    # HAProxy listens on a public IP address using the standard HTTPS port 443
+    binds = [ "192.0.124.96:443" ];
```
osnyx: It's not completely clear to me from the context what kind of IP address this is, due to the comment above I could interpret this to be the public IP of HAProxy.
Looking at line 363 further below though reads like this is the cluster-internal Kubernetes IP that HAProxy then forwards to from its public IP address (which is not specified at all in this config snippet).

Please clarify the comment.

nixos/services/k3s/frontend.nix
```diff
@@ -29,36 +29,51 @@ let
...
+      lib.assertMsg (conf.podPort == null || conf.lbServicePort == null)
+      "flyingcircus.kubernetes.frontend.${name}: podPort and lbServicePort are both set, choose one!";
+    assert
+      lib.assertMsg (conf.lbServicePort == null || conf.servicePort == null)
+      "flyingcircus.kubernetes.frontend.${name}: lbServicePort and servicePort are both set, choose one!";
     assert
-      lib.assertMsg (conf.lbServicePort != null || conf.podPort != null)
-      "flyingcircus.kubernetes.frontend.${name}: podPort or lbServicePort must be set!";
+      lib.assertMsg (conf.servicePort == null || conf.podPort == null)
+      "flyingcircus.kubernetes.frontend.${name}: servicePort and podPort are both set, choose one!";
```
osnyx, line 65: While these three assertions form a logically valid mutual exclusion of the three port options, I had to draw a truth table tu verify their correctness.

Have you considered using three implications instead of the form

```nix
lib.assertMsg (conf.lbServicePort != null -> conf.servicePort == null && conf.podPort == null) "flyingcircus.kubernetes.frontend.${name}: lbServicePort is not the only kbernetes port option set, choose a single one only!";
```

The downside is that error messages need to be rephrased.

doc/src/kubernetes.md
```diff
@@ -173,120 +173,182 @@ VMs and all services managed by our platform can be accessed as usual by
...
 
+Note that we are using *tcp* mode of HAProxy here which just passes the TCP
+stream to the ingress pods. Load balancing is done inside of Kubernetes. The
+ingress controller is responsible for terminating SSL, managing certificates
+and redirecting from HTTP to HTTPS if applicable. This is different from
+scenario 1 where HAProxy can do the load balancing in front of the Kubernetes
+cluster and nginx on the frontend VM has the responsibilities that the
+ingress controller has in this scenario.
+
+
```
osnyx, line 354: There is no szenario 3 covering the use of `servicePort`, are these szenarios simple enough that there is no need for such an illustrative szenario description?

osnyx (changes requested):
While I cannot confidently verify that all Kubernetes and cluster-networking details are correct, what I understand looks pretty good. Some details of the documentation could be reworked though.

doc/src/kubernetes.md
dpausp: We have both variants: haproxy listening on localhost and running in http mode if its used behind Nginx (scenario 1), and haproxy listening on FE when Nginx is not used (scenario 2). I added some additional hints to the reference config and documentation text to clarify.

nixos/services/k3s/frontend.nix
dpausp: I like the current version is better because the error messages show which port options are present.

doc/src/kubernetes.md
dpausp: I think that "scenario 3" is not very relevant and I wouldn't recommend it, but that may change with experience ;) It would be very similar to the first scenario and my main intention was to show the two different ways how haproxy can be used in the Kubernetes context.

osnyx (approved):

PR #614, opened 2022-10-13
[21.05] fc-ceph-clean-deleted-vms: continue at errors

osnyx:
When encountering errors at deleting rbd images or snapshots, or creating new rbd pools, the service does not immediately fail but continues processing the remaining tasks. This allows unaffected tasks to still be executed within their deadline and not being blocked. The errors encountered are still printed to the log and the overall exit code of the service remains a failing one, to make the issues appear in monitoring.

## Release process

Impact: internal only

Changelog:
[internal] make fc-ceph-clean-deleted-vms.service and fc-ceph-purge-old-snapshots.service more resilient against failure of individual subtasks

## Security implications

  - adds the *fail safely* property to both cleanup tasks
  - a series of tasks containing failures must still signal the overall failure to monitoring (non-zero exit status), despite continuing to process other non-failing tasks
  - tested behaviour of 3 VM image deletion tasks, one of them prepared to fail, in our dev cluster. A single failing task now does not block the processing of further non-failing tasks anymore
  - other cleanup tasks (snapshot deletions) have not been tested manually, but utilise the exact same mechanisms having been demonstrated as successful with the manual test
  - Overall, automated unit (pytest) and integration (NixOS) tests would be desirable, but require a lot of setup and mocking in this current state. Setting up that testing infrastructure for this project in this state is something I do not feel comfortable with right now, additionally the complexity of that task is not proportionate to the complexity of this small fix. A regular ticket has been created for that.

### DIFF
--- a/pkgs/fc/ceph/src/fc/ceph/maintenance/luminous.py
+++ b/pkgs/fc/ceph/src/fc/ceph/maintenance/luminous.py
@@ -86,7 +86,7 @@ class VolumeDeletions(object):
                                 file=sys.stderr,
                             )
                             traceback.print_exc()
-                            status_code = 10
+                            status_code = max(status_code, 10)
                             print("Continuing...", file=sys.stderr)
 
                     if base_image is None:
@@ -100,7 +100,7 @@ class VolumeDeletions(object):
                             file=sys.stderr,
                         )
                         traceback.print_exc()
-                        status_code = 11
+                        status_code = max(status_code, 11)
                         print("Continuing...", file=sys.stderr)
 
         return status_code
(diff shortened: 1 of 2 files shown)


pkgs/fc/ceph/src/fc/ceph/maintenance/luminous.py
```diff
@@ -76,21 +112,33 @@ class MaintenanceTasks(object):
...
+                        print("Continuing...", file=sys.stderr)
+
+        return status_code
 
     def clean_deleted_vms(self):
         ceph = Cluster()
         directory = fc.util.directory.connect()
         volumes = VolumeDeletions(directory, ceph)
-        volumes.ensure()
+        volume_statuscode = volumes.ensure()
```
ctheune, line 141: unused?

ctheune (changes requested):
Looks good in general, but return code passing needs to be reviewed that all layers pass them on correctly.

PR #639, opened 2022-11-29
[21.11] add nodejs builds to our Hydra jobset

osnyx:

It turned out that hosts on 21.11 had to build nodejs-18_x locally, as the package in our overlay now differs from upstream due to the sqlite CVE fixes. This caused excessive build times on hosts.
This PR adds all nodejs releases of 21.11 to our Hydra build job set to ensure they are present in the binary cache.

## Release process

Impact:

Changelog: internal Hydra jobset config only

## Security implications

 - None, just a build pipeline jobset change
 - the "slim" variant of an already allowed insecure package isn't any more insecure than the full variant
  - [x] Hydra jobset succeeds

### DIFF
--- a/release/default.nix
+++ b/release/default.nix
@@ -106,10 +106,19 @@ let
     "backy"
   ];
 
+  # we need to ensure all nodejs packages are built, as they might differ from nixpkgs upstream
+  nodejsPkgNames = let
+    nodeReleases = [ 10 14 16 17 18 ];
+    in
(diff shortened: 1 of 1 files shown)


ctheune:
Merging despite the broken tests: apparently out test suite is causing false positives at the moment when run through hydra instead of interactively. @dpausp and @osnyx assured me that this is fine to merge.

PR #647, opened 2022-12-16
[21.05] kvm_host: add a proxy for faster RGW access

dpausp:
KVM hosts now run a haproxy which forwards traffic to storage servers running radosgw via the fast STO network.

 #PL-131066

## Release process

Impact:

Changelog:

* (internal) kvm_host: add a proxy for faster radosgw access using the storage network (#PL-131066).

## Security implications

  - Proxy must only be accessible through SRV, must not be public.
  - This is an performance optimization: instead of talking to a radosgw on a storage server directly, VMs talk to their KVM host which forwards traffic via the faster sto network. From a security perspective, there's no difference.
  - Checked on a dev KVM host that the expected haproxy config is generated, frontend is bound to srv IP and firewall rule for port 7480 is set properly.
  - Already running on a production VM via directory enc config.

### DIFF
--- a/nixos/roles/kvm.nix
+++ b/nixos/roles/kvm.nix
@@ -276,5 +276,37 @@ in
       "kernel.pid_max" = lib.mkForce "999999";  # mkForce to avoid conflict with ceph role
     };
 
+    # Run a proxy to give VMs running on this host fast access to radosgw.
+
+    flyingcircus.services.haproxy = {
+      enable = true;
+      enableStructuredConfig = true;
+
+      frontend = {
+        http-in = {
+          binds = [ "${head fclib.network.srv.v4.addresses}:7480" ];
+          default_backend = "s3";
+        };
+      };
+
+      backend = {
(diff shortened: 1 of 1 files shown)


nixos/roles/kvm.nix
```diff
@@ -276,5 +276,36 @@ in
...
+             in
+                "s3-${name} ${address}:7480 check inter 10s rise 2 fall 1 maxconn 40")
+            (fclib.findServices "ceph_rgw-internal-server");
+        };
+      };
+    };
+
+    networking.firewall.extraCommands = ''
+      # Accept traffic to the radosgw service
+      ${fclib.iptables "127.0.0.1"} -A nixos-fw -p tcp --dport 7480 -i brsrv -j nixos-fw-accept
```
ctheune: brsrv should use the proper API to look up the correct interface for this host

ctheune (changes requested):

PR #654, opened 2023-01-20
fc-collect-garbage: fix directory excludes, add history files and caches

dpausp:
History files for shells and .viminfo may contain store paths because someone viewed a store path, for example, which fc-userscan finds.

It can take a long time until these files are gone which means that a lot of paths are kept which are not really needed anymore.

Another problem is that fc-userscan also creates gcroots for invalid store paths which produces annoying output when garbage collecting.

Scanning Nix-internal dirs like .nix-defexpr and .cache/nix also makes no sense and only finds false positives.

PL-130486

## Release process

Impact:

Changelog:

* fc-userscan: fix directory path excludes and add zsh + bash shell history files, .viminfo and internal Nix directories to the exclude file. This avoids keeping around packages that are not used anymore. Please review the list of excluded directories in the documentation  (#PL-130486).

## Security implications

  - clean up unneeded packages earlier
  - checked on a test VM if unwanted files are excluded. This works for files in the home dir, but there's a problem in fc-userscan which renders paths with slashes useless. I added .cache/nix anyway.


osnyx (approved):
see #655

frlan:
LGTM

PR #655, opened 2023-01-20
fc-agent: integrate garbage collection code and use locking, keep build-only dependency outputs

dpausp:
## fc-agent: integrate garbage collection code and use locking

We have seen weird Nix store issues with concurrent gc and fc-manage runs, so
the new code uses the global agent lock to prevent this. There still can be
concurrent nix-build runs we don't control but at least we won't cause
trouble with our own management timers.

fc-collect-garbage.py was a standalone script which makes more sense as part
of the agent code. It can now be used interactively as `fc-collect-garbage`.
The move also brings proper logging, command line argument handling/help via
Typer and automated testing.

Rename files containing "garbagecollect" because the word order
collect-garbage is already used in more places and it's less confusing to use
the same order everywhere.

The userscan exclude file is now linked to /etc/userscan/exclude which is also
used in the userscan manpage to make it usable for interactive calls of
fc-userscan.

## garbagecollect: Keep outputs for build-time deps

Currently, we throw away nixpkgs, nixos-mailserver and more on every
GC run which are then re-downloaded by the next fc-manage run.

We have seen strange issues like partially deleted nixpkgs trees when the GC
and nix-build run concurrently. The GC deletes those temporary store paths
(they are not referenced anywhere after build) but nix-build needs them at
the same time. Note that there are other paths which are not kept even when
`keep-outputs` is switched on, still, so this is not a complete fix.

Keeping them also saves net/disk IO at the expense of some storage space.

## Release process

Impact:

Changelog:

* Prevent Nix store garbage collection and system rebuild task from running at the same which causes Nix store corruption in rare cases. As a side effect of the changes, service and sudo-srv users can now run the command `sudo fc-collect-garbage`to clean up unneeded paths from the Nix store with full speed, as opposed to the `fc-collection-garbage.service` which runs throttled and may take a long time. This is helpful for situations when disk space is critically low.
  The change also tells the garbage collector to keep build-only Nix dependencies until the dependent packages are collected. We have seen issues with files missing from the Nix store that are still expected to be there, breaking fc-manage until the corruption is fixed manually. We have also seen a Nix store path that got garbage-collected partially in one case. These are probably problems in Nix which have to be addressed upstream but our changes are expected to prevent that in the meantime (#PL-131232).

## Security implications

  - avoid corruption of the Nix store which can have serious impact on operations, at least breaking system rebuilds
  - checked on a test VM if garbage collection still works and that it waits until the lock can be obtained if fc-manage is running at the same time. 
  - automated NixOS gc test still works, wrote a pytest for the Python part


pkgs/fc/agent/fc/manage/collect_garbage.py
```diff
@@ -0,0 +1,136 @@
...
+            "userscan-user",
+            _replace_msg="Scanning {homedir} as {name}",
+            homedir=user.pw_dir,
+            name=user.pw_name,
+        )
+
+        p = subprocess.Popen(
+            [
+                "fc-userscan",
+                "-r",
```
osnyx: Some of the short-form parameters here are non-obvious, it can improve the readability if these are commented on.

osnyx (approved):
LGTM.

Please still manually check whether the userscan excludes work as intended, if not document as a ticket.

ctheune (approved):

PR #656, opened 2023-01-23
[21.05] We revamped backy's poetry2nix integration.

ctheune:
Due to a bug in poetry2nix some package changes (like setuptools-scm) became incompatible. To get those fixed we're now pulling in Python 3.10 and the surrounding ecosystem (poetry, poetry2nix) via our upstream 22.11 tracking repository.
This also pulls in a newer backy revision that can be used with different Ceph releases again.

## Release process

depends on https://github.com/flyingcircusio/backy/pull/28

Impact:

- internal backup tooling only

Changelog:

- update backy to 2.5 (unreleased):
  - Separate pool into fast an slow. According to our statistics 90% of jobs are faster than 10 minutes and after that they might run for hours. We now run the configured amount of workers both in a fast and slow pool so that long jobs may delay other long jobs but fast jobs should be able to squeeze through quickly.
  - Port to Python 3.10+.
  - Continue deleting snapshots even if some are protected.
  - Improve detection for 'whole object' RBD export: we failed to detect this is Ceph is built with wrong version options and shows itself as 'Development' instead of a real version.
  - Do not count running jobs as overdue.
  - Regularly log overdue jobs.
  - Delete Tags from former schedules
  - Add forget subcommand

## Security implications

  - Properly managing dependencies.
  - verify backup still works in dev environment

  - [x] verify this builds and backups still work in dev environment: ~~1 backup failure possibly related, investigate this PL-131339~~
  - [x] automated hydra tests still pass


pkgs/overlay.nix
```diff
@@ -434,6 +436,12 @@ in {
     ];
   });
 
+  poetry = nixpkgs-22_11.poetry;
+  poetry2nix = (super.lib.debug.traceVal nixpkgs-22_11.poetry2nix);
```
osnyx: Do we still need the tracing? If yes, why?

pkgs/overlay.nix
```diff
@@ -508,9 +516,9 @@ in {
   sensu-plugins-systemd = super.callPackage ./sensuplugins-rb/sensu-plugins-systemd { };
 
   # fix CVE-2022-35737
-  sqlite = lib.lowPrio (super.sqlite.overrideAttrs (oldAttrs: {
-    patches = (if (oldAttrs ? "patches") then oldAttrs.patches else []) ++ [ ./sqlite/CVE-2022-35737.patch];
-  }));
+  # sqlite = lib.lowPrio (super.sqlite.overrideAttrs (oldAttrs: {
```
osnyx: Why don't we need to patch sqlite anymore?

pkgs/backy/nautilus.nix
```diff
@@ -3,24 +3,39 @@
...
-};
+let
+# src = ../path/to/checkout;
+  src = fetchFromGitHub {
+    owner = "flyingcircusio";
+    repo = "backy";
+    # TODO: this points to a branch head that only differs by the Nautilus-specific commits
+    # from what we use for Luminous, but not the latest `main` head as that currently
+    # fails to build in NixOS.
+    rev = "7914012a4e74dfeca597f9dcde237de5ee2f41b0";
```
osnyx: Now that the 21.05 poetry2nix build is fixed, I can also rebase the Nautilus changes onto it and see whether this simplifies the packaging here.

osnyx (changes requested):

pkgs/backy/nautilus.nix
ctheune: yes, please.

pkgs/overlay.nix
ctheune: no we don't, good catch!

pkgs/backy/nautilus.nix
```diff
@@ -3,24 +3,39 @@
...
       python-lzo = super.python-lzo.overrideAttrs (old: {
         buildInputs = [ lzo ];
       });
+      telnetlib3 = super.telnetlib3.overrideAttrs (old: {
+        buildInputs = [ super.setuptools ];
+      });
+      pytest-flake8 = super.pytest-flake8.overrideAttrs (old: {
+        buildInputs = [ super.setuptools ];
+      });
+      consulate = super.consulate.overrideAttrs (old: {
```
osnyx: For fc-qemu, we use a consulate built from a more recent commit than the last release, as upstream has not made a release for quite a while.
A manual test of backy on a dev host [indicates](https://github.com/flyingcircusio/backy/pull/28#issuecomment-1410777562) that now consulate 0.6 is too old for use in backy as well. Additionally, consulate currently has code that is deprecated in python-3.10, see https://github.com/osnyx/consulate/commit/f05c5b177e03c7e14e5f92c0026d5802c1348810 for a fix suggestion.

osnyx:
Note: fc.qemu also uses consulate and needs the updated fork at latest once we move it to python-3.10 as well, but this is better being taken care of separately and tested thoroughly.

osnyx:
This is ready for review. The fixups will be squashed once the corresponding PR in backy is merged and the revision has been bumped one last time here.

osnyx (approved):

ctheune (commented):
LGTM

PR #659, opened 2023-02-09
docker: restrict access to containers by resource group

dpausp:
docker binds container services to 0.0.0.0 and no firewall rules currently restrict access to containers. For SRV,
we want to apply the same restrictions as for normal host traffic, allowing only machines from the same RG.
FE access to containers is dropped completely.

PL-131042

## Release process

Impact:

Changelog:

- docker: restrict access to forwarded container ports so that they are only accessible to hosts within the same resource group, and not accessible over the public internet (PL-131042).

## Security implications

  - Ensure that applications running inside containers are not directly accessible over the public internet.
  - Tested manually to ensure:
    - The firewall activation script runs correctly when applying this change to an existing system.
    - The new firewall rules created by this change correctly block external access to ports forwarded to containers by docker.
    - The new firewall rules continue to allow outgoing connections from processes running inside containers.
    - Changes to firewall rules managed by the platform do not disrupt network connectivity for containers.

### DIFF
--- a/nixos/platform/firewall.nix
+++ b/nixos/platform/firewall.nix
@@ -33,7 +33,7 @@ let
     (lib.hasAttr "ethsrv" config.networking.interfaces)
     (lib.concatMapStringsSep "\n"
       (a:
-        "${fclib.iptables a} -A nixos-fw -i ethsrv " +
+        "${fclib.iptables a} -A fc-resource-group " +
         "-s ${fclib.stripNetmask a} -j nixos-fw-accept")
       rgAddrs);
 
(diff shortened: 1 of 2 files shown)


nixos/platform/firewall.nix
```diff
@@ -106,16 +106,34 @@ in
...
-            local = ''
-              # Local firewall rules.
-              set -v
-              ${readFile filteredRules}
-            '';
-          in lib.mkAfter (rg + local);
+        lib.mkMerge [
+          (lib.mkOrder 1100 ''
+            # FC firewall rules (1100)
+            set -x
```
dpausp: remove

dpausp (commented):
Looks good, will merge on Friday after release preparations to have more time for testing in staging.

PR #661, opened 2023-02-13
[22.05] Remove residual graylog port 9002 which seems unused.

ctheune:
This is a (likely small) security risk and it's not in use apparently.

(There are more places where 9002 is mentioned but that is accidental and we do not mix cleaning those up here.)

Fixes PL-130639

## Release process

Impact:

none

Changelog:

- Remove a residual open API access port to Graylog servers that has not been in use for a long time. Reduces the attack surface slightly. (PL-130639)

## Security implications

* Don't provide unnecessary access.

* manually checked on test38 that the virtual host and the firewall rule are gone

### DIFF
--- a/nixos/roles/graylog.nix
+++ b/nixos/roles/graylog.nix
@@ -130,8 +130,6 @@ in
 
     (lib.mkIf cfg.enable {
 
-      networking.firewall.allowedTCPPorts = [ 9002 ];
-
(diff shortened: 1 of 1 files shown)


ctheune:
We figured out that this is part of a longer series of changes that we simply WONTFIX and replace with Loki in the future.

PR #666, opened 2023-02-20
Switch development to NixOS 22.11

dpausp:
fc-22.11-dev will be the main development branch after merging this.

The following release notes cover items from the [NixOS 22.11 release notes](https://nixos.org/manual/nixos/stable/release-notes.html#sec-release-22.11) that may affect our customers as well as changes to the platform code made in the process of developing this platform version. 

It's recommended to upgrade older platform versions to 22.05 first to review deprecation warnings that could be breaking changes on 22.11. View deprecation warnings in `/var/log/fc-agent/build-output.log`, after building the system on 22.05 with `fc-manage switch`.

Significant breaking changes:

- nginx: now uses the *nginx* user to run the master process. This may cause problems
  when certificates are read from arbitrary directories, for example deployments.
  Normally, the built-in support for Letsencrypt should be used instead to avoid permission problems and make sure that certificates are rotated automatically. If using external certificates cannot be avoided, make sure that permissions allow read access for the *nginx* user, for example by applying `setfacl -Rm u:nginx:rX` to the certificate directory. 
- Deprecated settings `logrotate.paths` and `logrotate.extraConfig` have been
  removed. Please convert any uses to `services.logrotate.settings` instead before upgrading.
- `kibana` roles have been removed. Machines that use kibana should stay on 22.05 for now and move to OpenSearch/OpenSearch Dashboards later.
- `graylog`/`loghost` roles have been removed. Machines that use graylog/loghost should stay on 22.05 for now. We are working on a new logging stack on 22.11+ which is planned to be based on [Grafana Loki](https://grafana.com/oss/loki/).

Other notable changes:

- PHP is now built in NTS (Non-Thread Safe) mode by default. For Apache and
  mod_php usage, we enable ZTS (Zend Thread Safe) mode. This has been a
  common practice for a long time in other distributions.
- openssh was updated to version 9.1, disabling the generation of DSA keys
  when using `ssh-keygen -A` as they are insecure. Also, `SetEnv` directives in
  `ssh_config` and `sshd_config` are now first-match-wins.
- Python now defaults to 3.10, updated from 3.9. Python 3.11 is now stable.
- PHP now defaults to PHP 8.1, updated from 8.0.
- OpenSSL now defaults to OpenSSL 3, updated from 1.1.1.
- The nodePackages package set now defaults to the LTS release in the nodejs
  package again, instead of being pinned to `nodejs-14_x`. `nodejs-10_x` has been removed.
- The option `services.grafana.extraOptions` has been removed. For a detailed migration guide, please  review the release notes of NixOS 22.11.
- See the [release notes of NixOS 22.11](https://nixos.org/manual/nixos/stable/release-notes.html#sec-release-22.11-notable-changes) for more details.

Significant package updates:

- docker-compose: 1.29 -> 2.12
- git: 2.36 -> 2.38
- glibc: 2.34 -> 2.35
- haproxy: 2.5 -> 2.6
- k3s: 1.23 -> 1.25
- keycloak: 18 -> 20
- nix: 2.8 -> 2.11
- openssh: 9.0 -> 9.1
- postfix: 3.6.6 -> 3.7.3
- powerdns: 4.6 -> 4.7
- rabbitmq: 3.9 -> 3.10
- roundcube: 1.5 -> 1.6
- systemd: 250 -> 251
- telegraf: 1.22 -> 1.24
- varnish: 7.1 -> 7.2
- zlib: 1.2.12 -> 1.2.13
- zsh: 5.8 -> 5.9

## Release process

Impact:

Changelog:

(announce in changelog that there's a new platform version and main development has switched to it, add release notes based on this PR message)

## Security implications

  - We want to follow NixOS releases to get upstream security fixes.
  - We mostly rely on NixOS development processes and expect that the NixOS release notes tell us about new security implications and possible breakage when upgrading systems.
  - Large parts of our platform features are covered by NixOS tests and important security-related aspects like audit and sudo are also tested.
  - Check if moving back to upstream defaults that are overridden by our platform code could improve security: nginx
  - The PR also includes security-relevant changes to sudo rules:
    - users in the `admins` group can run commands without password that were already allowed for `sudo-srv` users: iotop, fc-manange, fc-maintenance show/list/delete and systemctl
    - `admins` group also is allowed to run (already prepared) maintenances manually without password, even if they are scheduled for a later time. They may not add new maintenances, though, because this would give unlimited root access.
    - `sudo-srv` users can sudo run gitlab commands without password. This is required for manually running rake tasks, for example.
  - Reviewed NixOS release notes for breaking changes, security-related and other significant changes: nothing security-related that needs special attention, significant changes are covered in our release notes.
  - nginx now uses the nginx user as master user like upstream which is more secure, automated test and manual checks confirm that this works properly for us.
  - All tests are green, various VMs already running on 22.11 without problems.


tests/lamp/vm-test.nix
```diff
@@ -1,6 +1,8 @@
 import ../make-test-python.nix ({ version ? "" , tideways ? "", lib, ... }:
 {
   name = "lamp";
+  #extraPythonPackages = p: with p; [ types-setuptools ];
```
dpausp: does this work now? If not, add a comment about what's happening

nixos/roles/statshost/default.nix
```diff
@@ -612,8 +612,8 @@ in
             login_cookie_name = "grafana9_session";
           };
 
-          ldap = {
-            enabled = "true";
+          "auth.ldap" = {
```
dpausp, line 451: squash with previous grafana commit

tests/default.nix
```diff
@@ -56,9 +56,11 @@ in {
 
   lampVm = callTest ./lamp/vm-test.nix { };
   lampVm80 = callTest ./lamp/vm-test.nix { version = "lamp_php80"; };
-  lampVm80_tideways = callTest ./lamp/vm-test.nix { version = "lamp_php80"; tideways = "1234"; };
+  # XXX: fails at the tideways check since 9d6bb722c1527a154726a4458ad7d5a865d7eea7
```
dpausp: does this still fail?

nixos/platform/agent.nix
```diff
@@ -20,6 +20,40 @@ let
...
+    #compdef fc-maintenance
+
+    _fc_maintenance_completion() {
+      eval $(env _TYPER_COMPLETE_ARGS="''${words[1,$CURRENT]}" _FC_MAINTENANCE_COMPLETE=complete_zsh fc-maintenance)
+    }
+
+    compdef _fc_maintenance_completion fc-maintenance
+
+
+    #compdef fc-slurm
```
dpausp: fc-slurm doesn't exist yet :)

nixos/platform/default.nix
```diff
@@ -87,6 +87,19 @@ in {
       description = "Where to find the ENC services json file.";
     };
 
+    flyingcircus.hostRgwAddress = mkOption {
+      default = null;
```
dpausp, line 91: readonly?

dpausp (commented):
some changes

tests/default.nix
dpausp: yes, the daemon runs but the PHP module doesn't appear in phpinfo

tests/lamp/vm-test.nix
dpausp: changed: using `Version` from `packaging.version` works with type-checking.

tests/default.nix
dpausp: it's fixed now in 5712094

nixos/platform/default.nix
```diff
@@ -87,6 +87,19 @@ in {
...
+      type = with types; nullOr str;
+      description = ''
+        IP address for the radosgw object storage proxy running on the
+        virtualization host on port 7480. It allows VMs to access the
+        S3(-compatible) storage via the fast storage network.
+
+        This is the same as the `rgw.local` entry in `/etc/hosts`. Set if the
```
osnyx, line 98: foo

osnyx:
```suggestion
        This is the same as the `rgw.local` entry in `/etc/hosts`. Value is set if the
```

Clarify that *Set* describes the state of the option value, not the verb/ action of the possibility that you can/ should set it for virtual machines.

If we decide to mark it as read-only, possibly even extend to *Value is set by the platform if…*

osnyx (changes requested):

osnyx (approved):

dpausp:
removed graylog/loghost after review

osnyx (approved):

dpausp:
Merging after a team review of the important and security-relevant aspects mentioned in the PR message.

PR #667, opened 2023-02-21
Restore build compatibility with Nix 2.3/ NixOS 21.11

osnyx:
To ensure upgradability from NixOS 21.11 to 22.05, the `builtins.ceil` had to be removed as it is only introduced in Nix 2.4, while NixOS 21.11 still evaluates the config with Nix 2.3 at upgrade time.

Semantically, and with the currently chosen rotation times, this does not change anything.

Fixup for 9992ef28ef3af9d53afeda153804604a3a991824

## Release process

Impact: none for customers

Changelog: restores the upgrade path from 21.11 to 22.05

## Security implications

  - [x] automated tests must still pass
  - [x] as this is an issue in upgrading between releases, verify the upgrade path to be fixed
  - [x] consider how to avoid this class of issues in the future
  - [x] NixOS tests still pass
  - [x] successfully upgraded a testvm from 21.11 to 22.05; the fc-agent platform code is always built in such scenarios
  - [x] for future discussions on regularly verifying upgradability: PL-131310

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -227,7 +227,12 @@ in
       # rule, no logrotate rule needed
       services.logrotate.settings = {
         "/var/log/fc-agent.log" = {
-          rotate = builtins.ceil (logDaysKeep / 30);
+          # `builtins.ceil` is only available in Nix 2.4+ <=> NixOS 22.05+, blocking
+          # updates from 21.11. The alternative, integer division, implicitly applies a
+          # `floor`. With the current logDaysKeep = 180, this does not make any difference anyways.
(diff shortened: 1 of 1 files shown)


leona-ya:
You could probably also use `builtins ? ceil` (evaluates to true/false) or `builtins.hasAttr "ceil" builtins` to check if `ceil` exists and then use it.

osnyx:
Good idea. It does not make any difference right now with the values being `180 / 30`, but I still added this case distinction. This makes it more obvious that one of the branches can just be deleted at the next release.

Of course this change is tested manually as well.

PR #672, opened 2023-03-01
Add a stub test for nodejs packages

dpausp:
Ensures that Hydra builds the packages, if needed.

PL-131331

## Release process

Impact:

Changelog:

## Security implications

  - avoid high load on VMs caused by package building (which may even repeat forever when hitting the agent timeout)
  - automated test makes sure that packages are built on our hydra.


dpausp:
merging this myself as it's only a test which doesn't affect machines

PR #673, opened 2023-03-01
[22.05] Add a stub test for nodejs packages

dpausp:
Backport of #672


dpausp:
merging this myself as it's only a test which doesn't affect machines

PR #675, opened 2023-03-08
Slurm platform integration

dpausp:
Implement slurm roles

- slurm-controller: central control service (server, slurmctld)
  - should be set up first, nodes later.
  - controller doesn't start when no nodes are present
  - drain jobs on all nodes before going into maintenance
  - resume all in-service nodes after coming back from maintenance
- slurm-node: compute node that does the work (client, slurmd)
  - automatically join the partition all
  - drain jobs before going into maintenance
- slurm-dbdserver: job accounting, has to run on the same machine as the controlle
  - Starts dbdserver and a local Percona 8 DB configured for local access
    by the slurm user to the slurm_acct_db database.
  - stores job comments.
  - uses jobacct_gather/linux.
- slurm-external-dependency
  Machines use the role slurm-external-dependency to signal that
  they have to be in service in order for the cluster to run jobs,
  for example databases. These machines are checked by the controller
  for their service state before nodes are enabled. If any of these
  machines are not in service, no nodes will be set to ready.
- fc-slurm command is used by agent for maintenance entry+exit and can
  be used as interactive tool
- sudo-srv users are allowed to to sudo-run fc-slurm and other commands as slurm user
- set real memory and cpus to reasonable defaults, configurable with options
  Currently, memory and CPU has to be the same on all nodes and controller machines,
  configure manually if that's not the case.
- default settings
  - job priority: priority/multifactor, QOS=2000
  - TaskPlugin task/affinity
- telegraf metrics and sensu checks
  - sensu checks for controller readyness (ping) and node state via
    fc-slurm check which uses pyslurm (slurm C API) to get information.
    Shows basic job stats if everything is OK.
  - telegraf gets metrics from fc-slurm metrics in JSON format. Field naming
    is taken from the GO prometheus exporter to fit the dashboard available for that.
- agent: use Python 3.10
- logging: separate caller info and verbose options in init_logging

PL-131067

## Release process

Impact:

Changelog:

## Security implications

  - As slurm is basically a "remote shell" which is designed to execute arbitrary commands wtih the same UID on the target machine as the user starting the command on another machine, and even providing interactive remote shells, it must be documented that Slurm clusters must only be run in resource groups designated for this purpose, without other applications and sensitive data which doesn't have to be in the same RG. 
  - Access to the slurm API is restricted to machines that participate in the Slurm cluster (= have a slurm-* role)
  - Other than that, we rely on the Slurm permission system, for example to deny admin commands run by unprivileged users
  - manual testing on slurm test cluster
    - slurm admin commands are only usable as slurm user
    - Python tests cover all slurm maintenance operations
    - checked Munge setup which manages the secret key needed to access the Slurm API. The key is provided via directory secrets, generated for slurm-controller services and only accessible by machines with `slurm-*` roles.
  - role documentation states that slurm clusters must run in a separate RG

### DIFF
--- a/doc/src/slurm.md
+++ b/doc/src/slurm.md
@@ -1,0 +1,259 @@
+(nixos-slurm)=
+
+# Slurm Workload Manager
+
+:::{note}
+Slurm support is in beta. Feel free to use it, but we suggest contacting
+our support before putting anything into production.
+:::
+
+[Slurm](https://www.schedmd.com/) is an open source, fault-tolerant, and
+highly scalable cluster management and job scheduling system. Slurm consists
+of various services which are represented by separate Flying Circus roles
+documented below.
+
+The remainder of this documentation assumes that you are aware of the basics of
+Slurm and understand the general terminology.
+
+We provide version 22.5.0 of Slurm.
+
+## Basic architecture and roles
+
+:::{warning}
+Keep in mind that Slurm is built to execute arbitrary commands on any Slurm node
+with the permissions of the user starting the command on possibly another
+machine. Keep sensitive data away from Slurm nodes and isolate Slurm as much as
+possible, using a dedicated resource group without other applications.
+:::
+
+You can run one Slurm cluster per resource group. We generally recommend to
+use separate clusters (and thus separate resource groups) for independent
+projects. This will give you the most flexibility and will integrate
+optimally into our platform aligning well on topics like access management,
+monitoring, SLAs, maintenance, etc.
+
+A resource group with Slurm roles can have additional machines which provide
+additional services which may be needed to run jobs in Slurm. Such required
+machines can also be included in the coordination of automated maintenance
+as described later.
+
+Machine authentication is handled by `munge`, using a shared secrets generated
+by our central management directory. New worker nodes are automatically added
+to existing clusters.
+
+
+### slurm-controller
+
+:::{notice}
+For new clusters, it's recommended to first set up a controller and add nodes
+after that. The controller service will only start if there's at least one
+node.
+:::
+
+This role runs {command}`slurmctld`. We add basic Cluster readiness monitoring
+via Sensu and telemetry via Telegraf which can be ingested by a
+{ref}`nixos-statshost` and displayed using a Grafana dashboard.
+
+At the moment, we only support exactly one controller per cluster.
+
+Maintenance of a machine with this role means that all worker nodes are
+drained and set to DOWN first. Maintenance activities only start when no jobs
+are running anymore in the whole cluster.
+
+After finishing a platform management task run (which happens every 10 minutes),
+the controller sets all nodes to `ready` that have been set to `down` by an
+automated maintenance if the nodes and all external dependency machines are not
+in maintenance.
+
+
+### slurm-dbdserver
+
+:::{notice}
+At the moment, this role must run on the same machine as `slurm-controller`.
+:::
+
+Runs `slurmdbd` which is needed for job accounting. Automatically sets up a
+{ref}`nixos-mysql` database with our platform defaults and
+monitoring/telemetry.
+
+
+
+### slurm-node
+
+Runs `slurmd` which is responsible for processing jobs. There should be
+multiple nodes in your cluster for production use but applying this role to a
+machine which is also running the controller services is also supported for
+testing purposes.
+
+Before running maintenance activities, the node is drained and stops accepting new
+jobs. Nodes don't set themselves to `ready` after maintenance. Instead, the
+controller activates nodes which are not in maintenance anymore after its own
+platform management task run (every 10 minutes).
+
+### slurm-external-dependency
+
+This role does not provide any Slurm services but something that is needed to
+run jobs via Slurm, for example a database accessed by job scripts. When such
+machines go into maintenance, all nodes are drained first, like for a
+controller maintenance. After the external dependency machine has finished
+maintenance, the next run of the platform management task on the controller will set
+the nodes to `ready`.
+
+## General Interaction
+
+The usual slurm commands are installed globally on every Slurm machine.
+
+In general, all users can run slurm commands on all machines with a `slurm-*`
+role. Some commands require the use of `sudo -u slurm` to run as slurm user.
+This is allowed for(human) user accounts with the `sudo-srv` permission
+without password.
+
+Use `slurm-readme` to show dynamically-generated documentation specific for
+this machine.
+
+
+## fc-slurm command
+
+
+You can use :command:`fc-slurm` to manage the state of slurm compute nodes.
+This command is also used by our platform management task before and after
+maintenance, to fetch telemetry data from Slurm and running monitoring
+checks.
+
+Dump node state info as JSON:
+
+`fc-slurm all-nodes state`
+
+Drain all nodes (no new jobs allowed) and set them to DOWN afterwards:
+
+`sudo fc-slurm all-nodes drain-and-down`
+
+:::{note}
+Nodes that have been drained/downed manually are not set to `ready`
+automatically by the platform management task. You have to do that
+manually with the following command.
+:::
+
+Mark all nodes as READY:
+
+`sudo fc-slurm all-nodes ready`
+
+
+### Cheat sheet
+
+View the dynamically-generated documentation for a machine:
+
+```console
+$ slurm-readme
+```
+Show the current configuration:
+
+```console
+$ slurm-show-configuration
+```
+
+Show running/pending jobs
+
+```console
+$ squeue
+```
+
+Show partition state:
+
+```console
+$ sinfo
+```
+
+Show node info:
+
+```console
+$ sinfo -N
+```
+
+Show job accounting info:
+
+```console
+$ sacct
+```
+
+## Configuration reference
+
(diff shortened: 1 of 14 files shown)


doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+### slurm-controller
+
+This role runs {command}`slurmctld`. We add basic Cluster readiness monitoring
+via Sensu and telemetry via Telegraf which can be ingested by a
+{ref}`nixos-statshost` and displayed using a Grafana dashboard.
+
+At the moment, we only support exactly one controller per cluster.
+
+Maintenance of a machine with this role means that all worker nodes are
+drained and set to DOWN first. Maintenance activites only start when no jobs
```
ctheune: activities

doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+are running anymore in the whole cluster. After finishing a maintenance run
+(which happens every 10 minutes), the controller sets all nodes that have
```
ctheune: Rephrase, I got confused by the "maintenance runs finish every 10 minutes" ...

doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+been downed by the automated maintenance task to ready if they are not in
```
ctheune: "ready" could be emphasized, e.g. in a literal way with quotes or the specific slurm term.

doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+maintenance themselves and all external dependency machines are in not in
```
ctheune: all are not -> "no are"

(also: duplicate "in")

doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+
+
+### slurm-dbdserver
+
+Runs `slurmdbd` which is needed for job accounting. Automatically sets up a
+{ref}`nixos-mysql` database with our platform defaults and
+monitoring/telemetry.
+
+
+At the moment, we assume that this role is applied to the same machine as
```
ctheune: This role must run on the same machine as the ... .

ctheune: We should have an assertion to protect against using it incorrectly.

doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+machines go into maintenance, all nodes are drained first, like for a
+controller maintenance. After the external dependency machine has finished
+maintenance, the next maintenance run on the controlle will set the nodes to
+ready.
+
+## General Interaction
+
+The usual slurm commands are installed globally on every Slurm machine.
+
+In general, all users can run slurm commands on all machines with a `slurm-*`
```
ctheune, line 106: As you said this is slurm standard behaviour and e.g. NDT was expecting this. However: our communication around the login permission is that you can't interact with services. This is still true, as slurm really just is a remote shell to all systems with your user permissions so that _does_ fit in our model, I think. Communicating with an ISO would still be a good idea IMHO. 

Sounds like we need to document this in our ticket system so that an auditor would be able to find out how this came to be.

doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+Show job accounting info:
+
+```console
+$ sacct
+```
+
+## Configuration reference
+
```
ctheune, line 180: The known limitations would IMHO be better to bring them here up front, maybe even as warnings.

nixos/roles/slurm/default.nix
```diff
@@ -0,0 +1,520 @@
...
+  inherit (config) fclib;
+  controllerEnabled = config.flyingcircus.roles.slurm-controller.enable;
+  nodeEnabled = config.flyingcircus.roles.slurm-node.enable;
+  dbdserverEnabled = config.flyingcircus.roles.slurm-dbdserver.enable;
+  extDependencyEnabled = config.flyingcircus.roles.slurm-external-dependency.enable;
+  anyRoleEnabled = controllerEnabled || nodeEnabled || dbdserverEnabled || extDependencyEnabled;
+
+  serviceHostName = service: lib.head ( lib.splitString "." service.address );
+
+  # XXX: We expect controlMachine and dbdserverService to be the same at the moment.
```
ctheune, line 17: -> assertion :)

nixos/roles/slurm/default.nix
```diff
@@ -0,0 +1,520 @@
...
+          enforced, users can be limited by association to whatever job
+          size or run time limits are defined.
+        '';
+      };
+
+      extraRequiredMachines = mkOption {
+        type = with types; listOf str;
+        default = [];
+        description = ''
+          XXX: this should probably be a separate role
```
ctheune, line 107: apparently this can go away

doc/src/slurm.md
```diff
@@ -0,0 +1,249 @@
...
+A resource group with slurm roles can have additional machines which provide
+additional services which may be needed to run jobs in Slurm. Such required
+machines can also be included in the coordination of automated maintenanced
+as described later.
+
+Machine authentication is handled by munge, using a shared secrets generated
+by our central management directory. New worker nodes are automatically added
+to existing clusters.
+
+For new clusters, it's recommended to first set up a controller and add nodes
```
ctheune, line 48: maybe move this into a warning section for the controller

pkgs/fc/agent/fc/util/slurm.py
```diff
@@ -0,0 +1,743 @@
...
+        case DrainingAction.DRAIN:
+            state_change_drain = {
+                "node_names": node_name,
+                "node_state": pyslurm.NODE_STATE_DRAIN,
+                "reason": reason,
+            }
+            result = update_nodes(state_change_drain)
+            log.debug("node-update-result", result=result)
+
+    for ii in range(timeout):
```
ctheune: This is a very interesting example where the pure function design seems to reach its limits. Here's a suggestion that does two things: a) lets try using an object model for nodes instead of just operating on structs and b) try to simplify the logging. The _replace_msg mechanism could be handled similar to a `gettext` feature where we have a central mapping that checks for the log message (e.g. "drain-finished" and looks up whether a _replace_msg value is available for this).

```
log = None
timeout = 1


def get_node_info(a):
    pass


def foo():

    node = ...(log=log)

    for ii in range(timeout):
        pause = min([15, 2**ii])
        log.debug("drain-wait", sleep=pause)
        time.sleep(pause)

        node.update_info()
        if node.is_drained():
            node.log.info(
                "drain-finished",
                iterations=ii,
            )
            return

```

ctheune: It's fine to do this in a next step.

pkgs/fc/agent/fc/util/slurm.py
```diff
@@ -0,0 +1,743 @@
...
+    )
+    raise NodeStateTimeout(remaining_node_states)
+
+
+def down(log, node_name, reason: str, strict_state_check: bool = False):
+    log = log.bind(node=node_name)
+    log.debug("down-start")
+
+    node_info = get_node_info(node_name)
+    state, *flags = node_info["state"].split("+")
```
ctheune, line 341: This is also a good place for applying in an object model.

pkgs/fc/agent/fc/util/slurm.py
```diff
@@ -0,0 +1,743 @@
...
+    reason_must_match,
+    skip_in_maintenance,
+    directory,
+):
+    log = log.bind(node=node_name)
+    node_info = get_node_info(node_name)
+    state, *flags = node_info["state"].split("+")
+    log.debug("ready-pre-node-state", state=state, flags=flags)
+
+    if state in ("ALLOCATED", "IDLE", "MIXED"):
```
ctheune, line 388: This whole section is hard to swallow/consume as a human the same way as above. I think we can also use the object model to make this more expressive. Currently my brain really doesn't want to parse this long page of nested ifs sprinkled with logging code.

pkgs/fc/agent/fc/util/slurm.py
```diff
@@ -0,0 +1,743 @@
...
+
+    if skip_in_maintenance and not is_node_in_service(directory, node_name):
+        log.info(
+            "ready-pre-not-in-service",
+            node=node_name,
+            _replace_msg="{node} is still in maintenance, skipping.",
+        )
+        return ReadyPreCheckResult(state, flags, action=False)
+
+    if "*" in state:
```
ctheune, line 437: In an object model world, this could also be made more expressive. Either through a state check API or having states be smart objects, too.

```
if node.state.node_not_responding:
```

pkgs/fc/agent/fc/util/slurm.py
ctheune: Also, I think with the object model we'd need to split this up in separate modules, this is getting too long IMHO.

ctheune (changes requested):
Please double check for the smaller changes. The object model notes should go back into the ticket to follow up in a second step.

pkgs/fc/agent/fc/util/slurm.py
dpausp: I integrated the exponential backoff, object model will be implemented later.

ctheune (approved):
LGTM

PR #678, opened 2023-03-13
[21.05] fc-ceph maintenance: lock timeouts

osnyx:

fc-ceph maintenance enter can block for a very long time on a broken/ busy cluster. If it is called by the fc-agent, this also blocks the agent execution, causes the agent lock to be held for a very long time and thus blocks all other agent actions.

On a busy or broken cluster, the rbd-locktool actions might take a very long time. These subprocess calls need to have a timeout.
As we cannot know whether e.g. the lock has been acquired successfully despite the subprocess (seemingly) getting killed by a timeout, the lock needs to be freed (or at least an attempt has to be made) explicitly in case of a locking timeout. This free of course gets its own timeout as well.

In case the unlock operation/ maintenance leave operation times out, it
is retried several times. If all retries exceeded, the situation
might need to be handled by an operator though.

**Note**: This has only been implemented for Ceph Nautilus so far, as Jewel and Luminous support will soon be deprecated in fc-nixos

## Release process

Impact: internal only

Changelog:
- introduce timeouts for `fc-ceph` maintenance hooks
- better signalling of exitcodes to fc-agent
- add unit tests for fc-ceph maintenance hooks

## Security implications

  - [x] avoid stale and blocking locks in fc-ceph maintenance hooks -> fail securely
  - [x] improve test coverage of ceph maintenance hooks 
  - [x] unit tests pass
  - [x] successful manual test in dev cluster (note: timeout situations not tested manually)
  - [ ] hydra build still passes

### DIFF
--- a/pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
+++ b/pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
@@ -4,6 +4,7 @@ import json
 import subprocess
 import sys
 import traceback
+from time import sleep
 
 import fc.util.directory
 from fc.ceph.api import Cluster, Pools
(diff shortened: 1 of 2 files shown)


ctheune (approved):
LGTM

PR #683, opened 2023-03-23
[21.11] Various postgresql role improvements

dpausp:
Backport of #682 (without the doc changes)

- telegraf and Sensu now use the `fcio_monitoring` database user and database, which they can access using peer auth with an ident mapping from system to database users.
- set Restart=always for the postgresql service
- improve NixOS test with various permission and monitoring subtests.
- add `postgresql-alive` Sensu check which connect via the UNIX socket
- `postgresql-listen` Sensu checks now use a simple connection check with check_tcp to avoid setting up authentication and speed up the tests.

PL-131358


dpausp:
check for lamp packages always fails on 21.11 pull requests, we can merge this.

PR #689, opened 2023-04-06
[22.05] fc-agent check shows (deprecation) warnings, related improvements

dpausp:
The main goal was to display deprecation warnings via the `fc-agent` sensu check. 

Besides that, this fixes a deprecation warning in PostgreSQL which is always displayed by mistake and would create a lot of noise in Sensu. I also reduced noise (no caller info anymore) when using the `-v` flag on agent commands which is handy to display warnings on build as they appear. Check warnings/errors are now separated by newline instead of a space for better readability in Sensu.

## Release process

Impact:

Changelog:

- `fc-manage check` now displays NixOS (deprecation) warnings, read from `/etc/fcio_nixos_warnings`. These warnings should be fixed before upgrading the platform (PL-131380).
- Verbose output of management commands like `fc-manage -v switch` is now more readable. The flag can be used to see debug messages, build output and NixOS (deprecation) warnings (PL-131380). 
- postgresql: fix deprecation warning which was always triggered, even if no legacy config in `/etc/local/postgresql/14` was present, for example (PL-131380).

## Security implications

  - improves visibility of (deprecation) warnings that could lead to failing system builds on newer versions, for example. 
  -tried out fc-manage check on a test VM, looked at sensu output with and without warnings

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -84,6 +84,14 @@ in
   };
 
   config = mkMerge [
+    {
+      # Write NixOS warnings to a file, separated by two newlines.
+      # We use that for `fc-manage check` to display (deprecation) warnings.
+      environment.etc."fcio_nixos_warnings".text =
+        lib.optionalString (config.warnings != [])
+          ((lib.concatStringsSep "\n\n" config.warnings) + "\n");
+    }
+
     (mkIf cfg.agent.install {
       environment.systemPackages = [ pkgs.fc.agent ];
 
--- a/pkgs/fc/agent/fc/manage/manage.py
+++ b/pkgs/fc/agent/fc/manage/manage.py
@@ -301,10 +301,10 @@ class CheckResult:
 
     def format_output(self) -> str:
         if self.errors:
-            return "CRITICAL: " + " ".join(self.errors + self.warnings)
+            return "CRITICAL: " + "\n".join(self.errors + self.warnings)
 
         if self.warnings:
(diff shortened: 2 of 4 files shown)


nixos/platform/agent.nix
```diff
@@ -84,6 +84,14 @@ in
   };
 
   config = mkMerge [
+    {
+      # Write NixOS warnings to a file, separated by two newlines.
+      # We use that for `fc-manage check` to display (deprecation) warnings.
+      environment.etc."fcio_nixos_warnings".text =
```
osnyx, line 90: PR descritpion says `/etc/fcio/nixos/warnings", this says `/etc/fcio_nixos_warnings`.
If we go with the subdirectory structure, it might make sense at some point to unify the directory structure of all fc tools.

osnyx (approved):
LGTM, although I don't know whether it makes sense to cover this (NixOS side, agent side, or both) with a smoke test at least. On the other hand, it is a rather small feature addition.

pkgs/fc/agent/fc/manage/manage.py
```diff
@@ -389,6 +389,19 @@ def check(log, enc) -> CheckResult:
     else:
         errors.append("`nixos` channel not set.")
 
+    nixos_warnings_file = Path("/etc/fcio_nixos_warnings")
```
osnyx, line 392: This is consistent with the NixOS module side at least, so apparently the PR description is incorrect.

PR #692, opened 2023-04-11
matomo role, 4.13.3 -> 4.14.2

dpausp:
Add Matomo role, switch service to modifiable webroot

- matomo-setup-update copies Matomo files from Nix store to /var/lib/matomo/share.
- matomo.js is writable for Matomo (needed for plugins),
  same for legacy piwik.js
- js directory is writable for Matomo (tagmanager puts container*js files there)
- plugins can be installed by service users/deployments by putting plugin bundles
  in into `/var/lib/matomo/plugins`
- `sudo -u matomo matomo-console` can now be run by service users.
- Limit permissions as much as possible, based on Matomo recommendation and own experiments.
- Fix all Matomo diagnostic warnings and security issues reported by the SecurityInfo plugin.
- Fix matomo-setup-update unit dependencies and use RemainAfterExit to make sure that permissions are
  always set correctly and updates are checked before starting Matomo.
- matomo-archive-processing now only starts when config exists.
- Add some sensu checks for permissions, unexpected files, config file and version.

PL-131353

## Release process

Impact:

Changelog:

* matomo: add role which supports external plugins and "containers" via tagmanager (PL-131353).

## Security implications

  - force use of SSL for UI 
  - limit file permissions to a usable minimum: we disallow updating Matomo from the GUI, also installing plugins from the Marketplace is not allowed. The web root is now in a writable directory instead of the Nix store but permissions for the matomo server user are mostly read-only, with exceptions: js dir, matomo.js, piwik.js. They can be modified by admin UI and matomo-console.
  - plugin installation is limited to service users
  - nginx only serves static files from matomo core and js files needed for tracking, php files disallowed (PHP requests forwarded to php FPM)
  - tested new installation, upgrade from pre-role Matomo, with and without plugin dir hacks, looked at permissions many times
  - automated test checks if Matomo runs, sensu checks work and permissions are set correctly


nixos/services/matomo.nix
```diff
@@ -135,78 +244,184 @@ in {
...
-            cp -r "${cfg.package}/share/config" "${dataDir}/"
-            chmod -R u+rwX,g+rwX,o-rwx "${dataDir}"
-
-            # check whether user setup has already been done
-            if test -f "${dataDir}/config/config.ini.php"; then
-              # then execute possibly pending database upgrade
-              matomo-console core:update --yes
-            fi
+
+        setfacl -m u:nginx:r ${matomoTrackerFile}
```
dpausp, line 422: duplicate

nixos/roles/matomo.nix
```diff
@@ -0,0 +1,101 @@
...
+      matomo-unexpected-files = {
+        notification = "Cannot get Matomo version";
+        command = ''
+          sudo -u matomo ${matomoConsole}/bin/matomo-console diagnostics:unexpected-files \
+            && echo "OK, no unexpected files found in Matomo installation."
+        '';
+      };
+
+      matomo-version = {
+        notification = "Unexpected files found in Matomo web root.";
```
osnyx: The notification strings appear to be mixed up with the check above.

tests/default.nix
```diff
@@ -27,8 +27,7 @@ in {
   # the machine config doesn't have the overlay and the test is always running
   # the upstream version of docker, including dependencies.
   # When in doubt, it's better to write our own test or copy&paste from nixpkgs.
-  inherit (pkgs.nixosTests)
-    matomo;
+  # inherit (pkgs.nixosTests)
```
osnyx, line 30: Do we plan to adopt this again in the future? Otherwise: delete.

osnyx (changes requested):

tests/default.nix
dpausp: we will probably need this later, makes sense to keep it.

nixos/roles/matomo.nix
```diff
@@ -0,0 +1,101 @@
...
+      };
+      matomo-permissions = {
+        notification = "Matomo permissions are wrong";
+        command = ''
+          sudo ${matomoCheckPermissions}/bin/matomo-check-permissions
+        '';
+      };
+
+      matomo-unexpected-files = {
+        notification = "Cannot get Matomo version";
```
osnyx: This notification is wrong as well, had been mixed up with the one below (before it got fixed).

osnyx (approved):

PR #698, opened 2023-04-24
Opensearch roles

dpausp:
Add opensearch roles

- Add roles opensearch and opensearch_dashboards
- We use a static user instead of the upstream DynamicUser to allow running
  scripts as opensearch user.
- state dir is the upstream default of `/var/lib/opensearch`. We often changed
  these defaults to something in `/srv`. A link from `/srv/opensearch` points to
  the state dir in case someone is searching for the state in the wrong place.
- `opensearch-plugin` command can be run as opensearch user.
- Add an option to the upstream service to be able to access the config YAML.
- `opensearch-readme` displays a formatted (rich-cli) Markdown README file
  from the local config dir.
- local config dir isn't actually used for config anymore, it just has the
  readme and example NixOS config.
- Listens on SRV like Elasticsearch did.

## Release process

Impact:

Changelog:

- add opensearch and opensearch_dashboards roles, version 2.6.0. They are intended to replace Elasticsearch/Kibana 7.10.2 but the roles should only be used for new installations right now. We will provide automatic migrations from ES to OpenSearch later (PL-130611).

## Security implications

  - mostly the same as Elasticsearch which we want to replace with OpenSearch
  - service implementation is taken from 23.05
  - OpenSearch only listens on SRV
  - state dir only readable by opensearch user
    -  upstream uses DynamicUser which would be more secure but we want to be able to use `opensearch-*` scripts as opensearch users so we had to disable the option. This is not a big issue because we allow full API access to data anyway and permissions are still limited.  
  - automated test checks functionality, sensu checks and open ports
  - manually checked on test VM that a fresh installation and upgrading from pre-role opensearch works, checked open ports

### DIFF
--- a/doc/src/opensearch.md
+++ b/doc/src/opensearch.md
@@ -1,0 +1,150 @@
+(nixos-opensearch)=
+
+# OpenSearch
+
+Managed instance of [OpenSearch](https://opensearch.org) which was originally
+forked from Elasticsearch 7.10.2.
+
+
+## Interaction
+
+Run `opensearch-readme` to show a dynamic README file which shows information
+about the running OpenSearch config and role documentation.
+
+The `opensearch-plugin` command can be run as `opensearch` user.
+Changing to that user is allowed for service and sudo-srv users:
+
+```shell
+sudo -u opensearch bash
+opensearch-plugin list
+```
+
+### API
+
+The OpenSearch API is listening on the SRV interface. You can access
+the API of nodes in the same project via HTTP without authentication.
+Some examples:
+
+Show active nodes:
+
+```shell
+curl example00:9200/_cat/nodes
+```
+
+Show cluster health:
+
+```shell
+curl example00:9200/_cat/health
+```
+
+Show indices:
+
+```shell
+curl example00:9200/_cat/indices
+```
+
+## Configuration
+
+The role works without additional config for single-node setups.
+By default, the cluster name is the host name of the machine.
+
+Custom config can be set via NixOS options and is required for multi-node
+setups.
+
+Example:
+
+```nix
+{ config, pkgs, lib, ...}:
+{
+  flyingcircus.roles.opensearch = {
+    clusterName = "example";
+    nodes = [ "example00", "example02" ];
+    heapPercentage = 50;
+
+    #Only for initialization of new multi-node clusters!
+    initialMasterNodes = [ "example00" ];
+  };
+  services.opensearch.settings = {
+    "action.destructive_requires_name" = true;
+  };
+}
+
+```
+
+See `/etc/local/opensearch/opensearch.nix.example`.
+
+Copy the content to `/etc/local/nixos/opensearch.nix` to include it in
+the system config.
+
+To activate config changes, run `sudo fc-manage switch`.
+
+Run `opensearch-show-config` as `service` or `sudo-srv` user to see
+the active configuration used by OpenSearch.
+
+### Role NixOS Options
+
+**flyingcircus.roles.opensearch.clusterName**
+
+The cluster name OpenSearch will use. By default, the host name is
+used. Because of this, you have to set the cluster name explicitly
+if you want to set up a multi-node cluster.
+
+**flyingcircus.roles.opensearch.nodes**
+
+Names of the nodes that join this cluster and are eligible as masters.
+By default, all OpenSearch nodes in a resource group are part of this cluster
+and master-eligible.
+
+Note that all of them have to use the same clusterName which must be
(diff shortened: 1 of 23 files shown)


nixos/services/opensearch.nix
```diff
@@ -0,0 +1,324 @@
...
+      {
+        commands = [ "ALL" ];
+        groups = [ "sudo-srv" "service" "opensearch" ];
+        runAs = "opensearch";
+      }
+    ];
+
+    flyingcircus.services.sensu-client = {
+      expectedDiskCapacity = {
+        # XXX look at opensearch docs for these values
```
dpausp: can we describe what these numbers are?

nixos/services/opensearch.nix
```diff
@@ -0,0 +1,324 @@
...
+            migrateDataDir = ''
+              echo "Running as $(id opensearch)."
+              if ls -A /srv/opensearch/*; then
+                echo "Old data dir /srv/opensearch is not empty."
+                if ls -A /var/lib/opensearch/*; then
+                  echo "Not moving old data, new data dir /var/lib/opensearch already has content!"
+                else
+                  echo "Migrating existing data to /var/lib/opensearch"
+                  mv -v /srv/opensearch/* /var/lib/opensearch/
+                fi
```
dpausp, line 172: delete /srv/opensearch when empty

dpausp: link /srv/opensearch -> /var/lib/opensearch

nixos/services/opensearch.nix
```diff
@@ -0,0 +1,324 @@
...
+              fi
+
+              chown -R opensearch:opensearch ${cfgUpstream.dataDir}/
+            '';
+          in
+            lib.mkBefore
+              (lib.optionals (usingDefaultDataDir && usingDefaultUserAndGroup) [
+              "+${pkgs.writeShellScript "opensearch-migrate-datadir" migrateDataDir}"
+            ] ++ [
+              "${pkgs.writeShellScript "opensearch-link-jdk" "ln -sfT ${pkgs.jre_headless} ${cfgUpstream.dataDir}/jdk"}"
```
dpausp, line 183: is this still neccessary?

tests/default.nix
```diff
@@ -80,6 +80,9 @@ in {
   nfs = callTest ./nfs.nix {};
   nginx = callTest ./nginx.nix {};
   nodejs = callTest ./nodejs.nix {};
+  opensearch = callTest ./opensearch.nix {};
+  # nodejs 10 needed, not available anymore
+  #opensearch_dashboards = callTest ./opensearch_dashboards.nix {};
```
dpausp: this should work

doc/src/opensearch.md
```diff
@@ -0,0 +1,150 @@
...
+
+```
+
+See `/etc/local/opensearch/opensearch.nix.example`.
+
+Copy the content to `/etc/local/nixos/opensearch.nix` to include it in
+the system config.
+
+To see the final rendered config for OpenSearch, use the
+`opensearch-show-config` command as service or sudo-srv user.
```
osnyx: Due to the ordering of the 3 sections, they might potentially be read as 3 steps in order and it could be thought that `opensearch-show-config` is showing the config yet to be activated by the following switch.

nixos/roles/opensearch.nix
```diff
@@ -0,0 +1,140 @@
...
+      ```nix
+      ${replaceStrings ["# "] [""] exampleConfig}
+      ```
+
+      See `${localConfigDir}/opensearch.nix.example`.
+
+      Copy the content to `/etc/local/nixos/opensearch.nix` to include it in
+      the system config.
+
+      To see the final rendered config for OpenSearch, use the
```
osnyx: see comment about docs.

versions.json
```diff
@@ -2,8 +2,8 @@
   "nixpkgs": {
     "owner": "flyingcircusio",
     "repo": "nixpkgs",
-    "rev": "bed4ce58b2497f5af5dd8f98a43e349b2cbd57d9",
-    "sha256": "PRNi9I7p1ZpV+IMfFc7z3tMSebPbYof3jmH1d1Ndw4k="
+    "rev": "962bd9e48ae9fababd62bed7e023e3aa169d982b",
```
osnyx: Is this also included in another nixpkgs update PR? Does this affect other packages as well?

osnyx (approved):

versions.json
dpausp: removed commit

PR #700, opened 2023-05-02
[21.05] Sensu check: check Ceph fill ratio after snapshot restore

osnyx:

PL-130691

## Release process

Impact: internal only

Changelog:
- add a Sensu check to ceph mons that checks whether restoring the largest RBD image snapshot per Ceph crush root could exceed the fill ratio thresholds of the cluster, only done for Nautilus+ ceph releases
- some minor refactorings to Nix Ceph role platform code for proper parameter passing to the check

## Security implications

  - [x] improve monitoring for potential operation tasks
  - [x] must not introduce regressions: tests still need to pass
  - [x] new code needs to have test coverage
  - [x] check full integration of check and results into platform
  - [x] check motivated by previous operation situation, we want to avoid them in the future by being aware of problematic fill ratios
  - [x] NixOS tests still pass
  - [x] python check code has (some) unit test coverage
  - [x] manually deployed and tested in dev with creating warning and critical situations, checking that warnings actually appear in dev sensu's Uchiwa
  - [x] an integration NixOS test had been created and was useful to discover another edge case, but unfortunately that test hangs most of the time and had to be disabled again


osnyx:
still contains several FIXMEs which are markers for review.

nixos/roles/ceph/mon.nix
```diff
@@ -285,6 +289,35 @@ in
...
+            thresholds = {
+              # use canonical, non-camelCase form of ceph settings
+              nearfull = config.flyingcircus.services.ceph.allMergedSettings.mon."mon osd nearfull ratio";
+              full = config.flyingcircus.services.ceph.allMergedSettings.mon."mon osd full ratio";
+            };
+            ceph_roots = config.flyingcircus.services.ceph.server.root_rbdpool_mapping;
+          };
+        in {
+          ceph_snapshot_restore_fill = {
+            notification = "There is enough space to be able to restore the largest " +
```
osnyx: Notification should describe exactly the opposite condition, i.e. what to neotify when the check actually triggers.

ctheune: The UI isn't really helpful about this. I'd agree with writing it down as a "this is wrong" wording, but I don't feel strongly about it as it's really not used much in our context.

nixos/services/ceph/server.nix
```diff
@@ -33,6 +33,18 @@ in
           + "Only modify if really necessary, otherwise the default ceph package from the defined series is used.";
         default = fclib.ceph.releasePkgs.${cfg.cephRelease};
       };
+      root_rbdpool_mapping = lib.mkOption {
```
ctheune: I mentally missed translating "root" to "crush root", I'd like to make this more explicit: `crushroot_to_rbdpool_mapping`.

nixos/services/ceph/server.nix
```diff
@@ -33,6 +33,18 @@ in
...
+        default = {
+          default = [ "rbd.hdd" ];
+          ssd = [ "rbd.ssd" ];
+        };
+        type = let t = lib.types; in t.attrsOf (t.listOf t.str);
+        description = ''
+          Mapping of which rbd pools are operated under which crush root.
+          Currently only used by the check_snapshot_restore_fill check.
+          Backgound: We operate our rbd.hdd and rbd.hdd pools under different Ceph crush
```
ctheune: Typo: one of those is likely meant to be `rbd.ssd`

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,229 @@
+"""Check that the largest RBD image snapshot of each Ceph cluster root can be restored,
+without getting the cluster fill over a near_full threshold.
```
ctheune: fill → full? filled?

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,229 @@
...
+    if not largest_snaps:
+        print(
+            "INFO: There are no snapshots in the considered pools, nothing to do."
+        )
+        return exit_code
+    for (root, (pool, snap)) in largest_snaps.items():
+        try:
+            root_stats = cluster_fill_stats[root]
+        except KeyError:
+            # FIXME: should this also cause a WARN?
```
ctheune: this would also be a good use of exit code 3 (UNKNOWN)

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,229 @@
...
+            print(
+                f"INFO: Root {root} not in fill stats data, something is wrong here."
+            )
+            continue
+        try:
+            fill_now_bytes = int(root_stats["kb_used"] * 1024)
+            total_bytes = int(root_stats["kb"] * 1024)
+        except KeyError:
+            print(f"WARN: Unable to get cluster fill levels for root {root}.")
```
ctheune: this is also more of an UNKNOWN case

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,229 @@
...
+            exit_code = max(exit_code, EXIT_WARN)
+        # Note: fill_now_bytes takes into account the sparseness of images and only
+        # includes the actual allocated data in the cluster, while snap["size_bytes"]
+        # describes the provisioned size (100% allocation). This can lead to unintuitive
+        #  (but somehow correct) results.
```
ctheune: This is absolutely fine. We err on the side of caution here and rather get an early warning than a disaster.

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,229 @@
...
+                    < pool_largest_snap["size_bytes"]
+                ):
+                    root_largest_snap = (pool, pool_largest_snap)
+
+        if root_largest_snap and root_largest_snap[1]:
+            largest_snaps[root] = root_largest_snap
+    return largest_snaps
+
+
+def largest_snap_per_pool(poolio: rados.Ioctx, poolname: str):
```
ctheune: There's two little interesting improvements here:

```python
    largest_snap = dummy = {"size_bytes": -1}
     # in principle, only the ioctx determines the pool used. But for cleanliness always
     # instantiate a new RBD per pool
     rbdpool = rbd.RBD()
     for imgname in rbdpool.list(poolio):
         # we need an img object to query snapshots and their size
         try:
             snaps = rbd.Image(poolio, imgname).list_snaps()
             for snap in snaps:
                 if (snap["size"] > largest_snap["size_bytes"]):
                     largest_snap = {
                         "imgname": imgname,
                         "snapname": snap["name"],
                         "size_bytes": snap["size"],
                     }
         except rbd.ImageNotFound:
             # as listing images and then accessing the images is not atomic, it is
             # possible for the image to already be deleted. Ignore this, the changes
             # will be considered at the next check run.
             print(
                 f"INFO: Image {poolname}/{imgname} not found, possibly has been "
                 "deleted during the check run"
             )
             continue
     if largest_snap is dummy:
        return None
     return largest_snap
```

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,229 @@
...
+        if fill_ratio_after_restore >= thresholds.full:
+            # FIXME: Do we want this to be critical?
+            print(
+                f"CRITICAL: Restoring the snapshot "
+                f"{pool}/{snap['imgname']}@{snap['snapname']} "
+                # wording: "could", because we do not know the actual allocation ratio
+                # of the snapshot, the calculation considers the worst case of 100% of
+                # all snapshot blocks being allocated *and* not shared with the main
+                # image (or image being flattened afterwards).
+                f"with {snap['size_bytes'] / 1024**3 :,.2f}GiB could raise the cluster root "
```
ctheune: "raise" is a bit weird here.

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,229 @@
...
+        # includes the actual allocated data in the cluster, while snap["size_bytes"]
+        # describes the provisioned size (100% allocation). This can lead to unintuitive
+        #  (but somehow correct) results.
+        fill_ratio_after_restore = (
+            snap["size_bytes"] + fill_now_bytes
+        ) / total_bytes
+        # Note: This also triggers when the cluster is above the thresholds even without
+        # restoring a anapshot
+        if fill_ratio_after_restore >= thresholds.full:
+            # FIXME: Do we want this to be critical?
```
ctheune: yes but no but yes. we don't want to have to get up at 3am, but we really do want to see this and want to have a ticket.

ctheune (changes requested):

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,251 @@
...
+
+    # repacking the required values serves as an implicit schema validation
+    try:
+        thresholds = Thresholds(
+            nearfull=float(config["thresholds"]["nearfull"]),
+            full=float(config["thresholds"]["full"]),
+        )
+        pool_roots = config["ceph_roots"]
+        for root, pools in pool_roots.items():
+            assert len(pools) > 0
```
ctheune: this causes a crash and will reduce visibility, however, maybe that's acceptable because if that happens we need to (timely) fix that anyway. let's leave as it is.

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,251 @@
...
+            for pool in pools:
+                assert isinstance(pool, str)
+    except (KeyError, ValueError, AssertionError):
```
ctheune: this should be fixed: assign the exception and at least print out a repr() of it

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,251 @@
...
+    try:
+        thresholds = Thresholds(
+            nearfull=float(config["thresholds"]["nearfull"]),
+            full=float(config["thresholds"]["full"]),
+        )
+        pool_roots = config["ceph_roots"]
+        for root, pools in pool_roots.items():
+            assert len(pools) > 0
+            for pool in pools:
+                assert isinstance(pool, str)
```
ctheune: add messages to the assertions so we can understand them if they trigger

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,251 @@
...
+            continue
+    return None if largest_snap is dummy else largest_snap
+
+
+def get_cluster_fillstats(connection, pool_roots):
+    """Parses the cluster root stats into a more digestible dictionary structure.
+    Only considers the pool roots of interest to us, specified in `pool_roots`."""
+    cluster_roots = (
+        node
+        for node in json.loads(
```
ctheune: i'd extract the json.loads() expression into a separate variable

pkgs/fc/check-ceph/nautilus/tests/test_check_snapshot_restore.py
```diff
@@ -0,0 +1,202 @@
...
+
+def test_fill_warnings(cluster_stats, snap_data, default_pool_roots):
+    thresholds = snapcheck.Thresholds(nearfull=0.85, full=0.95)
+    # retrieve and create mock data
+    largest_snaps = {
+        rootname: ("nopoolname", snap_data) for rootname in default_pool_roots
+    }
+    roots = list(default_pool_roots.keys())
+    # normal case: exit code 0
+    assert 0 == snapcheck.eval_fill_warnings(
```
ctheune: This would be a tiny bit more readable if this was semantic: assert STATE_OK = ...

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
ctheune: Or even better: make the output of the subprocess call the input into this (maybe expect the json.loads to have already been handled so you can provide test data more easily (and don't have to wrap them into multi-line json strings).

the same is true for any interaction with the cluster, actually: let `main()` have an early block that does all the outside-world interactions and then have the other functions operate on that data. this will result in a lot of mocking to become a lot easier in the tests.

pkgs/fc/check-ceph/nautilus/fc/check_ceph/check_snapshot_restore.py
```diff
@@ -0,0 +1,251 @@
...
+                ):
+                    root_largest_snap = (pool, pool_largest_snap)
+
+        if root_largest_snap is not root_dummy:
+            largest_snaps[root] = root_largest_snap
+    return largest_snaps
+
+
+def largest_snap_per_pool(poolio: rados.Ioctx, poolname: str):
+    largest_snap = dummy = {"size_bytes": -1}  # a pseudo-optional value
```
ctheune: This is also a candidate for splitting between io-oriented code and the actual processing.

pkgs/fc/check-ceph/nautilus/tests/conftest.py
```diff
@@ -0,0 +1,145 @@
...
+                next_el = next(it)
+            except StopIteration:
+                break
+            if isinstance(next_el, Exception):
+                raise next_el
+            else:
+                yield next_el
+
+
+class ImageMockNautilus:
```
ctheune, line 181: if you do the io-encapsulated rewrite, this becomes much easier, too as you don't have to mock those calls but can simply prove that your consuming code can handle the different structures.

ctheune (changes requested):
Getting there. Let's take a look at the test structure when you're done with the current feedback.

ctheune:
We've invested a lot here and @osnyx moved it forward while learning a number of things about design. We could keep re-iterating here ad nauseam, but: I'd like to move forward and rather improve later when needed. This isn't code that will cause our clusters to explode, so let's leave it as is.

PR #702, opened 2023-05-03
platform: reset IPv6 address generation sysctl upon link configuration.

sysvinit:
When new hosts are booted from the bootstrap image for the first time, they use dhcpcd for dynamic network configuration. dhcpcd disables the kernel's handling of automatic IPv6 configuration and performs this function itself, however this can sometimes lead to dhcpcd removing link-local addresses when it exits without the kernel replacing them. There is a sysctl for controlling the kernel's automatic address configuration behaviour, which can be set when interfaces are configured during system activation to restore the default (and desired) behaviour.

## Release process

Impact: Internal

Changelog:

## Security implications

  - Ensure that network devices are configured appropriately so that IPv6 works as expected.
  - Tested manually in a freshly bootstrapped VM where link-local addresses were missing on all external (i.e. non-loopback) interfaces, verified that link-local addresses were present on all interfaces after applying this change.

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -220,6 +220,14 @@ in
                 sysctl net.ipv6.conf.$IFACE.autoconf=0
                 sysctl net.ipv6.conf.$IFACE.temp_valid_lft=0
                 sysctl net.ipv6.conf.$IFACE.temp_prefered_lft=0
+
+                # If an interface has previously been managed by dhcpcd this sysctl might be
+                # set to a non-zero value, which disables automatic generation of link-local
+                # addresses. This can leave the interface without a link-local address when
+                # dhcpcd deletes addresses from the interface when it exits. Resetting this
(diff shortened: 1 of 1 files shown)


osnyx (changes requested):
LGTM in general, but how about adding a comment to the sysctl *why* it needs to be explicitly set to *that particular* value?

sysvinit:
Would something like "Set to 0 to request that the kernel generate link-local addresses automatically" be appropriate? Or would you prefer a slightly longer and detailed comment in the script?

osnyx:
> Would something like "Set to 0 to request that the kernel generate link-local addresses automatically" be appropriate? Or would you prefer a slightly longer and detailed comment in the script?

Sounds almost good enough, maybe also reference that this needs to be set to fix/ override what DHCPD had done? Because otherwise we might think in the future that this is a superfluous default behaviour setting and clean it up.

osnyx (approved):

PR #714, opened 2023-05-25
Switch development to 23.05

dpausp:
fc-23.05-dev will be the main development branch after merging this.

The [upgrade notes in this PR](https://doc.flyingcircus.io/roles/fc-23.05-staging/upgrade.html) include items from the the [NixOS 23.05 release notes](https://nixos.org/manual/nixos/stable/release-notes.html#sec-release-23.05) that may affect our customers as well as changes to the platform code made in the process of developing this platform version. 

## Release process

Impact:

Changelog:

(announce in changelog that there's a new platform version and main development has switched to it, add release notes based on this PR message)

## Security implications

  - We want to follow NixOS releases to get upstream security fixes.
  - We mostly rely on NixOS development processes and expect that the NixOS release notes tell us about new security implications and possible breakage when upgrading systems.
  - Large parts of our platform features are covered by NixOS tests and important security-related aspects like audit and sudo are also tested.
  - Security-related platform differences compared to upstream NixOS 23.05 must be documented.
   - The most significant change upstream is the removal of algorithms from `libxcrypt` that are not in the `strong` list. We add back `sha256` because it's unrealistic to replace all SHA256 password hashes in the near future. This is still acceptable for existing hashes but we recommend to use better algorithms for password hashing in the future, like `yescrypt`.
  - Reviewed NixOS release notes for breaking changes, security-related and other significant changes: significant changes are covered in our release notes. 
  - All tests are green, various test VMs already running on 23.05 without problems:
    - mailserver
    - Gitlab 
    - a VM with typical roles
    - Jitsi
    - statshost
    - k3s
    - slurm


pkgs/fc/agent/default.nix
```diff
@@ -69,6 +69,6 @@ py.buildPythonPackage rec {
     xfsprogs
   ];
   dontStrip = true;
-  passthru.pythonDevEnv = py.withPackages (_: checkInputs ++ propagatedBuildInputs);
+  passthru.pythonDevEnv = python310.withPackages (_: checkInputs ++ propagatedBuildInputs);
```
osnyx, line 74: small nitpick: right now we use the name indirection `py` to pin packages to a certain version with necessary. When changing the pinning, only that name binding needs to be specified.
Now you introduce the explicit specification of a python release a 2nd time by directly specifying `python310` here.
It'd be cleaner to stack name indirections like `py` and `pyPkgs` or `py.packages`. Or at least also use another alias name here for `python310`, that is defined as a let binding directly next to `py` to keep version specifications next to each other.

nixos/services/sensu/client.nix
```diff
@@ -105,6 +105,15 @@ let
     };
   };
 
+  systemdUnitCheckOptions = {
```
osnyx, line 108: typo in commit message:
*seen* -> *seem* in
> which doesn't seen to
> work properly anymore

Also: *separated checks* -> *separate checks*, as their defining factor is not that they're segmented (separated) from each other but standalone checks (separate).

doc/src/upgrade.md
```diff
@@ -115,119 +112,84 @@ following common breaking changes and role-specific notes below.
...
-  been removed.
+- NixOS now defaults to using nsncd (a non-caching reimplementation in Rust)
+  as NSS lookup dispatcher, instead of the buggy and deprecated
+  glibc-provided nscd.
+- The `NodeJS` packages have been renamed to a more usual naming scheme,
+  for example `nodejs-19_x` is now `nodejs_19`.
+- The `dnsmasq` service now takes configuration via the
+  `services.dnsmasq.settings` attribute set. The option
+  `services.dnsmasq.extraConfig` will be deprecated when NixOS 22.11 reaches
+  end of life.
```
osnyx: IMHO this does not really make it clear whether it is already deprecated/ removed in 23.05. Because 22.11 reaches EOL *after* the initial release of 23.05, so does the deprecation only affect 23.11?

important_packages.json
```diff
@@ -113,6 +113,7 @@
   "phpPackages.composer",
   "pkg-config",
   "podman",
+  "poetry",
```
osnyx, line 116: typo in commit message

osnyx (changes requested):
Looks good overall, just a few typo hints or phrasing suggestions.

osnyx (approved):

PR #718, opened 2023-06-01
[23.05] devhost: Add VM operation mode (PL-131470)

leona-ya:

## Release process

Impact: None, since only activated if the new function is explicitly enabled.

Changelog:

- devhost: Add new feature to use VMs instead of containers. The new feature is not enabled be default (PL-131470)

## Security implications

  - No change for existing customer, who use the devhost feature (i.e. default operation mode should be container)
    - Container mode should still work
  - Dev VM SSH should not be accessible from outside (they use an insecure ssh key)
  - just a port of #704 to 23.05, so it should work
  - currently not testable as hardware is on 21.05 and nested virtualisation is not yet supported

---

Just a port of #704 to 23.05. The other PR should be reviewed first.


leona-ya:
cherry-picked all commits from #704, rebased on `fc-23.05-dev`

ctheune (approved):

PR #725, opened 2023-06-07
Improve Kubernetes monitoring and telemetry

sysvinit:
We currently do not have useful telemetry or metrics for k3s nodes on our platform. This PR makes a number of changes to fix and extend the Sensu checks, and to enable the Telegraf plugins for collecting metrics from the Kubernetes control plane.

The commit messages have detailed commentary on the changes here. In summary:
- The read-only control port on the `kubelet` process has been enabled on the worker nodes to allow the Telegraf `kubernetes` plugin to work correctly.
- Sensu checks for the k3s systemd units have been added.
- Additional Sensu health checks have been added for control plane HTTP endpoints.
  - The apiserver component running on the k3s server has been configured to allow unauthenticated clients, to which  only the health check and readiness endpoints are exposed. These are used by Sensu for checking liveness of the apiserver.
- Additional Sensu checks (from `pkgs.sensu-plugins-kubernetes`) and the Telegraf `kube_inventory` plugin require access to authenticated endpoints on the apiserver. k3s supports loading of additional vendor-supplied Kubernetes manifests found in `/var/lib/k3s/server/manifests`, which is the mechanism it uses for bundling Traefik and other third-party components. This PR adds some services to add further manifests for setting up authentication roles and exporting access tokens for Sensu and Telegraf so they can access these endpoints.

PL-131009

## Release process

Impact: internal

Changelog:

## Security implications

  - Telemetry for Kubernetes components should be available in order to ensure these components are functioning correctly.
  - The Kubernetes environment provided by the `k3s-server` and `k3s-agent` roles should have suitable default configuration to ensure secure operation (e.g. prevent unauthorised access, etc).
  - Hydra tests have been added to test the setup logic for the authentication tokens for Sensu and Telegraf.
  - Manually verified that the apiserver HTTPS endpoint is correctly firewalled on the frontend interface, and (non-exhaustively) tested that an unauthenticated client cannot access sensitive API endpoints as described in the upstream Kubernetes documentation.

### DIFF
--- a/nixos/roles/k3s/server.nix
+++ b/nixos/roles/k3s/server.nix
@@ -71,6 +71,178 @@ let
     rm /tmp/$name.kubeconfig
   '';
 
+  additionalManifests = let
+    serviceAccount = name: {
+      apiVersion = "v1";
+      kind = "ServiceAccount";
+      metadata = {
+        name = "io.flyingcircus.service.${name}";
+        namespace = "kube-system";
+      };
+    };
+    serviceAccountSecret = name: {
+      apiVersion = "v1";
+      kind = "Secret";
+      type = "kubernetes.io/service-account-token";
+      metadata = {
+        name = "io.flyingcircus.service-token.${name}";
+        namespace = "kube-system";
+        annotations."kubernetes.io/service-account.name" =
+          "io.flyingcircus.service.${name}";
+      };
+    };
+    authorizationApi = m: {
+      apiVersion = "rbac.authorization.k8s.io/v1";
+    } // m;
+    clusterRole = c: {
+      kind = "ClusterRole";
+    } // (authorizationApi c);
+    clusterRoleBinding = c: {
+      kind = "ClusterRoleBinding";
+    } // (authorizationApi c);
+
+    manifests = [
+      (serviceAccount "sensu-client")
+      (serviceAccount "telegraf")
+      (serviceAccountSecret "sensu-client")
+      (serviceAccountSecret "telegraf")
+      (clusterRole {
+        metadata.name = "flyingcircus:sensu-client";
+        rules = [{
+          apiGroups = [""];
+          resources = ["nodes"];
+          verbs = ["get" "list"];
+        }];
+      })
+      (clusterRoleBinding {
+        metadata.name = "flyingcircus:sensu-client:viewer";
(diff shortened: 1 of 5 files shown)


nixos/roles/k3s/server.nix
```diff
@@ -71,6 +71,185 @@ let
...
+      rc="$?"
+
+      if [ "$rc" = 0 ]; then
+         break
+      fi
+      sleep 1
+    done
+
+    if [ "$rc" != 0 ]; then
+      echo 'cloud not read secret token' 2>&1
```
dpausp: cloud -> could

nixos/roles/k3s/server.nix
```diff
@@ -71,6 +71,185 @@ let
...
+      exit 1
+    fi
+
+    base64 -d "$tokendir/$user.b64" > "$tokendir/$user.tmp"
+    if [ "$?" != 0 ]; then
+      echo 'could not decode secret token' 2>&1
+      exit 1
+    fi
+
+    ${pkgs.kubectl}/bin/kubectl get -n kube-system -o jsonpath='{.data.token}' secret "$secret" | \
```
dpausp: This seems redundant, we already have `$tokendir/$user.tmp` created by the code above?

dpausp (changes requested):
Looks great overall, just some cosmetic changes. Please squash some commits, we usually have one per PR but a few are also fine where it makes sense.

nixos/roles/k3s/server.nix
sysvinit: Fixed in [15c9e05](https://github.com/flyingcircusio/fc-nixos/pull/725/commits/15c9e059c359fac41adef953ddd7a0d8a410ae1c)

nixos/roles/k3s/server.nix
sysvinit: That's dead code which wasn't properly removed when I refactored that script, oops. Fixed in [15c9e05](https://github.com/flyingcircusio/fc-nixos/pull/725/commits/15c9e059c359fac41adef953ddd7a0d8a410ae1c).

sysvinit:
> Please squash some commits, we usually have one per PR but a few are also fine where it makes sense.

Ack, squash incoming.

PR #732, opened 2023-06-19
Fix redis configuration warnings, use instanced server configuration

sysvinit:
The NixOS module for Redis at some point gained support for multiple server instances, and the old options for configuration of a single global instance have been renamed. We still use the old options, which causes warnings when evaluating host configurations. This change updates our use of the Redis module to use the instanced configuration for the default (nameless) server instance.

PL-131437

## Release process

Impact: internal.

Changelog:

## Security implications

  - Redis configuration should not use renamed/deprecated module options, to ensure the platform continues to function correctly in the future.
  - Hydra tests for Redis pass on this PR branch.
  - Hydra tests for Gitlab (as a consumer of the Redis module) pass on this PR branch.

### DIFF
--- a/doc/src/redis.md
+++ b/doc/src/redis.md
@@ -20,8 +20,9 @@ In previous versions, custom redis configuration could be set
 via {file}`/etc/local/redis/custom.conf` which is not supported anymore.
 
 If you need to change the behaviour of Redis, you define your redis
-configuration with the NixOS option `services.redis.settings`. See the
-NixOS manual for further information.
+configuration with the NixOS option
+`services.redis.servers."".settings`. See the NixOS manual for further
+information.
 
 Regarding setting the redis password, see the section on redis [passwords](#password).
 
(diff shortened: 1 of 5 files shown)


doc/src/redis.md
```diff
@@ -20,8 +20,9 @@ In previous versions, custom redis configuration could be set
 via {file}`/etc/local/redis/custom.conf` which is not supported anymore.
 
 If you need to change the behaviour of Redis, you define your redis
-configuration with the NixOS option `services.redis.settings`. See the
-NixOS manual for further information.
+configuration with the NixOS option
+`services.redis.servers."".settings`. See the NixOS manual for further
+information.
```
osnyx, line 28: It might be worth explaining the unusual option name with the empty instance string here, at least briefly. Otherwise readers might deem this to be a typo.

sysvinit: Something like this?
```
If you need to change the behaviour of Redis, you define your Redis
configuration with the NixOS option
`services.redis.servers."".settings`. NixOS supports multiple
instances of Redis on a single host, so this option sets configuration
for the default instance with an empty instance name. See the NixOS
manual for further information.
```

osnyx (approved):

PR #751, opened 2023-07-18
agent: handle postpone/tempfail in maintenance enter commands

dpausp:
agent: handle postpone/tempfail in maintenance enter commands

The postpone/tempfail protocol is now properly handled by the
RequestManager.

Maintenance enter commands can return EXIT_POSTPONE (69) or
EXIT_TEMPFAIL (75) to stop execution before maintenance requests
are run, typically when pre-conditions are not met.

The important difference is that temporary failures are expected
to go away with the next agent run, so the machine stays in maintenance
mode and just tries again (until the retry limit is hit).

Postponing is for conditions that will likely not be met within minutes,
so a new execution time for activities is requested from the directory
and the node is put back in service for now.

Tempfail and postpone can be ignored with the `force_run` flag if
needed for manual cleanup.

Dev env changes:

I had to move pytestCheckHook to nativeCheckInputs to make nix-shell
add `bin/pytest` to PATH.

doCheck is not explictly set to true (which is the default) to make
switching off checks during development easier.

Updates pytest-structlog to the current version.

PL-130625

## Release process

Impact:

Changelog:

- agent: handle special exit codes of maintenance enter commands (defined by `flyingcircus.agent.maintenance.enter.*`) correctly. They can now use EXIT_POSTPONE (69) or EXIT_TEMPFAIL (75) to stop execution before maintenance requests
are run, typically when pre-conditions are not met. This is currently not used by platform code but will be useful to prevent redundant machines from having service interruptions caused by automated maintenance at the same time (PL-130625). 

## Security implications

  - nothing new, just affects handling of maintenance enter commands, no relevant additional attack surface 
  - checked on a test VM that postpone and tempfail enter commands work, Python tests cover new functionality

### DIFF
--- a/pkgs/fc/agent/fc/maintenance/tests/test_reqmanager.py
+++ b/pkgs/fc/agent/fc/maintenance/tests/test_reqmanager.py
@@ -7,7 +7,7 @@ import textwrap
 import unittest.mock
 import uuid
 from io import StringIO
-from unittest.mock import Mock, call
+from unittest.mock import MagicMock, Mock, call
 
 import freezegun
 import pytest
@@ -15,6 +15,7 @@ import pytz
 import shortuuid
 from fc.maintenance.activity import Activity, ActivityMergeResult, RebootType
 from fc.maintenance.estimate import Estimate
+from fc.maintenance.reqmanager import PostponeMaintenance, TempfailMaintenance
 from fc.maintenance.request import Attempt, Request
 from fc.maintenance.state import ARCHIVE, EXIT_POSTPONE, State
 from rich.console import Console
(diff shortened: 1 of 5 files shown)


pkgs/fc/agent/fc/maintenance/tests/test_reqmanager.py
```diff
@@ -240,13 +241,82 @@ def test_explicitly_deleted(connect, reqmanager):
...
+
+    assert log.has("enter-maintenance")
+    connect_mock().mark_node_service_status.assert_called_with("host", False)
+
+
+def test_enter_maintenance_postpone(log, reqmanager, monkeypatch):
+    monkeypatch.setattr(
+        "fc.util.directory.connect", connect_mock := MagicMock()
+    )
+    # monkeypatch.setattr("subprocess.run", run_mock := MagicMock())
```
osnyx, line 260: Is there a reason to keep this as a comment?

osnyx (approved):

pkgs/fc/agent/fc/maintenance/tests/test_reqmanager.py
dpausp: no, I'll remove it in the next agent PR

PR #756, opened 2023-07-27
Implement mechanism to avoid simultaneous maintenance of machines

dpausp:
Add a new NixOS option
`flyingcircus.agent.maintenanceConstraints.machinesInService` which is used to specify other machines that are not allowed to be in maintenance mode at the same time when a machine is trying to enter maintenance.

This is implemented by`fc-maintenance constraints` which takes multiple `--in-service` arguments to ask the directory API about the "in service" status of the given machines, returning EXIT_POSTPONE any condition is not met. The command is configured by the NixOS option to run as maintenance enter command.

This adds basic user documentation about our maintenance system which also shows how to use the new option.

PL-131290

## Release process

Impact:

Changelog:

- Document how automated maintenance works and how to avoid simultaneous maintenance of machines using the new NixOS option `flyingcircus.agent.maintenanceConstraints.machinesInService` (PL-131290).

## Security implications

  - nothing new, this is just a convenient way to specify a maintenance enter command which checks the state of other machines. 
  - checked on a test VM that the constraint is enforced correctly and maintenance works when conditions are met. 
  - Python tests cover the new `fc-maintenance constraints` command.

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -77,6 +77,21 @@ in
         type = types.bool;
       };
 
+      maintenanceConstraints = {
+        machinesInService = mkOption {
+          default = [];
+          type = with types; listOf str;
+          description = ''
+            Machines that must not be in maintenance mode at the same time.
+            After entering maintenance mode, the agent will check if a listed
+            machine is also in maintenance and leave maintenance if it finds
+            one. Due maintenance activities will be postponed in that case.
+            The name of the current machine is ignored here so you can use the same
+            value for this option on all affected machines.
+          '';
+        };
+      };
+
       updateInMaintenance = mkOption {
         default = attrByPath [ "parameters" "production" ] false cfg.enc;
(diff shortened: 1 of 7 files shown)


nixos/platform/agent.nix
```diff
@@ -77,6 +77,19 @@ in
...
 
+      maintenanceConstraints = {
+        machinesInService = mkOption {
+          default = [];
+          type = with types; listOf str;
+          description = ''
+            Machines that must not be in maintenance mode at the same time.
+            After entering maintenance mode, the agent will check if a listed
+            machine is also in maintenance and leave maintenance if it finds
+            one. Due maintenance activities will be postponed in that case.
```
osnyx, line 88: We might want to add the information, that the local hostname is ignored, here as well. This can be useful in the future when using the web search portal as a major entrypoint to configuration options.

osnyx (approved):
LGTM, possibly extend the option description as described.

PR #763, opened 2023-08-07
Handle empty Kubernetes auth tokens during server initialisation

sysvinit:
The k3s-server role bundles some vendor manifests to set up authentication to allow Sensu and Telegraf to access the Kubernetes API server. However, there is a failure case which is not handled correctly, where the token loader services attempt to read the content of the corresponding Secret resource after it has been loaded from the vendor manifests, but before the control plane has added the secret token. This will result in the loader script writing an empty access token to the token file, which will cause Telegraf to fail to start.

This PR fixes this issue by adding a check in the token loader script to ensure that the token is non-empty, and amends the execution condition in the systemd unit template so that the loader script gets executed both when the token file does not exist and when it exists but is empty.

PL-131681

## Release process

Impact: internal.

Changelog:

## Security implications

  - Handle failure cases in Kubernetes initialisation to avoid generating invalid configuration and preventing Telegraf from starting.
  - Manually verified in a test VM that applying this change to a system which is in a stuck state with invalid empty token files recovers from this state and starts Telegraf properly when switching to the new configuration.

### DIFF
--- a/nixos/roles/k3s/server.nix
+++ b/nixos/roles/k3s/server.nix
@@ -204,7 +204,8 @@ let
     rc=0
     for i in 1 2 3 4 5; do
       "$kubectl" get -n kube-system -o jsonpath='{.data.token}' \
-        secret "$secret" > "$tokendir/$user.b64"
+        secret "$secret" > "$tokendir/$user.b64" && \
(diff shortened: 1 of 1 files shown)


osnyx (approved):
LGTM

PR #765, opened 2023-08-10
[23.11] Adapt tests for upstream changes

sysvinit:
Upstream has changed the handling of guest network interfaces in the test library, which now unconditionally creates udev rules to force network interfaces to have specific names. Unfortunately these rules clash with and take precedence  over our own rules in the platform, which means that a variety of tests were failing because network interfaces did not have the expected names.

This PR refactors the existing tests to use the new `virtualisation.interfaces` option, which allows us to explicitly specify the desired name for each interface.

## Security implications

  - The platform test suite should be kept compatible with the upstream test functionality to ensure that tests continue to run correctly
  - Tested manually in Hydra that tests previously disabled due to network-related failures now run again correctly.


dpausp:
Some tests are still broken. Will be fixed later.

PR #766, opened 2023-08-10
gitlab-runner: fix docker executor by pinning older runner version

dpausp:
gitlab-runner doesn't work properly with Go 1.20.6. Talking to a docker daemon via UNIX socket fails with "invalid Host header". Pin our version of gitlab-runner to the one from release 2023_017 which uses the previous Go version until
https://gitlab.com/gitlab-org/gitlab-runner/-/issues/36051 is fixed in gitlab-runner.

PL-131694

## Release process

Impact:

Changelog:

* gitlab-runner: go back to previous runner version 16.1.0 which fixes the docker executor (PL-131694).

## Security implications

  - this version of gitlab-runner uses a Go version with security issues but we have to go back to the old version to make docker work again. The override should be removed as soon as possible when the problem is fixed upstream. 
  - tested on Gitlab staging VM

### DIFF
--- a/pkgs/overlay.nix
+++ b/pkgs/overlay.nix
@@ -50,6 +50,12 @@ in {
 
   innotop = super.callPackage ./percona/innotop.nix { };
 
+  # XXX: pinned to the latest 16.1.0 version using Go 1.20.5 (from release 2023_017)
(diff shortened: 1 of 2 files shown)


osnyx (approved):
Ewww, but fine as a temporary workaround.

PR #778, opened 2023-09-05
[21.05] extract osds into separate units

ctheune:
We had hosts run out of memory and the OSDs were all subsumed in the same cgroup which caused them to be killed together because systemd values consistency in a unit's state ...

There might be interesting cases here if NixOS decides to activate/deactivate anything, but hopefully the current setup will cause the "group unit" to catch this. I can imagine cases where this might not be the case, but I can't show that they actually exist currently.

Re PL-131646

## Release process

Impact: internal only

Changelog:
- move Ceph OSD processes into individual systemd service units per OSD
  - adjust `fc-ceph` accordingly
- rename service for managing all OSDs from `fc-ceph-osds.service` -> `fc-ceph-osds-all.service` 

## Security implications

  - [x] automatic transition of running osd services at unattended upgrade time
  - [x] availability: prevent systemd from stopping all osds at the failure of a single one, just because it considers all of them to belong together
  - [x] manually tested and documented proper automatic transition of running OSD services at unattended system upgrade
  - [x] extensive manual `fc-ceph` operations tests in dev cluster


osnyx (approved):
LGTM now, but to be fair I did most of the fine-tuning and adjustments of the core approach, so someone else needs to review as well.

PR #783, opened 2023-09-14
slurm: improve maintenance resilience: munge key setup, wait after ready_all

dpausp:
slurm: improve munged key setup, Restart=always

- fc-set-munge-key rewritten in Python with additional checks and atomic
  key update, done only if needed instead of overwriting it all the
  time.
  We experienced a failure caused by a race condition between
  munge reading the key and the key service running at the same time
  which used a naive shell script. The script used `>` to overwrite the
  existing key which left the file empty for a short amount of time,
  causing munged to fail on startup.
- Add key service to multi-user.target to start it on system switch.
- munged.service dependency on the key service is relaxed to "wants"
  because munged might still be able to start when the key is already
  correct.
- munged.service now uses Restart=always. Without munge, slurm doesn't
  work at all, better to try keeping it up at all times.

- wait 3 seconds after ready_all

## Release process

Impact:

Changelog:

 - slurm: make maintenance and munge key setup more resilient. We have experienced slurm downtime requiring manual intervention and later a false alarm once after a system update (PL-131740).

## Security implications

- availability: make sure that munged doesn't fail on startup and stays active as much as possible
- key security: must only be readable by the munge service
  - manual tests on test slurm cluster that key initialization and updating works. When munge is already running, munge is restarted by the key service.
  - checked munged.service if Restart=always is set.
  - checked that fc-slurm all-nodes ready still works and waits at the end

### DIFF
--- a/nixos/roles/slurm/fc-set-munge-key.py
+++ b/nixos/roles/slurm/fc-set-munge-key.py
@@ -1,0 +1,77 @@
+import filecmp
+import hashlib
+import json
+import os
+import shutil
+import sys
+from pathlib import Path
+
+import pystemd.systemd1
+from rich import print
+
+required_env_vars = {"encServicesPath", "mungeKeyFile"}
+missing_env_vars = required_env_vars - os.environ.keys()
+
+os.umask(0o0266)
+
+if missing_env_vars:
+    print(f"error: missing environment variables: {missing_env_vars}")
+    sys.exit(1)
+
+enc_services_path = Path(os.environ["encServicesPath"])
+if not enc_services_path.exists():
+    print(f"error: services files at {enc_services_path} is missing")
+    sys.exit(2)
+
+munge_key_path = Path(os.environ["mungeKeyFile"])
+
+with open(enc_services_path) as f:
+    enc_services = json.load(f)
+
+controller_service = [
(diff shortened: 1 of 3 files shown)


nixos/roles/slurm/fc-set-munge-key.py
```diff
@@ -0,0 +1,77 @@
...
+
+service_pw = controller_service[0]["password"] + "\n"
+munge_key = hashlib.sha256(service_pw.encode()).hexdigest()[:64]
+
+key_dir = munge_key_path.parent
+if not key_dir.exists():
+    print(f"init: creating key dir at {key_dir}")
+    key_dir.mkdir()
+
+tmp_key = key_dir / "tmp-set-munge-key"
```
osnyx, line 47: We still always use the same file name – this time of the tmpfile. So there could still be the very slight possibility of both processes having opened the same file concurrently and thus writing garbage.

Or do we rely on systemd here for ensuring only a single service run concurrently?

osnyx (approved):

nixos/roles/slurm/fc-set-munge-key.py
dpausp: I trust systemd here :)

PR #785, opened 2023-09-16
Save release ENC metadata, use it in motd

dpausp:
Release metadata is now read from enc.json and written to releases.json,
with the version label as key. The data can then be accessed via a
read-only NixOS option flyingcircus.platform.release which is used to
add the proper changelog URL and release name to the motd file. This
file can now also be accessed at /etc/motd.

This also adds the changelog URL and release name to the summary for
an UpdateActivity, which is used as maintenance request comment.

PL-131736

## Release process

Impact:

Changelog: none

## Security implications

  - known releases file should be owned by root and readable by anyone (contained metadata is known anyway), other ENC files should only be readable by root
  - checked on a test VM that motd is rendered correctly with a known release and that it doesn't break with local dev checkouts; checked file permissions in /etc/nixos

### DIFF
--- a/nixos/platform/shell.nix
+++ b/nixos/platform/shell.nix
@@ -12,6 +12,9 @@ let
 in
 {
   config = {
+
+    environment.etc.motd.text = config.users.motd;
+
     environment.interactiveShellInit = ''
       export TMOUT=43200
     '';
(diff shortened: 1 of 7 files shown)


nixos/platform/shell.nix
```diff
@@ -64,15 +67,16 @@ in
...
     in
     ''
       Welcome to the Flying Circus!
 
       Status:     https://status.flyingcircus.io/
       Docs:       https://doc.flyingcircus.io/
-      ${ opt (parameters ? release_changelog ) ("ChangeLog:  " + parameters.release_changelog)}
-      Release:    ${ opt has_release_name "${parameters.release_name} (" + config.system.nixos.label + opt has_release_name ")"}
+      Release:    ${opt isStableRelease "${release.release_name} (" + config.system.nixos.label + opt isStableRelease ")"}
+      ${opt isStableRelease ("ChangeLog:  " + release.release_changelog)}
```
osnyx, line 79: This can look weird if there is no Changelog (that field is optional). We can of course say that all stable releases *should* have a changelog and failing to provide one is the core issue.

osnyx (approved):

nixos/platform/shell.nix
dpausp: I leave it like that to show that something we expect is actually missing.

PR #787, opened 2023-09-19
[21.05] Improve kvm log cleanup

ctheune:

## Release process

Impact:

none

Changelog:

* Safely improve log file cleanup on KVM hosts for orphaned log files due to live migration. (PL-131715)

## Security implications

  * [x] we do not want to accidentally loose current logs
  * [x] owasp should fit (unfortunately we need root as the log files are root owned)
  * [x] quoting for potential user input (happens automatically by pathlib)
  * tested manually, multiple times with real data on kyle08 (by archiving the current logs and using a separate work directory from the actual log directory)

### DIFF
--- a/pkgs/fc/qemu/clean-logs.py
+++ b/pkgs/fc/qemu/clean-logs.py
@@ -1,0 +1,67 @@
+"""
+A helper to clean up log files from fc.qemu.
+
+The logrotate targets aren't smart enough because we have VMs that move around
+and thus would end up with missing files, but `missingok` doesn't help if the
+pattern doesn't match any more.
+
+In addition we may have files that have not been written to for a long time but
+do correspond to a VM still running here.
+
+So this script deletes log files, iff:
+
+1. the file is not open, AND
+2. it is older than 14 days.
+
+"""
+import datetime
+import pathlib
+import re
+import subprocess
+import time
+
+MAX_AGE = datetime.timedelta(days=14)
+VM_LOG_DIR = pathlib.Path("/var/log/vm")
+VM_LOG_NAME_PATTERN = re.compile(r"[0-9a-zA-Z]+\.log")
+
+# Find logs of running VMs
+
+print("Checking open files ...")
+
+lsof = subprocess.run(
+    ["lsof", "-F", "n", "+D", VM_LOG_DIR], capture_output=True
+)
+# Expected output:
+# $ lsof +D /var/log/vm -F n
+# p23052
+# n/var/log/vm/release2305dev01.log
+# p319664
+# n/var/log/vm/mailstubtest01.log
+# p321165
+# n/var/log/vm/services34.log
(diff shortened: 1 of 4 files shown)


pkgs/fc/qemu/clean-logs.py
```diff
@@ -0,0 +1,54 @@
...
+    open_files.add(path)
+
+print("Checking for files to unlink...")
+
+mtime_cutoff = time.time() - MAX_AGE.total_seconds()
+for logfile in VM_LOG_DIR.glob("*"):
+    if str(logfile) in open_files:
+        continue
+    is_old = logfile.stat().st_mtime < mtime_cutoff
+    is_empty = not logfile.stat().st_size
```
osnyx, line 60: I'm not entirely sure about the reason for this condition: Is this there to leave old non-empty files in the responsibility of logrotate?

pkgs/fc/qemu/clean-logs.py
```diff
@@ -0,0 +1,54 @@
...
+
+MAX_AGE = datetime.timedelta(days=14)
+VM_LOG_DIR = pathlib.Path("/var/log/vm")
+VM_LOG_NAME_PATTERN = re.compile(r"[0-9a-zA-Z]+\.log")
+
+# Find logs of running VMs
+
+print("Checking open files ...")
+
+lsof = subprocess.run(["lsof", "+D", VM_LOG_DIR], capture_output=True)
```
osnyx: We already discussed out-of-band that this command is wrong and its output is not suited to the following parsing code.

osnyx (changes requested):

pkgs/fc/qemu/clean-logs.py
```diff
@@ -0,0 +1,67 @@
...
+# n/var/log/vm/mailstubtest01.log
+# p321165
+# n/var/log/vm/services34.log
+# p332886
+# n/var/log/vm/services08.log
+# p333454
+
+open_files = set()
+
+for open_file in lsof.stdout.decode("ascii").splitlines():
```
osnyx, line 48: In case we ever get non-ascii VMs, this might break. An `errors="ignore"` makes this more resilient.

Also, encoding could already be specifid to `subprocess.run`, but I guess that's a matter of taste.

osnyx (changes requested):

pkgs/fc/qemu/clean-logs.py
ctheune: I did that on purpose. Ignoring errors here means that we won't properly match filenames. I decided that we really do not expect non-ASCII to appear here due to other impacts in the chain.

osnyx: ack, let it crash then, we'll hopefully notice.

osnyx (approved):

PR #805, opened 2023-10-11
Build more packages on Hydra and add them to important packages

dpausp:
When rolling out the last release with the glibc (mass rebuild) changes, we noticed some packages
that are used in production and were not covered by our Hydra jobsets and/or important-packages.json

Add some packages for building on hydra

firefox and awscli are used in production and have a long build time.
Adding python3Packages.boto3 would also make sense as hydra-built
package but the code doesn't support creating jobs from nested package
sets currently.

gitlab: enable registry in test

This test makes sure that gitlab-container-registry is built by our
Hydra, if neccessary and serves as a smoke test for the registry
service.

lamp package test: use the correct nixpkgs, not the bootstrap nixpkgs

Hydra should build/test the lamp packages for the nixpkgs version pinned
by fc-nixos. Before, the test pulled in pkgs from the bootstrap nixpkgs
which is often different as it is updated separately in Hydra jobset
settings.

This change is mostly for completeness, the lamp VM tests already did the right
thing and make sure that the packages are built on Hydra.

PL-131831

## Release process

Impact:

Changelog:

-awscli, firefox: track version changes in our release change logs. They are also built by our Hydra (cache) now to make sure that machines don't have to compile these if the upstream NixOS cache doesn't have them, yet. We noticed this when rolling out the last release which caused a mass rebuild of packages which was due to the glibc security patch (PL-131831).

## Security implications

  -  reduce risk of availability problems on production systems when systems try to compile expensive packages
  - track version changes and security fixes of important packages
  - gitlab test still runs, other than that no manual testing needed


osnyx (approved):
I've already stated my concerns about fallbak imports in function headers, but as the lamp test already had such a construct before I'm also okay with keeping that pattern there for now.

PR #806, opened 2023-10-11
[21.05] fc-ceph: improve osd safety check UX

osnyx:
In accordance to everyday operations usage, reduce the strictness of the
default safety check when destroying OSDs. By default, we use
`ok-to-stop` which allows degraded data redundancy as long as the
cluster stays IO-operational.
The old `safe-to-destroy` check behaviour is available under the new
flag `--strict-safety-check`.

This overall increases the UX safety of common operations tasks, as we accept temporarily degrading the redundancy for rebuilds as long as data stays available. Previously, we had to disable safety checks altogether, now the default safety check allows these operations but still prevents us from breaking availability.

The check mechanism has also been added to `fc-ceph osd deactivate`. The `reactivate` operation does its own checking anyways and is not affected.

## Release process

Impact: internal only

Changelog:
- fc-ceph:
  - reduce strictness of default osd destruction safety checks
  - rename flags `--unsafe-destroy` -> `--no-safety-check`
  - add safety checks for deactivate

## Security implications

  - The UX of our tooling should be designed along our common operation actions. Safety checks should support the operator from accidentally entering dangerous states, while not influenceing common operations too much. 
  - no new (known) regressions must be introduced
  - [x] extended test cases to cover new functionality
  - [x] manually tested in dev cluster

### DIFF
--- a/pkgs/fc/ceph/src/fc/ceph/osd/nautilus.py
+++ b/pkgs/fc/ceph/src/fc/ceph/osd/nautilus.py
@@ -184,7 +184,8 @@ class OSDManager(object):
                     thread = threading.Thread(
                         target=lambda: self._deactivate_single(
                             osd, as_systemd_unit, flush
-                        )
+                        ),
+                        name=id_,
                     )
                     thread.start()
                     threads.append(thread)
@@ -193,6 +194,9 @@ class OSDManager(object):
 
             for thread in threads:
                 thread.join()
+
+            deactivated_osds = ", ".join([str(t.name) for t in threads])
+            print("Successfully deactivated OSDs", deactivated_osds)
 
     def _deactivate_single(
         self,
@@ -492,7 +496,7 @@ class GenericOSD(object):
                 print(
                     # fmt: off
                     "OSD not safe to destroy:", e.stderr,
-                    "\nTo override this check, remove th `--strict-safety-check` flag. "
+                    "\nTo override this check, remove the `--strict-safety-check` flag. "
                     "This can lead to reduced data redundancy, still within safety margins."
                     # fmt: on
                 )
--- a/pkgs/fc/ceph/src/fc/ceph/osd/nautilus.py
+++ b/pkgs/fc/ceph/src/fc/ceph/osd/nautilus.py
@@ -153,6 +153,7 @@ class OSDManager(object):
         self,
         ids: str,
         unsafe_destroy: bool,
+        strict_safety_check: bool,
         force_objectstore_type: Optional[str] = None,
     ):
         ids = self._parse_ids(ids, allow_non_local=f"DESTROY {ids}")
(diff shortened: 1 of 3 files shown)


pkgs/fc/ceph/src/fc/ceph/osd/nautilus.py
```diff
@@ -469,26 +487,61 @@ def _create_crush_and_auth(self, volume, crush_location):
...
             except CalledProcessError as e:
                 print(
                     # fmt: off
                     "OSD not safe to destroy:", e.stderr,
-                    "\nTo override this check, specify `--unsafe-destroy`. This can "
+                    "\nTo override this check, remove the `--strict-safety-check` flag. "
+                    "This can lead to reduced data redundancy, still within safety margins."
+                    # fmt: on
+                )
+                # do we have some generic or specific error return codes?
```
ctheune: You could look at the typical errno ones. Maybe EBUSY or something similar?

pkgs/fc/ceph/src/fc/ceph/osd/nautilus.py
```diff
@@ -469,26 +487,61 @@ def _create_crush_and_auth(self, volume, crush_location):
 
         run.ceph("osd", "crush", "add", self.name, str(weight), crush_location)
 
-    def purge(self, unsafe_destroy):
-        """Deletes an osd, including removal of auth keys and crush map entry"""
+    @staticmethod
+    def run_safety_check(ids: List[int], strict_safety_check: bool):
```
ctheune, line 491: We could turn this into a decorator to simplify the individual use cases so that this code could be a wrapper around the protected actions

```
@safety_check
def destroy(...)
   ...

def safety_check(f):
    def wrapped(self, *args, **, no_safety_check, strict_safety_check, **kw):
        # ... do the safety check here
        f(self, *args, **kw)
    return wrapped
```

There are specific helpers in the stdlib to assist in the wrapping (I think there's a decorator decorator) so that the wrapped methods get proper naming, etc.

ctheune (changes requested):
Generally looks good with some suggestions.

pkgs/fc/ceph/src/fc/ceph/osd/nautilus.py
osnyx: *edit: this remark did not make any sense*

osnyx: The issue here is that both the safety check as well as the destroy/deactivate functions require the osd ID or list of osd IDs as an argument.
I do not immediately see a simple way to refactor this in a non-hacky way (e.g. accessing the kwargs of the wrapped function) and thus do not see the effort being worth it. Especially as the usage semantics between destroy and deactivate are slightly different as well.

ctheune (approved):

ctheune:
Happy with this. However, hydra is red?

PR #811, opened 2023-10-16
agent: maintenance monitoring

dpausp:
agent: maintenance check for Sensu, more metrics

- fc-maintenance check is run by Sensu every 3 minutes.
  and can also used by sudo-srv, service and admins groups with sudo.
- Check for stuck requests (existing for more than 7 days).
- Report maintenance duration and run time of running requests.
- enter_maintenance now creates a marker file at
  /var/spool/maintenance/in_maintenance, leave_maintenance deletes it.
- Alert when maintenance mode is activated for a long time
  (warning: 30min, critical 60min)
- Attempts are now created and saved before the activity starts
  so we can use it to measure time while it is running.
- More request metrics
  - count running
  - count waiting_for_schedule (created, but not scheduled)
  - count scheduled
  - time of next due request
- Add current maintenance metrics
  - duration of maintenance
  - current request run time

agent: do not acquire lock when getting metrics

We also want metrics during a maintenance run.
It's generally safe to call ReqManager methods as long as
they don't modify request state.

__enter__ now only acquires the lock. Reading the config is
moved to __init__ so that methods can read config without
locking.

This also includes a fix to include pytest in the Python dev
environment again as the pytestCheckHook doesn't propagate pytest
itself.

There's a possibility that loading requests will lead to an error
which means that a "read-only" method might change state because
requests that failed to load are archived immediately. This should
be fixed later as it can be annoying for testing/debugging.

## Release process

Impact:

Changelog:

## Security implications

  - The change is about discovering problems with automated maintenance, like maintenance mode being activated for too long or requests not being executed for a long time. These problems could affect availability (we don't send alerts when machines are in maintenance mode) and security in general when updates are not rolled out in time.
  - Checked metrics and verified check output on a test VM in various phases of the maintenance workflow (no requests, (multiple) request(s) waiting to be scheduled or already scheduled, running requests)

### DIFF
--- a/pkgs/fc/agent/fc/maintenance/reqmanager.py
+++ b/pkgs/fc/agent/fc/maintenance/reqmanager.py
@@ -16,10 +16,12 @@ from typing import NamedTuple
 
 import fc.maintenance.state
 import fc.util.directory
+import pytz
 import rich
 import rich.syntax
 import structlog
 from fc.maintenance.activity import RebootType
+from fc.util.checks import CheckResult
 from fc.util.time_date import format_datetime, utcnow
 from rich.table import Table
 
@@ -97,6 +99,7 @@ class ReqManager:
         self.requestsdir = self.spooldir / "requests"
         self.archivedir = self.spooldir / "archive"
         self.last_run_stats_path = self.spooldir / "last_run.json"
+        self.maintenance_marker_path = self.spooldir / "in_maintenance"
         for d in (self.spooldir, self.requestsdir, self.archivedir):
             if not d.exists():
                 os.mkdir(d)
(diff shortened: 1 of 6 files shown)


pkgs/fc/agent/fc/maintenance/reqmanager.py
```diff
@@ -870,6 +907,107 @@ def reboot_and_exit(self, requested_reboots):
...
+
+        metrics = self.get_metrics()
+
+        # Are we in maintenance mode? Check maintenance duration and add info
+        # about the currently running request, if any.
+        if maint_duration := metrics["in_maintenance_duration"]:
+            # 15 minutes as estimate for total request run time is used here.
+            # We could calculate it from the actual request estimates but it
+            # wouldn't change much in reality and having a fixed value is
+            # better for alerting, I think.
```
osnyx, line 923: Okay with this, but let's keep an eye on how this works out in production.

pkgs/fc/agent/fc/maintenance/reqmanager.py
```diff
@@ -870,6 +907,107 @@ def reboot_and_exit(self, requested_reboots):
...
+                    "seconds ago."
+                )
+        else:
+            ok_info.append("Machine is in service.")
+
+        longest_in_queue_hours = (
+            metrics["request_longest_in_queue_duration"] / 3600
+        )
+        # XXX: We probably want to get the lead time from the directory for a
+        # better threshold.
```
osnyx, line 954: Does this have a corresponding ticket yet?

osnyx (approved):

PR #814, opened 2023-10-19
[22.11] opensearch: auto-migrate existing elasticsearch data

dpausp:
(not on 23.05 because we don't have the elasticsearch roles there anymore).

Only works when all indices have been indexed on ES 7.*
ES 6 indices need an re-index before migrating to OpenSearch.

Preserves the old ES data dir in case something goes wrong. Files are copied using reflinks so data is only duplicated when data files are changed by OS. The old dir can be deleted safely when OS is working as expected.

PL-131553

## Release process

Impact:

Changelog:

- opensearch: implement automated migration from ElasticSearch 7. Note that migrations from ES6 require re-indexing on ES7 before migrating to OpenSearch. See the OpenSearch role docs for details (PL-1315530).

## Security implications

  - have safe, automated data migration where possible to avoid manual interactions with elevated privileges 
  - checked that migration from ES7 to OpenSearch works on a test VM and that the migration code doesn't overwrite existing data. Tried with Indices created on ES7 and indices created on ES6/re-indexed on ES7.

### DIFF
--- a/doc/src/elasticsearch.md
+++ b/doc/src/elasticsearch.md
@@ -14,8 +14,13 @@ they expect the unfree versions of Elasticsearch. Note that *x-pack* features
 are not available in the free version.
 
 Elastic doesn't provide updates to the 7.x line anymore so 7.10.2
-will be the last available version. We are planning to move to OpenSearch
-instead which is a fork of Elasticsearch 7.10.2.
+will be the last available version.
+
+:::{warning}
+This is the last platform version that supports Elasticsearch.
+Migrate to {ref}`nixos-opensearch` which is a fork of Elasticsearch
+7.10.2.
+:::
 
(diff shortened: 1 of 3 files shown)


doc/src/elasticsearch.md
```diff
@@ -131,6 +131,12 @@ Nodes should be upgraded one at a time to ensure continous operation of the
 cluster. Upgrading nodes is done by changing the role of the machine to
 *elasticsearch7*.
 
+### Upgrade/Migration to OpenSearch
+
```
dpausp, line 140: mention that ES is unsupported on newer platforms and that it must be migrated to OS before upgrading the system.

osnyx (approved):

osnyx (approved):

PR #815, opened 2023-10-19
agent: resume interrupted requests, other related improvements

dpausp:
agent: implement resuming interrupted requests

Before, requests that had been persisted in state "running" were just
ignored by the ReqManager and had to be cleaned up manually.

Now, the default behaviour defined in Activity.resume() is to put the
activity into an error state (special value EXT_INTERRUPTED = 80) to
avoid re-running non-idempotent activity that could be unsafe to run
again.

Idempotent acivities (UpdateActivity, VMChangeActivity) are just
run again.

The legacy reboot activity is a special case: it was usually interrupted
by the planned reboot triggered in activity.run() so we just mark it
as successful to clean it up.

agent: handle legacy reboot activities properly

Old reboot activities rebooted during activity execution which we don't
want to do anymore with the new ReqManager design. It's possible that
they have been created before switching to new agent code which tries to
execute them, leading to an unexpected reboot. Instead, we now use the
new implementation which is basically a no-op and defers rebooting to
the ReqManager.

VMChangeActivity: do not check for reboot when value is None

Avoids unneccessary work and log messages that make no sense.

agent: make fc-maintenance --run-all-now safer

Before, "run all" really meant this, so even finished requests executed
when `--run-all-now` was given. The flag is now interpreted to include
pending requests into the runnable request list but not more. If you
really want every request to run (which is dangerous because requests
might not be idempotent when they succeeded, for example), add
--force-run which also ignores postpone and tempfail situations caused
by maintenance enter commands.

PL-131849

## Release process

Impact:

Changelog:

- (internal, I guess)

## Security implications

Mostly changes to automated actions, no new capabilities for normal users.

  - reduce risk of unintended, possibly dangerous behaviour when special flags like --run-all-now are used   
  - avoid cluttering monitoring and logs with things that could be cleaned up automatically  
  - checked on a test VM that --run-all-now and --force-run work as expected (tried with an ShellScriptActivity which was already in State.success. --run-all-now didn't run it again, but --force-run did)
  - checked on a test VM that running, non-idempotent activites are put in a failure state (tried with an ShellScriptActivity in State.running) and that legacy reboot requests are a no-op now.
  - automated tests cove resuming behaviour of activities and general execution of requests

### DIFF
--- a/pkgs/fc/agent/fc/maintenance/cli.py
+++ b/pkgs/fc/agent/fc/maintenance/cli.py
@@ -155,6 +155,11 @@ def run(run_all_now: bool = False, force_run: bool = False):
     If you still want to execute requests even if a maintenance enter command returned
     with EXIT_TEMPFAIL or EXIT_POSTPONE, add the --force-run flag.
 
+    WARNING: ignoring the result of enter commands can be dangerous!
+
+    The --force-run flag also runs requests in state 'success' again when they are still
+    in the queue after a recent system reboot.
+
     After executing all runnable requests, requests that want to be postponed
     are postponed (they get a new execution time) and finished requests
     (successful or failed permanently) moved from the current request to the
--- a/pkgs/fc/agent/fc/maintenance/activity/__init__.py
+++ b/pkgs/fc/agent/fc/maintenance/activity/__init__.py
@@ -2,6 +2,7 @@
 from enum import Enum
 from typing import NamedTuple, Optional
 
+import fc.maintenance.state
 import structlog
 from fc.maintenance.estimate import Estimate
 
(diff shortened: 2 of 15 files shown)


pkgs/fc/agent/fc/maintenance/cli.py
```diff
@@ -155,6 +155,11 @@ def run(run_all_now: bool = False, force_run: bool = False):
     If you still want to execute requests even if a maintenance enter command returned
     with EXIT_TEMPFAIL or EXIT_POSTPONE, add the --force-run flag.
 
+    WARNING: ignoring the result of enter commands can be dangerous!
+
+    The --force-run flag also runs requests in state 'success' again when they are still
```
dpausp, line 160: clarify: can the flags be used separately?

dpausp: and: should --force-run affect tempfail/postpone decisions in enter_maintenance

dpausp: yes, they can be used separately. `--force-run` already affects tempfail/postpone decisions (added in 06361039f0450deda60001a23fe7d2a868d26a03)

osnyx (approved):

PR #816, opened 2023-10-19
Fix various issues with release info handling

dpausp:
- Don't break system build when release changelog is missing for a stable release, use the changelog overview page as fallback value.
- Update release info from fresh ENC data, not previous ENC data loaded from disk.
- Handle null/None values for `release_changelog` in ENC parameters and treat them as empty string.

PL-131852

## Release process

Impact:

Changelog:

## Security implications

  - tested on a VM with dev checkout, rolling release channel and production channel

### DIFF
--- a/pkgs/fc/agent/fc/util/enc.py
+++ b/pkgs/fc/agent/fc/util/enc.py
@@ -222,25 +222,38 @@ def get_release_info(log, enc):
     release_name = params.get("release_name")
     known_releases = [r["release_name"] for r in releases.values()]
 
+    # We treat named releases as immutable, which means that we won't update
+    # release metadata objects once they are written to `release_info_path`.
+    # It's possible to change metadata for a build version by adding a new
+    # named release with different changelog URL (the only metadata field for
+    # now which is user-provided in the directory) which will overwrite the
+    # existing value.
     if release_name and release_name not in known_releases:
         environment_url = params["environment_url"]
+        # version example: 23.05.2820.18d29027
         version = nixos.channel_version(environment_url)
-        releases[version] = {
+        release_metadata = {
             "environment": params["environment"],
             "environment_url": environment_url,
             "first_seen_at": utcnow().isoformat(),
             "release_name": release_name,
-            "release_changelog": params.get("release_changelog", ""),
+            "release_changelog": params.get("release_changelog") or "",
         }
+        log.debug("release-info-add", **release_metadata)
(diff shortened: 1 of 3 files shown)


pkgs/fc/agent/fc/util/enc.py
```diff
@@ -222,25 +222,38 @@ def get_release_info(log, enc):
     release_name = params.get("release_name")
     known_releases = [r["release_name"] for r in releases.values()]
 
+    # We treat named releases as immutable, which means that we won't update
+    # release metadata objects once they are written to `release_info_path`.
+    # It's possible to change metadata for a build version by adding a new
```
osnyx, line 227: After going through the whole enc retrieval flow I can confirm that updating the release metadata will replace the existing data of a release with the same version.
Good.

pkgs/fc/agent/fc/util/enc.py
```diff
@@ -294,9 +306,14 @@ def update_enc(log, tmpdir, enc_path):
...
+    # new enc.json and other inventory files.
+    pre_enc = load_enc(log, enc_path)
+    update_inventory(log, pre_enc)
     # Reload to pick up possibly updated enc.json.
     enc = load_enc(log, enc_path)
     update_enc_nixos_config(log, enc, enc_path)
     write_system_state(log)
+    # Placed at the end because failures here are the least problematic ones.
+    # We can continue with missing release info data.
+    write_release_info(log, enc)
```
osnyx, line 319: And this is the change that prevents us from persisting the release metadate before we know it can be instantiated correctly. Also good.

pkgs/fc/agent/fc/util/tests/test_enc.py
```diff
@@ -43,6 +43,7 @@ def test_initialize_enc_should_not_crash_when_initial_data_missing(
     assert log.has("initialize-enc-initial-data-not-found")
 
 
+@unittest.mock.patch("fc.util.enc.write_release_info")
```
osnyx, line 46: Is it possible to add some test cases with variations of valid but potentially plausible ENC json payloads?

osnyx (approved):
Some further test coverage of exactly the problematic points we're trying to cover could be useful.

dpausp:
We decided to not go forward with this change. We can keep the current state for now but the release info feature will be simplified which makes most of the changes in this PR obsolete. The old_enc/enc distinction in enc.py might still be useful, though.

PR #818, opened 2023-10-23
agent: reduce directory errors, easier debugging, various improvements

dpausp:
A collection of commits to reduce the need for sudo for non-invasive fc-maintenance commands, improve error handling, logging, avoiding unneccessary directory API calls and retrying failed calls. See commit messages for details.

## Release process

Impact:

Changelog:

 * `fc-maintenance` commands for viewing requests (`show` and `list`) can now be used without sudo by all users. The same applies to the commands used for monitoring and telemetry (`check` and `metrics`).
 * `fc-maintenance show <request ID prefix>`: now has improved output and also works for archived maintenance requests while active requests are preferred when there are multiple matches.
 * Command output generated during execution of an update activity is now logged to separate files in `/var/log/fc-agent`, like for `fc-manage` calls.
* Unhandled errors in `fc-manage` and `fc-maintenance` are now logged properly to the system journal and can be viewed with `journalctl SYSLOG_IDENTIFIER=fc-agent`.

## Security implications

This change affects sudo rules for fc-maintenance commands but the set of commands that can be run and the allowed groups have not been extended.

  - reduce need to use sudo for commands that only access data readable by any user  
  - make sure that invasive commands cannot be run by normal users without sudo and that sudo permissions are limited properly (admin, sudo-srv and service may only run fc-maintenance (-v) delete without password)
  - checked on a test VM that fc-maintenance commands work as intended, checked sudo rules, checked logs
  - automated test have been extended

### DIFF
--- a/pkgs/fc/agent/default.nix
+++ b/pkgs/fc/agent/default.nix
@@ -35,6 +35,20 @@ let
     buildInputs = [ pytest structlog ];
   };
 
+  stamina = py.buildPythonPackage rec {
+    pname = "stamina";
+    version = "23.1.0";
+    format = "pyproject";
+
+    src = fetchPypi {
+      inherit pname version;
+      hash = "sha256-sWzj1S1liqdduBP8amZht3Cr/qkV9yzaSOMl8qeFR4Y=";
+    };
+
+    nativeBuildInputs = with py; [ hatchling hatch-vcs hatch-fancy-pypi-readme ];
+    propagatedBuildInputs = with py; [ structlog tenacity typing-extensions ];
+  };
+
 in
 buildPythonPackage rec {
   name = "fc-agent-${version}";
@@ -68,6 +82,7 @@ buildPythonPackage rec {
     py.structlog
     py.typer
     py.pyyaml
+    stamina
     util-linux
   ] ++ lib.optionals stdenv.isLinux [
     dmidecode
--- a/pkgs/fc/agent/default.nix
+++ b/pkgs/fc/agent/default.nix
@@ -15,6 +15,7 @@
 , xfsprogs
 , pytest
 , structlog
+, enableSlurm ? stdenv.isLinux
 }:
 
 let
@@ -72,10 +73,11 @@ buildPythonPackage rec {
     dmidecode
     gptfdisk
     multipath-tools
-    py.pyslurm
     py.pystemd
     py.systemd
     xfsprogs
+  ] ++ lib.optionals enableSlurm [
+    py.pyslurm
   ];
   dontStrip = true;
   doCheck = true;
--- a/pkgs/fc/agent/fc/maintenance/reqmanager.py
+++ b/pkgs/fc/agent/fc/maintenance/reqmanager.py
@@ -59,6 +59,14 @@ def require_directory(func):
     return with_directory_connection
 
 
+class RequestsNotLoaded(Exception):
+    def __init__(self):
+        super().__init__(
+            "Accessing requests is not possible because ReqManager has not been "
+            "initialized."
+        )
+
+
 class PostponeMaintenance(Exception):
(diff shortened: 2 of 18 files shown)


pkgs/fc/agent/fc/util/directory.py
```diff
@@ -18,6 +20,40 @@ def __init__(self, url, *args, **kw):
...
+        )
+        return retry(wrapper)
+
+    def __repr__(self):
+        """ServerProxy.__repr__ leaks the directory password, override it."""
+        host = self._ServerProxy__host.split("@")[1]
+        handler = self._ServerProxy__handler
+        return f"Directory API ({host}{handler})"
+
+
```
ctheune, line 67: retry behaviour should be customizable here. default, based on our code base, IMHO should be `off`

ctheune (commented):
Looks good in general, see the small comment about making retry behaviour configurable.

pkgs/fc/agent/fc/maintenance/tests/test_cli.py
```diff
@@ -43,9 +43,11 @@ def app_main_args(tmp_path, agent_maintenance_config):
 
 
 @pytest.fixture
-def invoke_app(app_main_args):
+def invoke_app(app_main_args, monkeypatch):
     runner = typer.testing.CliRunner()
 
+    monkeypatch.setattr("os.getuid", lambda: 0)
```
osnyx: As this now causes all tests with that fixture to be run as root, it might make sense to codify this in the fixture name.
I understand that this simplifies most of the test cases or is even explicitly intended. Yet, we need to be careful to always be aware of this when writing test cases where the invoking user is relevant.
I see that in `test_invoke_metrics` you explicitly do __not__ use this fixture, as running it as non-root is the intended case there.

pkgs/fc/agent/default.nix
```diff
@@ -34,6 +35,20 @@ let
...
+    version = "23.1.0";
+    format = "pyproject";
+
+    src = fetchPypi {
+      inherit pname version;
+      hash = "sha256-sWzj1S1liqdduBP8amZht3Cr/qkV9yzaSOMl8qeFR4Y=";
+    };
+
+    nativeBuildInputs = with py; [ hatchling hatch-vcs hatch-fancy-pypi-readme ];
+    propagatedBuildInputs = with py; [ structlog tenacity typing-extensions ];
```
osnyx, line 49: Just a note, as this confused me: We rely on name scope shadowing here, as pkgs.tenacity [is a graphical audio editor](https://tenacityaudio.org/) instead of the python retry library.
Totally fine to do so, it can be a matter of taste to make this distinction clear by still explicitly writing `py.tenacity`.

pkgs/fc/agent/fc/maintenance/reqmanager.py
```diff
@@ -640,7 +766,7 @@ def _handle_enter_tempfail(
                 "run-all-now-tempfail",
                 _replace_msg=(
                     "Run all mode requested but a maintenance enter "
-                    "command had a temporary failure."
+                    "command had a temporary) failure. "
```
osnyx: Typo? I am missing the opening bracket.

osnyx (changes requested):
LGTM overall, but with a few minor changes.
Also this is still a draft, requiring fixing the pre-commit failures and the usual PR documentation.

pkgs/fc/agent/default.nix
dpausp: I think it's ok here because `with py;` is close to the place where the package is used and adding another explicit `py.tenacity` looks strange . In general, qualifying packages is a good idea when packages from multiple sources are mixed in a list.

osnyx (approved):

PR #825, opened 2023-11-08
varnish: simple multihost configuration

PhilTaken:

A simple multihost configuration for varnish that separates the configuration into host-specific modules that are being loaded during startup and can be hot-reloaded if the configuration changes.

For backwards-compatibility reasons, the /etc/local/varnish/default.vcl will be used as the varnish config if it exists and there is no conflicting nix configuration. 

If both a /etc/local/varnish/default.vcl and flyingcircus.services.varnish.virtualHost Nix configurations are found, the rebuild will fail with an appropriate error message. 

## Release process

Impact:  

* Varnish will be restarted, so caches will be cold.

Changelog: Added multi-host functionality to the webproxy role via `flyincircus.services.varnish` 

## Security implications

The defaults work the same, specific host configuration is not affected since the only relevant change the configuration being split up into multiple files.

The tests have been adjusted to test the relevant configuration changes.

### DIFF
--- a/doc/src/webproxy.md
+++ b/doc/src/webproxy.md
@@ -26,16 +26,16 @@ and how you are used to configure, start, stop and maintain these packages.
 
 ### Role configuration
 
-The role currently supports two different ways to configure varnish.
+The role currently supports two different ways to configure Varnish.
 Please note that all configuration has to be performed as a service user.
 
 The recommended way is to use Nix. For an overview of the available configuration
 options, see the Varnish module in our [infrastructure repository](https://github.com/flyingcircusio/fc-nixos).
-As with all nixos modules, put your configuration into an appropriately named file
+As with all NixOS modules, put your configuration into an appropriately named file
 in the {file}`/etc/local/nixos` directory, e.g. {file}`/etc/local/nixos/varnish.nix`.
 
-You can also put your verbatim varnish configuration into {file}`/etc/local/varnish/default.vcl`.
-Please note that this way of configuring varnish is deprecated and will likely
+You can also put your verbatim Varnish configuration into {file}`/etc/local/varnish/default.vcl`.
+Please note that this way of configuring Varnish is deprecated and will likely
 be removed in the future.
 
 ### Monitoring
--- a/doc/src/webproxy.md
+++ b/doc/src/webproxy.md
@@ -12,7 +12,7 @@ and how you are used to configure, start, stop and maintain these packages.
 - **configuration file locations:**
 
   Since we use NixOS, configuration files have to be edited in
-  {file}`/etc/local/varnish`, followed by a NixOS rebuild which copies them into
+  {file}`/etc/local/nixos`, followed by a NixOS rebuild which copies them into
   the Nix store and activates the new configuration. To do so, run the command
   {command}`sudo fc-manage --build`.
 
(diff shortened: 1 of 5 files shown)


doc/src/webproxy.md
```diff
@@ -26,9 +26,17 @@ and how you are used to configure, start, stop and maintain these packages.
...
 ### Role configuration
 
-Your custom configuration goes to
-{file}`/etc/local/varnish/default.vcl`. Please note that all
-configuration has to be performed as a service user.
+The role currently supports two different ways to configure varnish.
+Please note that all configuration has to be performed as a service user.
+
+The recommended way is to use Nix. For an overview of the available configuration
+options, see the Varnish module in our [infrastructure repository](https://github.com/flyingcircusio/fc-nixos).
```
ctheune, line 33: deep link would be good. what's the status on the option search?

nixos/services/varnish/default.nix
```diff
@@ -0,0 +1,130 @@
...
+          .port = "80";
+        }
+
+        sub vcl_recv {
+          return (synth(503, "Varnish is starting up"));
+        }
+      '' else cfg.config;
+    };
+
+    systemd.services.varnish = let
```
ctheune, line 129: ok, i'd like to double check the whole regular start + reload case because this isn't immediately obvious WHY this works ... :)

tests/webproxy.nix
```diff
@@ -25,18 +31,21 @@ import ./make-test-python.nix ({ pkgs, ... }:
...
     webproxy.wait_for_unit("varnishncsa.service")
+    webproxy.wait_for_unit("helloserver.service")
 
-    webproxy.execute("""
-        echo 'Hello World!' > hello.txt
-        ${pkgs.python3.interpreter} -m http.server 8080 >&2 &
-        """)
-
-    url = 'http://localhost:8008/hello.txt'
+    url = 'http://localhost:${builtins.toString varnishport}/hello.txt'
```
ctheune, line 73: Ok, I think we should cycle through a few specialisations to show that adding/changing/removing vhosts works.

ctheune (changes requested):
Looks good in general. Lets go through it together on Monday.

doc/src/webproxy.md
PhilTaken: it just entered draft status: https://github.com/flyingcircusio/fc-nixos/pull/828

I'm just waiting for my test vm to build and all hydra tests to pass

tests/webproxy.nix
PhilTaken: definitely, will look into it

nixos/roles/webproxy.nix
```diff
@@ -57,23 +35,29 @@ in
...
         '';
-        "local/varnish/default.vcl.example".text = vclExample;
-        "current-config/varnish.vcl".source = configFile;
       };
 
       flyingcircus.services.sensu-client.checks = {
         varnish_status = {
           notification = "varnishadm status reports errors";
-          command = "${cfg.package}/bin/varnishadm -n ${cfg.stateDir} status";
+          command = "${cfg.package}/bin/varnishadm";
```
ctheune: the `status` is still needed

ctheune (changes requested):
Overall looking good, test coverage as mentioned would be great.

tests/webproxy.nix
```diff
@@ -47,9 +81,10 @@ import ./make-test-python.nix ({ pkgs, ... }:
     with subtest("varnishncsa should log requests"):
         webproxy.wait_until_succeeds(f"{curl} && grep -q 'GET {url} HTTP/' /var/log/varnish.log")
 
-    with subtest("changing config and reloading should activate new config"):
-        webproxy.execute('ln -sf /etc/local/varnish/newconfig.vcl /etc/current-config/varnish.vcl')
-        webproxy.succeed('systemctl reload varnish')
-        webproxy.succeed('varnishadm -n /var/spool/varnish/webproxy vcl.list | grep active | grep -q newconfig')
+    with subtest("varnish pid should be the same across small configuration changes"):
```
ctheune, line 84: Getting there ... :)

I'd appreciate if you could double check that the configs (as visible in varnishadm) are as expected and are changed in the reload.

ctheune (changes requested):
Small final test improvement, please. :)

ctheune (approved):

PR #826, opened 2023-11-12
Build all packages from the important packages list

dpausp:
release: build important packages on Hydra

For each package name from important-packages.json we now create a Hydra
job with the `importantPackages.` plus the package name as job name.
This also works for nested sets like `python3Packages.supervisor`.

This makes sure that we have these packages in our cache even if we
differ from upstream (changes to our nixpkgs fork, for example) or
upstream hydra didn't build them, yet.

release: fix excluding nested packages, refactor modified pkg jobs

Jobs for "modified" packages from our overlay (it's all packages from
the overlay actually, not only the modified ones) are now generated from
the same code path as the jobs for important packages.

To be able to use that code, packages from the overlay are turned into a
list of "dotted names" like `adns` or `fc.blockdev`. This allows us to
exclude specific packages from nested sets (only `fc` at the moment)
using the dotted name. The previous code only supported excluding
things on the top level of the overlay.

The exclusion list is now used for both important and overlay packages.

PL-131581

## Release process

Impact: 

Changelog: (internal)

## Security implications

  - only affects Hydra jobs, no impact on machines 
  - checked that jobs build on Hydra


dpausp:
Will be rebased + merged after the nixpkgs update

release/default.nix
```diff
@@ -99,34 +99,53 @@ let
     }
   ];
 
-  modifiedPkgNames = attrNames (import ../pkgs/overlay.nix pkgs pkgs);
-
+  # Recursively searches a package attrset for derivations and returns a list
+  # of attribute paths as "dotted names", like "adns" or "fc.agent".
+  getDottedPackageNames = attrs: prevPath:
```
osnyx: While I think I understand what `prevPath` is used for, neither its name semantics (what does the abbreviation mean) nor its type (list of strings) is immediately obvious.
You might want to add some hints to the comments or adapt the argument name.

release/default.nix
```diff
@@ -99,34 +99,53 @@ let
...
-  includedPkgNames = [
-    "awscli"
-    "calibre"
-    "element-web"
-    "firefox"
-    "gitlab-ee"
-  ];
+  overlay = import ../pkgs/overlay.nix pkgs pkgs;
+  overlayPkgNames = getDottedPackageNames overlay [];
+  overlayPkgNamesToTest = lib.subtractLists excludedPkgNames overlayPkgNames;
```
osnyx, line 134: Has O(nm) complexity. Not super nice, but we also do not have that many elaborated data structures in Nix available, so probably fine, especially given that the excluded packages tend to be small.

release/default.nix
```diff
@@ -99,34 +99,53 @@ let
...
+      job = lib.hydraJob (lib.attrByPath path null pkgs);
+    in
+      map
+        (system: lib.setAttrByPath (path ++ [ system ]) job)
+        supportedSystems;
 
-  testPkgs = listToAttrs (map (n: { name = n; value = pkgs.${n}; }) testPkgNames);
+  pkgNameListToHydraJobs = pkgNameList:
+    # Merge the single-attribute sets from pkgNameToHydraJobs into one big attrset.
+    lib.foldl'
```
osnyx, line 151: strict version because it is still small enough to be evaluated on our small hydra machines?

> The difference is that evaluation is forced upon access. Usually used
with small whole results (in contrast with lazily-generated list or large
lists where only a part is consumed.)

osnyx (changes requested):

release/default.nix
dpausp: I can name it `visitedAttrPath` and add a comment. `path` is used in `attrsets.nix` but `attrPath` is also there for the same thing.

release/default.nix
dpausp: I have thought about this before and decided that it is ok for the small structures here. Also it's only in the release code path.

release/default.nix
dpausp: `foldl'` is typically used in similar functions in nixpkgs and it's the builtin function for that. I don't think it even makes a difference here because everything will be evaluated anyway. The strict variant is probably better for error messages.

osnyx (approved):

PR #713, opened 2023-05-25
Flake-Info patches

PhilTaken:
@flyingcircusio/release-managers

## Release process

The gist of this patch will have to be applied to all release branches in order to index them with https://github.com/flyingcircusio/fc-search

Impact:

Changelog:

## Security implications

- [ ] [Security requirements](https://wiki.flyingcircus.io/System_Development_Guideline#Security_requirement_principles_and_testing) defined? (WHERE)
- [ ] Security requirements tested? (EVIDENCE)

As part of the adoption of nixos-search for the custom flyingcircus nixos module several changes had to be made. Generally these were limited to very small changes or adding a default text for options to prevent it from evaling deeper-nested configs that rely on hardware-specific configuration and cannot be generalized very well.

I also took the liberty of adjusting the way the mail module imports nixos-mailserver to allow for a pure evaluation of the options-json generating package.

There is still a small bug on flyingcircus-managed vms that leads to a permission error:

```
phil@test21 ~/fc-nixos $ nix build
error: flake output attribute 'packages.x86_64-linux.default' is not a derivation

phil@test21 ~/fc-nixos $ nix repl
Welcome to Nix 2.11.1. Type :? for help.

nix-repl> :lf .
Added 15 variables.

nix-repl> :b packages.x86_64-linux.default
error: opening file '/etc/nixos/services.json': Permission denied

nix-repl>
```

However, when building on a machine without this file the error is not triggered:

```
fc-nixos on branch flake
λ ❱ nix build

fc-nixos on branch flake (took 2s)
λ ❱
```

I presume this is a nix bug when evaluating options, not sure how to work around it.

EDIT: this error can be resolved by using a newer version of nix on the VM:

```
phil@test21 ~ $ nix shell nixpkgs#nix
                                                                                                                                                                                                                                                             
phil@test21 ~ $ nix --version
nix (Nix) 2.13.3
                                                                                                                                                                                                                                                             
phil@test21 ~ $ nix build github:philtaken/fc-nixos/flake
                                                                                                                                                                                                                                                             
phil@test21 ~ $
```

PR #828, opened 2023-11-13
Patch for FC-Search

PhilTaken:
this is a revamped version of https://github.com/flyingcircusio/fc-nixos/pull/713 after it was reverted in 89ef0b400b35deb5626f06dc5dae48b78647f3c7.

There are no functionality changes, these changes just make it possible to evaluate the options and generate a json from them. 

## Release process

Impact:

1. several minor adjustments to modules options, none that change functionality anywhatsoever
2. migration from literalExample -> literalExpression for options

Changelog:

Fix module option evaluation to enable the planned fc-search option discoverability and documentation service https://github.com/flyingcircusio/fc-search

## Security implications

  - [x] no functionality changes to the platform itself
  - [x] this commit evaluates + builds on a VM

### DIFF
--- a/nixos/platform/users.nix
+++ b/nixos/platform/users.nix
@@ -131,6 +131,7 @@ in
 
     userDataPath = lib.mkOption {
       default = /etc/nixos/users.json;
+      defaultText = "/etc/nixos/users.json";
       type = path;
       description = "Where to find the user json file.";
     };
@@ -142,6 +143,7 @@ in
 
     permissionsPath = lib.mkOption {
       default = /etc/nixos/permissions.json;
+      defaultText = "/etc/nixos/permissions.json";
       type = path;
       description = ''
         Where to find the permissions json file.
--- a/nixos/roles/antivirus.nix
+++ b/nixos/roles/antivirus.nix
@@ -18,6 +18,7 @@ in
         type = lib.types.listOf lib.types.str;
         default = fclib.network.lo.dualstack.addresses ++
                   fclib.network.srv.dualstack.addresses;
+        defaultText = "addresses of the interfaces `lo` and `srv` (IPv4 & IPv6)";
       };
     };
   };
--- a/nixos/roles/ceph/mon.nix
+++ b/nixos/roles/ceph/mon.nix
@@ -19,6 +19,7 @@ in
       supportsContainers = fclib.mkDisableContainerSupport;
 
       primary = lib.mkOption {
+        defaultText = "false";
         default = (first_mon == config.networking.hostName);
         description = "Primary monitors take over additional maintenance tasks.";
         type = lib.types.bool;
@@ -26,6 +27,7 @@ in
 
       config = lib.mkOption {
         type = lib.types.lines;
+        defaultText = "";
         default = (lib.concatMapStringsSep "\n" (mon:
           let id = head (lib.splitString "." mon.address);
               addr = "${head (filter fclib.isIp4 mon.ips)}:${mon_port}";
--- a/nixos/roles/coturn.nix
+++ b/nixos/roles/coturn.nix
@@ -50,6 +50,19 @@ in
       config = mkOption {
         description = "Platform-configured options";
         type = types.attrs;
+        defaultText = {
+          hostname = "\${cfg.hostName}";
+          alt-listening-port = 3479;
+          alt-tls-listening-port = 5350;
+          listening-ips = "the addresses of networks `lo`, `srv` and `fe` (IPv4 & IPv6)";
+          listening-port = 3478;
+          lt-cred-mech = false;
+          no-cli = true;
+          realm = "\${cfg.hostName}";
+          tls-listening-port = 5349;
+          use-auth-secret = true;
+          extraConfig = [];
+        };
         default =  {
           hostname = cfg.hostName;
           alt-listening-port = 3479;
--- a/nixos/roles/memcached.nix
+++ b/nixos/roles/memcached.nix
@@ -36,6 +36,7 @@ in
         type = lib.types.listOf lib.types.str;
         default = fclib.network.lo.dualstack.addresses ++
                   fclib.network.srv.dualstack.addresses;
+        defaultText = "the addresses of the networks `lo` and `srv` (IPv4 & IPv6)";
       };
 
     };
--- a/nixos/roles/mysql.nix
+++ b/nixos/roles/mysql.nix
@@ -39,6 +39,7 @@ in
           type = lib.types.listOf lib.types.str;
           default = fclib.network.lo.dualstack.addresses ++
                     fclib.network.srv.dualstack.addresses;
+          defaultText = "the addresses of the networks `lo` and `srv` (IPv4 & IPv6)";
         };
 
         extraConfig = mkOption {
--- a/nixos/roles/webproxy.nix
+++ b/nixos/roles/webproxy.nix
@@ -28,6 +28,7 @@ in
 
       listenAddresses = lib.mkOption {
         type = lib.types.listOf lib.types.str;
+        defaultText = "the addresses of the networks `lo` and `srv` (IPv4 & IPv6)";
         default = fclib.network.srv.dualstack.addressesQuoted ++
                   fclib.network.lo.dualstack.addressesQuoted;
       };
--- a/nixos/services/ceph/client.nix
+++ b/nixos/services/ceph/client.nix
@@ -21,6 +21,30 @@ in
     flyingcircus.services.ceph = {
       config = lib.mkOption {
         type = lib.types.lines;
+        defaultText = ''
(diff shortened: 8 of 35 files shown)


flake.nix
```diff
@@ -0,0 +1,83 @@
...
+    in builtins.getFlake "github:${owner}/${repo}/${rev}";
+
+    inherit (nixpkgs) lib;
+    nixpkgsConfig = import ./nixpkgs-config.nix;
+
+    pkgsFor = system: import nixpkgs {
+      inherit system;
+      overlays = [ self.overlays.default ];
+      config = {
+        inherit (nixpkgsConfig) permittedInsecurePackages;
```
osnyx: Why not import more/ all config attributes? We additionaly have at least `allowedUnfreePackageNames` as well.

PhilTaken:
I just reworked the options generation by folding it into the release hydra jobs. Doing so eliminates most, if not all of the limitations from previous iterations that were imposed by the flake approach that would require a bigger revamp of most of the repository.

nixos/platform/users.nix
```diff
@@ -131,6 +131,7 @@ in
 
     userDataPath = lib.mkOption {
       default = /etc/nixos/users.json;
+      defaultText = "/etc/nixos/permissions.json";
```
osnyx:
```suggestion
      defaultText = "/etc/nixos/users.json";
```

nixos/roles/antivirus.nix
```diff
@@ -18,6 +18,7 @@ in
         type = lib.types.listOf lib.types.str;
         default = fclib.network.lo.dualstack.addresses ++
                   fclib.network.srv.dualstack.addresses;
+        defaultText = "addresses of the interfaces `lo` and `srv`";
```
osnyx:
```suggestion
        defaultText = "addresses of the interfaces `lo` and `srv` (both IPv4 & IPv6)";
```

nixos/roles/coturn.nix
```diff
@@ -50,6 +50,19 @@ in
       config = mkOption {
         description = "Platform-configured options";
         type = types.attrs;
+        defaultText = {
+          hostname = "\${cfg.hostName}";
+          alt-listening-port = 3479;
+          alt-tls-listening-port = 5350;
+          listening-ips = "the addresses of networks `lo`, `srv` and `fe`";
```
osnyx:
```suggestion
          listening-ips = "the addresses of networks `lo`, `srv` and `fe` (IPv4 & IPv6)";
```

nixos/roles/memcached.nix
```diff
@@ -36,6 +36,7 @@ in
         type = lib.types.listOf lib.types.str;
         default = fclib.network.lo.dualstack.addresses ++
                   fclib.network.srv.dualstack.addresses;
+        defaultText = "the addresses of the networks `lo` and `srv`";
```
osnyx:
```suggestion
        defaultText = "the addresses of the networks `lo` and `srv` (IPv4 & IPv6)";
```

nixos/roles/mysql.nix
```diff
@@ -39,6 +39,7 @@ in
           type = lib.types.listOf lib.types.str;
           default = fclib.network.lo.dualstack.addresses ++
                     fclib.network.srv.dualstack.addresses;
+          defaultText = "the addresses of the networks `lo` and `srv`";
```
osnyx:
```suggestion
          defaultText = "the addresses of the networks `lo` and `srv` (IPv4 & IPv6)";
```

nixos/roles/webproxy.nix
```diff
@@ -28,6 +28,7 @@ in
 
       listenAddresses = lib.mkOption {
         type = lib.types.listOf lib.types.str;
+        defaultText = "the addresses of the networks `srv` and `lo`";
```
osnyx:
```suggestion
        defaultText = "the addresses of the networks `lo` and `srv` (IPv4 & IPv6)";
```

nixos/services/ceph/client.nix
```diff
@@ -21,6 +21,30 @@ in
     flyingcircus.services.ceph = {
       config = lib.mkOption {
         type = lib.types.lines;
+        defaultText = ''
```
osnyx, line 24: Is it strictly necessary to provide a defaultText here? If not, I'd leave this one out here.

1. ceph is currently not used in our 23.11 release, the 21.05 one which is used has already diverged quite a bit.
2. The defaultText is not fully matching what's generated in the `default`, e.g `fsid` appears to be missin.
3. I see a maintenance burde of keeping the `defaultText` and the actual `default` in sync. If possible, I'd argue to generate the `defaultText` from `default` at eval time.

nixos/services/nginx/default.nix
```diff
@@ -163,6 +163,7 @@ in
     defaultListenAddresses = lib.mkOption {
       type = lib.types.listOf lib.types.str;
       default = fclib.network.fe.dualstack.addressesQuoted;
+      defaultText = "addresses of the `fe` network";
```
osnyx:
```suggestion
      defaultText = "addresses of the `fe` network (IPv4 & IPv6)";
```

nixos/services/prometheus.nix
```diff
@@ -74,7 +74,7 @@ let
 
   mkDefOpt = type : defaultStr : description : mkOpt type (description + ''
 
-    Defaults to <literal>${defaultStr}</literal> in prometheus
+    Defaults to <literal>${builtins.toString defaultStr}</literal> in prometheus
```
osnyx, line 77: This could also be converted to MySt markdown I guess.

nixos/services/redis.nix
```diff
@@ -27,6 +27,7 @@ in {
 
       listenAddresses = lib.mkOption {
         type = lib.types.listOf lib.types.str;
+        defaultText = "the addresses of the networks `lo` and `srv`";
```
osnyx:
```suggestion
        defaultText = "the addresses of the networks `lo` and `srv` (IPv4 & IPv6)";
```

release/default.nix
```diff
@@ -322,6 +326,30 @@ let
     in
     {
 
+      fc-options = let
```
osnyx: Possibly add a descriptive comment about the purpose of this image. According to line 392, this is more kind of a dummy/ proxy and not intended to output an image, but to grab the options from it.

osnyx (changes requested):
Mostly some cosmetic changes in the docstrings.

> The only major changes are in release/default.nix and release/generate-options.nix

There is no release/generate-options.nix in this diff though.

nixos/services/ceph/client.nix
PhilTaken: Some `defaultText` is necessary since the evaluation of the `default` value implies deep evaluation of values that are not set in a simple image that only serves to generate the documentation.

If the module is not used and there is a maintenance burden I would instead suggest we provide an empty or placeholder `defaultText`.

release/default.nix
PhilTaken: I decided to move the image and the generation of the options to the fc-search module to keep this part of the code cleaner and more flexible.

nixos/services/ceph/client.nix
osnyx: Yes, then let's move forward with such a placeholder.

release/default.nix
osnyx: I cannot find that module's code in this PR, I assume you're going to provide that separately at a later point?

nixos/roles/ceph/mon.nix
```diff
@@ -19,13 +19,15 @@ in
       supportsContainers = fclib.mkDisableContainerSupport;
 
       primary = lib.mkOption {
+        defaultText = "false";
```
osnyx, line 22: Technically not correct as it automatically evaluates to "true" on primary mons, but I am okay with leaving this as Ceph is unused in this release.

osnyx (approved):

release/default.nix
PhilTaken: yes, it lives in https://github.com/flyingcircusio/fc-search

since generating the options for each channel is just a matter of about 5 seconds, there is no need to do t on hydra at the cost of less maintainable code

nixos/roles/ceph/mon.nix
PhilTaken: in most cases it would be useful to add the default value behaviour to the description, especially when it depends on other, more intricate options

PhilTaken:
Would a backmerge onto 23.05-{prod, stag, dev}, 22.11-{prod, stag, dev} and maybe 22.05-{prod, stag, dev} be possible? A lot of VMs are still not on 23.11 yet.

osnyx:
I'd leave the decision to @dpausp.

PR #835, opened 2023-11-23
Split nginx logs per vhost

PhilTaken:

## Release process

Impact:

- no change to the default behavior
- add an option to enable per-vhost access and error logs 

Changelog:

- Added an option `flyingcircus.services.nginx.logPerVirtualHost` to enable per-vhost access and error logs in nginx under `/var/log/nginx/access-<vhost-name>.log` and `/var/log/nginx/error-<vhost-name>.log` respectively. This is the new default behavior.

## Security implications

  - [x]  must not introduce untested behavior
  - [x] the new behavior is tested

### DIFF
--- a/nixos/services/nginx/default.nix
+++ b/nixos/services/nginx/default.nix
@@ -57,8 +57,14 @@ let
 
   vhostsJSON = fclib.jsonFromDir localCfgDir;
 
-  mkVanillaVhostFromFCVhost = name: vhost:
-    (removeAttrs vhost [ "emailACME" "listenAddress" "listenAddress6" ]);
+  mkVanillaVhostFromFCVhost = name: vhost: let
+    servername = if vhost.serverName != null then vhost.serverName else name;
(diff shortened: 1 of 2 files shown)


PhilTaken:
I changed the default value of the new split-logfiles settings to "true".

Using variables to split up the logfiles at runtime is not feasible realistically, since the access_log will suffer a performance hit due to unbuffered writes and the error_log doesn't even allow for variables in the file path.

osnyx (commented):
Code LGTM.

But how does

> The default behavior to log all incoming requests in /var/log/nginx/access.log remains unchanged.
(changelog)

fit

> I changed the default value of the new split-logfiles settings to "true".

? Does the changelog need updating?

PhilTaken:
> Does the changelog need updating?

good catch, the default was changed after I wrote the changelog

osnyx (approved):

PR #845, opened 2023-12-05
[21.05] Host configuration for EVPN-VXLAN

sysvinit:
### Summary

This change refactors the platform network configuration code to support the use of EVPN-VXLAN. The VXLAN-based operating mode is enabled when a special underlay interface is found in the ENC interface configuration, which triggers the creation of special kernel devices (including the underlay loopback and the VTEP interfaces) and the configuration of FRR to advertise local VTEP endpoints and learn remote VTEP endpoints.

Note that as 21.05 does not contain an FRR package, the package (and NixOS module) from 23.05 is used instead.

### Commentary

- When the ENC data contains an interface on the new `ul` VLAN (the underlay segment over which the overlay is operated), a new `fclib.underlay` attribute is exposed, which contains underlay-related configuration parameters which don't fit cleanly into the existing network-related data structures -- in particular the AS number for BGP, the loopback VTEP address, and the (possibly multiple) ethernet interfaces connected to the underlay segment. VXLAN configuration is enabled when `fclib.underlay` is set to a non-`null` value.
- Unmanaged network interfaces (including the directory interface policies `unmanaged` and `null`) are now properly ignored by the platform code.
- Previously, the network configuration code only distinguished between "virtual" (bridge) and "physical" (ethernet) kernel network devices. This taxonomy has been extended, as there are now four types of kernel network device: ethernet interfaces, bridge interfaces, loopback interfaces (provided by the kernel "dummy" driver), and VXLAN interfaces. The systemd units which configure device link properties are now separated to reflect this distinction. Additionally, udev rules are only generated for ethernet kernel devices.
- Each VXLAN interface is unconditionally connected to a bridge interface, as this appears to be required by FRR to announce the associated VNI to other hosts, irrespective of whether one is actually otherwise necessary for the host configuration.

## Release process

Impact: internal.

Changelog:

## Security implications

  - The new underlay network is an internal and sensitive network segment, comparable with the storage or storage backend network, and should be treated and protected accordingly.
  - The BGP control plane is critical to the operation of hosts operating with VXLAN, and should be appropriately secured.
  - This PR has been manually tested and verified on a spare KVM host in DEV.
  - VM networking is achieved through the use of bridge interfaces, which connects them directly to layer 2 ethernet segments without traffic passing through the host's IP stack. The underlay network is never exposed directly to any VMs.
  - The BGP control plane operates exclusively over the underlay network. FRR is directly configured to only connect over the ethernet devices explicitly declared as underlay interfaces in the ENC data, and filtering is used to ensure that only routes for networks given in the ENC configuration are accepted.
  - Security considerations for storage and backup hosts are unchanged, as these do not directly run customer workloads.

### DIFF
--- a/nixos/lib/network.nix
+++ b/nixos/lib/network.nix
@@ -215,7 +215,7 @@ rec {
       lib.nameValuePair vlan (
       let
         priority = routingPriority vlan;
-        bridged = interface.bridged;
+        bridged = interface.bridged || interface.policy == "vxlan";
 
         mtu = if hasAttr vlan config.flyingcircus.static.mtus
               then config.flyingcircus.static.mtus.${vlan}
--- a/nixos/lib/network.nix
+++ b/nixos/lib/network.nix
@@ -5,6 +5,48 @@
 with builtins;
 let
   fclib = config.fclib;
+
+  encInterfaces = lib.attrByPath [ "parameters" "interfaces" ] {} config.flyingcircus.enc;
+
+  foldConds = value: conds:
+    let
+      f = base: elem: if elem.cond then throw elem.fail else base;
+    in lib.foldl f value conds;
+
+  interfaceData =
+    let
+      underlayInterfaces = lib.filterAttrs (name: value: name != "ul" && value.policy == "underlay") encInterfaces;
+      underlayCount = length (attrNames underlayInterfaces);
+      vxlanInterfaces = lib.filterAttrs (name: value: value.policy == "vxlan") encInterfaces;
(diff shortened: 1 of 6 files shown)


ctheune (approved):
Wow, this looks great.

Some of our code is getting quite complicated, but I'm happy moving this forward.

ctheune:
I'm converting this to a draft to avoid accidental merge. We'd like to merge this after the last 21.05 release which is currently in progress on dev and then let this simmer over the holidays in dev and staging.

PR #857, opened 2023-12-16
varnish: fix the multihost configuration

PhilTaken:
the multihost configuration was broken because although the correct configuration was applied initially, during a reload of the varnish service a blank configuration would be loaded because the reload script was not written with the legacy case in mind.

in order to fix this, and to simplify the varnish service logic, the legacy config from the "/etc/local/varnish" directory is now rendered as a virtualhost that accepts any incoming request

this, however, means that a case where there is a legacy config file as well as nix-defined varnish vhost configuration is harder to detect, which is why the corresponding assertion has been removed.

fixes #825 

## Release process

Impact:

Changelog:

## Security implications

  - [x] fixes previously broken behaviour 
  - [x] a new test case has been added to account for this

### DIFF
--- a/nixos/roles/webproxy.nix
+++ b/nixos/roles/webproxy.nix
@@ -10,7 +10,9 @@ let
   kill = "${pkgs.coreutils}/bin/kill";
 
   # if there is a default.vcl file, use that instead of the NixOS Varnish configuration
-  varnishCfg = fclib.configFromFile /etc/local/varnish/default.vcl null;
+  rawVarnishCfg = fclib.configFromFile /etc/local/varnish/default.vcl null;
+  # this is required for testing since the default.vcl file does not exist at build time
+  varnishCfg = if rawVarnishCfg == null && config.environment.etc ? "local/varnish/default.vcl" then config.environment.etc."local/varnish/default.vcl".text else null;
 in
 {
   options = with lib; {
@@ -38,7 +40,7 @@ in
       warnings = lib.optional (varnishCfg != null) "Configuring varnish via /etc/local/varnish/default.vcl is deprecated, please migrate your Configuration to Nix";
 
       assertions = [{
-        assertion = !((config.flyingcircus.services.varnish.virtualHosts != {}) && (varnishCfg != null));
+        assertion = builtins.length (builtins.attrNames config.flyingcircus.services.varnish.virtualHosts) <= 1 || builtins.isNull varnishCfg;
         message  = ''
           Please remove the file `/etc/local/varnish/default.vcl` if you want to specify your Varnish configuration in Nix code.
         '';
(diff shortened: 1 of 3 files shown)


nixos/roles/webproxy.nix
```diff
@@ -37,13 +39,6 @@ in
     (lib.mkIf fccfg.enable {
       warnings = lib.optional (varnishCfg != null) "Configuring varnish via /etc/local/varnish/default.vcl is deprecated, please migrate your Configuration to Nix";
 
-      assertions = [{
-        assertion = !((config.flyingcircus.services.varnish.virtualHosts != {}) && (varnishCfg != null));
```
ctheune, line 41: You could change that assertion to varnishCfg != and virtualHosts having more than 1 item.

nixos/roles/webproxy.nix
```diff
@@ -87,7 +82,12 @@ in
         extraCommandLine = "-s malloc,${toString cacheMemory}M";
         http_address = lib.concatMapStringsSep " -a "
           (addr: "${addr}:8008") fccfg.listenAddresses;
-        config = varnishCfg;
+        virtualHosts = lib.optionalAttrs (varnishCfg != null) {
+          "default-vcl" = {
```
ctheune, line 93: This is a bit unpredictable. Could we introduce an order here so that we'd either have this consistently be the first vhost or the last vhost?

ctheune (changes requested):

nixos/roles/webproxy.nix
PhilTaken: might try to make it the only vhost with lib.mkForce

nixos/roles/webproxy.nix
PhilTaken: the single item might be overriden forcefully, but at that point the assertion isn't effective anyways. will add

nixos/roles/webproxy.nix
PhilTaken: in fact, the assertion now takes care of that. adding a lib.mkForce here would lead to unintentional behaviour since it would just ignore the user-defined vhosts

ctheune (changes requested):
Looks good in general. As always and specifically because this is a regression, I need a test here. I tried it myself, but test coverage for assertions are hairy and none of the examples I found work here.

Maybe @dpausp or max or @leona-ya have an insight?

PhilTaken:
I tested the build failure manually by adding a vhost to a test with a `/etc/local/varnish/default.vcl` file to test that the build fails and it does. 

Testing this in CI would obviously be better, unfortunately the fc-manage script requires internet access which we do not provide in the VM tests. @ctheune possible follow-up on this: an offline switch feature for fc-manage?

ctheune (approved):

PR #870, opened 2024-01-02
Make Maximilian Bosch admin

Ma27:
Re FC-34774
Please backport as far as reasonable

## Release process

Impact:

Changelog:

* Introduce Maximilian Bosch as new global admin (FC-34774)

## Security implications

    - [x] strong protections for the private key implemented (Strong passphrase, separate from all other keys, not automatically activated on session logins)
    - [ ] reviewed setup


dpausp (approved):
LGTM. The corresponding issue says that there's management approval for this. I also checked that the root SSH public key is different from the one in the directory.

PR #875, opened 2024-01-05
[21.05] backup: fix race condition in restore-single-files.

ctheune:
losetup does trigger a partscan but returns before the udev rules have been processed. Add a polling loop to wait for the device to appear.

Fixes PL-132027

## Release process

Impact:

none

Changelog:

internal only

## Security implications

no specific requirements

functionally tested manually on backup server

### DIFF
--- a/pkgs/restore-single-files/restore-single-files.sh
+++ b/pkgs/restore-single-files/restore-single-files.sh
@@ -51,9 +51,13 @@ echo $LOOPDEV
 
 TERMINATE="umount $LOOPMNT; losetup -d $LOOPDEV; sleep 1; fusermount -u $FUSEMNT"
 trap "$TERMINATE" ERR 1 2 3 5 15
+LOOPPART="${LOOPDEV}p1"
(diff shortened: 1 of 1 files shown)


osnyx:
For completeness sake, I'd like to point to `udevadm settle` as a potential alternative to polling:

```
   udevadm settle [options]
       Watches the udev event queue, and exits if all current events are handled.

       -t, --timeout=SECONDS
           Maximum number of seconds to wait for the event queue to become empty. The default value is 120 seconds. A value of 0 will check if the queue is empty and always return immediately. A non-zero value will return an exit
           code of 0 if queue became empty before timeout was reached, non-zero otherwise.

       -E, --exit-if-exists=FILE
           Stop waiting if file exists.

       -h, --help
           Print a short help text and exit.

       See systemd-udev-settle.service(8) for more information.
```

But I also agree that for this simple case, polling is fine as well and probably less complex.

osnyx (approved):

PR #877, opened 2024-01-05
Add flyingcircus.passwordlessSudoPackages and use it

dpausp:
Add flyingcircus.passwordlessSudoPackages and use it

The new option built on top of `passwordlessSudoRules` is more
convenient, especially with recent sudo versions as it requires a rule
for the Nix store path and /run/current-system for typical use cases.

PL-132015

## Release process

Impact:

Changelog:

- (internal)

## Security implications

Older sudo versions resolved symlinks from /run/current-system to `/nix/store` so having rules for Nix store paths was enough.
This change is mostly a refactoring to reduce rule duplication. Using the new option in cases where the rule for the other path  didn't exist (either an oversight or the rule wasn't needed) adds some new generated rules but without giving additional sudo permissions. 

  - [x] this change should not introduce new possibilities to use sudo without password
  - looked at sudoers file generated on a test VM with many roles: no unexpected rules found
  - some NixOS tests check sudo rules

### DIFF
--- a/nixos/platform/default.nix
+++ b/nixos/platform/default.nix
@@ -188,10 +188,50 @@ in {
       };
     };
 
+    flyingcircus.passwordlessSudoPackages = mkOption {
+      description = ''
+        Works similar to security.sudo.extraRules, but `commands` are
+        interpreted as paths relative to a given `package` so binary can be
+        run from the full Nix store path with `sudo`.
+
+        This option also creates a corresponding rule for
+        `/run/current-system/sw/` to be able to run the globally installed
+        binary from $PATH with `sudo`.
+
+        You still need to add the package to `environment.systemPackages`
+        yourself to make it globally available.
+
+        Like `flyingcircus.passwordlessSudoRules`, this automatically sets
+        passwordless mode and places rules after rules with default order
+        number(uses `mkOrder 1100`)
+      '';
+
+      example = literalExpression ''
+        [
+          {
+            commands = [ "bin/a" "bin/b --option" ];
+            package = pkgs.example;
+            groups = [ "admins" ];
+          }
+        ]
+      '';
+
+      default = [];
+    };
+
     flyingcircus.passwordlessSudoRules = mkOption {
       description = ''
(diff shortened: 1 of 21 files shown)


nixos/platform/default.nix
```diff
@@ -188,12 +188,52 @@ in {
       };
     };
 
+    flyingcircus.passwordlessSudoPackages = mkOption {
+      description = ''
+        Works similar to security.sudo.extraRules, but `commands` are
+        interpreted as paths relative to a given `package` so binary can be
```
osnyx:
```suggestion
        interpreted as paths relative to a given `package` so the binary can be
```

nixos/platform/default.nix
```diff
@@ -188,12 +188,52 @@ in {
...
+        run from the full Nix store path with `sudo`.
+
+        This option also creates a corresponding rule for
+        `/run/current-system/sw/` to be able to run the globally installed
+        binary from $PATH with `sudo`.
+
+        You still need to add the package to `environment.systemPackages`
+        yourself to make it globally available.
```
osnyx, line 202: This approach has 2 pitfalls:

- user error: a user might either forget to add the package there, or add a different package (derivation) of the same tool to `environment.systemPackage` than used in the sudo rule. This can yield unexpected defaults.  
  To protect against this, we could check for the existence of each `package` in the list `environment.systemPackages` and warn/ error if it is not found.
- malicious user: Put a different package into `environment.systemPackages` than intended/ expected by the sudo rule. This allows a malicious user to e.g. run a shell script as a different user then.  
  The previously mentioned existence check is no good countermeasure, because users can add additional packages with a conflicting path but higher `meta.priority` instead to get them included in `/run/current-system/sw`. Additionally, tight now crafting such a config is only possible for users that can modify system config anyways, giving them acces to other code execution means anyways.

----

Given that additional checks are mainly a convenience feature but no good security feature, I am okay with not doing them and keeping it this way.

osnyx (approved):
One typo to fix, and some thoughts about plausibility checks; but overall LGTM and can be merged.

PR #878, opened 2024-01-05
Nix Flake for development and release management

dpausp:
Nix Flake for dev and release tooling

The change turns the repo into a Nix flake (flake.nix + flake.lock).
Dependencies are now defined as flake inputs.

This doesn't influence runtime behaviour of the platform and we still
use versions.json/versions.nix for dependencies but versions.json is now
generated by the flake with the same content as before.

To modularize and provide convenient abstractions for the flake file,
especially when dealing with support for multiple systems/architectures
we use [flake-parts](https://flake.parts/).

[devenv](devenv.sh) is used to define the development shell which
replaces shell.nix and provides unified access to various helper
scripts. The shell works for Macs (release tooling is usually run from
there) and on Linux (typically developer VMs).

As the top-level dir became too crowded, many scripts and files
from the top-level which are related to dependency management and the
release process moved to the `release` subdir. Their location isn't
important anymore as they are meant to be called through scripts
generated by devenv now. `./update-nixpkgs.py` is now invoked as
`update_nixpkgs`, for example.  See the comment at the top of
`flake.nix` for more info.

./dev-setup still exists as script for our "muscle memory" but is only a
facade now, just calling `dev_setup` running in a dev shell.

During the the top-level directory cleanup, fc/ceph/__init__.py which
serves no purpose was removed, too. This causes isort (for Python),
which is part of the pre-commit hook, to sort imports differrently
as fc.ceph is now recognized as external dependency. This explains
the changes in pkgs/fc/ceph/fc/ceph.

PL-131814

## Release process

Impact:

Changelog:

- (internal)

## Security implications

  - only dev/release tooling, must not have influence on build or runtime behaviour of the platform itself
  - tried out dev shell and all commands on a Mac for release management (updating nixpkgs, performing releases, getting channel URL, showing branch diffs). Generated versions.json is compatible to the old way of updating it, no semantic changes. Updating nix-phps is partially broken but this has nothing to do with this change.
  - tried out dev shell and all commands on a dev VM for platform development (building packages, creating channels dir, building a system from it)


flake.nix
```diff
@@ -0,0 +1,157 @@
...
+      inputs.nixpkgs-22_11.follows = "nixpkgs";
+      inputs.nixpkgs-23_05.follows = "nixpkgs";
+    };
+    devenv = {
+      url = "github:cachix/devenv";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = inputs@{ flake-parts, ... }:
```
osnyx, line 40: By not specifying `flake-parts` in inputs, we rely on the system's flake registry for resolving this.

IMHO doing this is kind of an antipattern, but this might be personal taste.

flake.nix
```diff
@@ -0,0 +1,169 @@
+{
+  description = "Flying Circus NixOS platform";
+
+  inputs = {
+    nixpkgs.url = "github:flyingcircusio/nixpkgs/nixos-23.11";
+    nixos-mailserver = {
+      url = "gitlab:flyingcircus/nixos-mailserver?host=gitlab.flyingcircus.io";
+      inputs.nixpkgs.follows = "nixpkgs";
+      inputs.nixpkgs-22_11.follows = "nixpkgs";
```
osnyx, line 30: At a quick look, this input is not used anywhere in nixos-mailserver, so we should remove it altogether over there.
The 23_05 input should probably be renamed to avoid confusion.

flake.nix
```diff
@@ -0,0 +1,169 @@
...
+      systems = [ "x86_64-linux" "x86_64-darwin" "aarch64-darwin" ];
+      perSystem = { config, self', inputs', pkgs, lib, system, ... }: {
+        # Per-system attributes can be defined here. The self' and inputs'
+        # module parameters provide easy access to attributes of the same
+        # system.
+
+        _module.args.pkgs = import inputs.nixpkgs {
+          inherit system;
+          overlays = [ (import ./pkgs/overlay.nix) ];
+          config = {};
```
osnyx: To which modules is this provided as args, is this only on flake-parts level or also on nixos-configurations level?
Because then this pkgs instance might be configured with a different nixpkgs config than the system packages of a NixOS installation.

pkgs/fc/ceph/fc/ceph/maintenance.py
```diff
@@ -4,9 +4,8 @@
 import subprocess
 import sys
 
-import fc.util.directory
-
 import fc.ceph.images
+import fc.util.directory
```
osnyx, line 8: This is unrelated. I guess it's caused by some pre-commit upgrade?

versions.nix
osnyx, line 1: The GitHub diff view for this being a symlink is pretty confusing.

osnyx (commented):
Mostly okay, with a few questions and concerns.

pkgs/fc/ceph/fc/ceph/maintenance.py
dpausp: It's ok, this is caused by the removal of `fc/ceph/__init__.py` (which is empty). isort now sees fc.ceph as external package and sorts it differently.

flake.nix
dpausp: The flake only affects release/dev tooling, there's no influence on the platform itself. I still added the nixpkgs config from default.nix to the flake for completeness.

flake.nix
dpausp: We could fix this in nixos-mailserver, but we'll keep the overrides for now to avoid pulling in additional nixpkgs we don't need.

osnyx (approved):

PR #881, opened 2024-01-11
rabbitmq 3.6.5 role for upgrades from NixOS 20.09

dpausp:
This is a new type of role: it only exists for upgrading existing VMs directly from NixOS 20.09. The rabbitmq service from the old version is "frozen" by generating local config that just takes the existing unit file and its dependencies and adds it as a systemd package. This way, the service doesn't even restart when the system is upgraded to 23.11 as nothing changed for the service.

The added platform code is mostly just lifted from the 20.09 rabbitmq36_5 role and the corresponding service.

PL-131333

## Release process

Impact:

Changelog:

- (no need to announce)

## Security implications

  - For rabbitmq itself, nothing should change. We replicate the behaviour of the role on NixOS 20.09. This change is about being able to upgrade VMs that still run legacy RabbitMQ versions. This is obviously a compromise regarding security but it's better than keeping old platform versions running. 
  - checked on a test VM that upgrading from 20.09 to 23.11 works without changing the rabbitmq service, rabbitmq itself listens on SRV 5672, management on 15672 and 25672, all interfaces (but protected by firewall and management requires login), basic functionality test of sending and receiving a message via Python `pika`.
  -  checked on another test VM that current rabbitmq role still runs as expected
  - automatic test covers current rabbitmq

### DIFF
--- a/doc/src/rabbitmq.md
+++ b/doc/src/rabbitmq.md
@@ -34,4 +34,41 @@ After upgrading a cluster, enable all feature flags:
 sudo -u rabbitmq rabbitmqctl enable_feature_flag all
 ```
 
+## Upgrading from NixOS 20.09 while keeping RabbitMQ 3.6.5
+
+To be able to upgrade NixOS 20.09 machines using the `rabbitmq36_5` role,
+we provide a way to keep the unchanged rabbitmq service running after the system
+upgrade. This conserves the specific rabbitmq config for the machine and
+cannot be used on new 23.11 machines.
+
+The upgrade process starts with generating Nix config on the running machine.
+Put the generated config in :file:`/etc/local/nixos/rabbitmq365-frozen.nix`.
+
+```shell
+#!/usr/bin/env sh
+service=$(realpath /etc/systemd/system/rabbitmq.service)
+storePath=${service%%/rabbitmq.service}
+cat <<EOF
+# Generated config to freeze the existing rabbitmq unit file and all dependencies
+# to keep it running after an upgrade to 23.11.
+{ lib, ... }:
+{
+  # This no-op option declaration is needed for building the 20.09 system.
+  options.flyingcircus.services.rabbitmq365Frozen.service = lib.mkOption {};
+
+  # On 20.09, this setting changes nothing. It's only effective on 23.11.
+  config.flyingcircus.services.rabbitmq365Frozen.service =
+    builtins.storePath $storePath;
+}
+EOF
+```
+
+After that, change the VM environment to a 23.11 edition, keep the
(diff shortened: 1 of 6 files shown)


nixos/roles/rabbitmq.nix
```diff
@@ -10,50 +10,51 @@ in
...
-    };
-
-  in lib.mkMerge [
+    (lib.mkIf (config.flyingcircus.roles.rabbitmq36_5.enable) {
+      flyingcircus.services.rabbitmq365Frozen.enable = true;
+    })
 
-    (lib.mkIf (enabled) {
-      flyingcircus.services.rabbitmq = serviceConfig;
+    (lib.mkIf (config.flyingcircus.roles.rabbitmq.enable) {
```
dpausp, line 50: move to service

nixos/roles/rabbitmq.nix
```diff
@@ -10,50 +10,51 @@ in
     mkRole = v: {
       enable = lib.mkEnableOption
         "Enable the Flying Circus RabbitMQ ${v} server role.";
-      supportsContainers = fclib.mkEnableContainerSupport;
     };
   in {
     flyingcircus.roles = {
-      rabbitmq = mkRole "3.8";
+      rabbitmq36_5 = {
```
dpausp, line 16: should we add the frozen suffix here or keep it as the old role?

doc/src/rabbitmq.md
```diff
@@ -34,4 +34,40 @@ After upgrading a cluster, enable all feature flags:
...
+service=$(realpath /etc/systemd/system/rabbitmq.service)
+storePath=${service%%/rabbitmq.service}
+
+cat <<EOF
+# Generated config to freeze the existing rabbitmq unit file and all dependencies
+# to keep it running after an upgrade to 23.11.
+{ lib, ... }:
+{
+  # This no-op option declaration is needed for building the 20.09 system.
+  options.flyingcircus.services.rabbitmq365Frozen.service = lib.mkOption {};
```
dpausp, line 57: use role instead of the service

osnyx (approved):
Once the already mentioned changes are integrated and the security considerations have been filled out, this LGTM.

nixos/roles/rabbitmq.nix
dpausp: no, must be the same as the old role to allow upgrades without changing the role in the directory.

doc/src/rabbitmq.md
dpausp: The option is actually only declared for the service which makes sense. So keep it as is.

PR #892, opened 2024-01-22
Fixup for old mongodb roles PR

dpausp:
- Fix telegraf for 3.2 and 3.4
- allow upgrades without switching off the role
   builtins.fetchClosure may not be available when building the system
   which is fixed by falling back to storePath in this case.

PL-131229

## Release process

Impact: 

Changelog: 

## Security implications

  - nothing new
  - checked on a test VM that telegraf works and getting the mongodb package works

### DIFF
--- a/nixos/roles/mongodb.nix
+++ b/nixos/roles/mongodb.nix
@@ -7,6 +7,21 @@ let
   mcfg = config.services.mongodb;
   fclib = config.fclib;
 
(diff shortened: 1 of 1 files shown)


osnyx (approved):
One potential security consideration: When pulling in more and more pinned old store paths and their dependencies, we need to raise our awareness that even seemingly-current releases can pull in outdated software.

PR #905, opened 2024-02-07
[PL-132155] roles/mailserver: require an actual url as imprintUrl

Ma27:
Even though the option was called `imprintUrl` the protocol scheme was prepended unconditionally. This means when defining an `imprintUrl` like

    https://example.org/impressum

a redirect to `https://https//example.org/impressum` will be issued which is pretty unintuitive.

Since the option claims to accept a URL I decided to change its semantics like this:

* A URL is expected now (as long as it's accepted by the `return` clause of nginx).

* For backwards-compatibility, a "URL" w/o a scheme (i.e. `example.com`) is also accepted. This behavior is deprecated however and gives the following warning at evaluation time:

      trace: warning: Specifying flyingcircus.roles.mailserver.imprintUrl without a protocol scheme is deprecated. Add https:// to the declaration (i.e. "example.com" -> "https://example.com") to get rid of this warning.

  We could drop the old behavior after e.g. the first platform release with NixOS 24.11. When doing that, the `apply` expression should be dropped and the validation should be done by using `nullOr (strMatching "")` instead.

A test-case to the integration test of the mailserver role was added which ensures that the redirect declared by `imprintUrl` works fine.

## Release process

Impact:

Changelog:
* mailserver: `imprintUrl` now accepts a protocol scheme. Specifying this option without a protocol scheme still works as before, but is deprecated and will raise a warning (PL-132155)

- [x] PR has internal ticket

## Security implications

    - Reduce the chance of human error: a misconfigured redirect to the imprint is accepted by the module and may be unnoticed. However, the lack of an imprint may violate the ToS of other mail providers which potentially causes the mailserver to be banned by other providers.
    - Additional testcase in the integration test ensures that an imprintUrl with a protocol scheme gets properly redirected.
    - Manually verified that an imprintUrl without a protocol scheme gets properly redirected and gives a deprecation warning.

### DIFF
--- a/tests/mail/default.nix
+++ b/tests/mail/default.nix
@@ -25,7 +25,7 @@ in
   name = "mailserver";
   nodes = {
     mail =
-      { lib, ... }: {
+      { config, lib, ... }: {
         imports = [ ../../nixos ../../nixos/roles ];
         config = lib.mkMerge [
           commonConfig
@@ -36,6 +36,7 @@ in
               enable = true;
               mailHost = "mail.example.local";
               webmailHost = "webmail.example.local";
+              imprintUrl = "https://example.com";
               domains = {
                 "example.local" = {
                   primary = true;
@@ -107,6 +108,10 @@ in
             # LEC is not able to call remote services
             services.nginx.virtualHosts."autoconfig.example.local" = {
               addSSL = lib.mkForce false;
+              enableACME = lib.mkForce false;
+            };
+            services.nginx.virtualHosts.${config.flyingcircus.roles.mailserver.mailHost} = {
+              forceSSL = lib.mkForce false;
               enableACME = lib.mkForce false;
             };
 
@@ -181,7 +186,7 @@ in
         ];
     };
   };
-  testScript = let
+  testScript = { nodes, ... }: let
     passwdFile = "/var/lib/dovecot/passwd";
     chpasswdNixPath = "${pkgs.fc.roundcube-chpasswd-py}/bin/roundcube-chpasswd";
     globalChpasswd = "/run/current-system/sw/bin/roundcube-chpasswd";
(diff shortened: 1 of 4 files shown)


tests/mail/default.nix
```diff
@@ -192,6 +197,24 @@ in
     with subtest("postsuper sudo rule should be present for sudo-srv group"):
       mail.succeed('grep %sudo-srv /etc/sudoers | grep -q postsuper')
 
+    with subtest("imprint redirects to example.com"):
+      client.wait_for_unit("network-online.target")
+
+      mail.wait_for_unit("nginx")
+      mail.wait_for_open_port(80, addr='${nodes.mail.networking.primaryIPAddress}')
+
+      imprint_request_headers = client.succeed("curl mail.example.local -I | xargs -I {} echo {}").split('\n')
```
osnyx:
```suggestion
      imprint_request_headers = client.succeed("curl http://mail.example.local -I | xargs -I {} echo {}").split('\n')
```

Talking about protocol specifications, this makes it a bit more clear, also in case curl defaults ever change.

osnyx: I am also not fully sure why the xargs construction is necessary, doesn't `curl -I` already print headers line by line to stdout by itself?

osnyx (changes requested):
Looks good overall, just a small nitpick.

tests/mail/default.nix
Ma27: > I am also not fully sure why the xargs construction is necessary, doesn't curl -I already print headers line by line to stdout by itself?

curl doesn't print terminal escape codes when piping it to another command and that way I don't have to bother about dealing with them. If you know a better way (didn't find a useful option in the man page and NO_COLOR=1 doesn't work either), let me know :)

osnyx: As I assume you've done some research and I don't know one either from the top of my head, I'm fine with this hack.

osnyx (approved):

tests/mail/default.nix
osnyx: I guess piping to `cat` would've worked as well though.

PR #908, opened 2024-02-08
sensu: fix check_ping

dpausp:
This worked until upstream changed the hard-coded paths for ping from a /run/wrappers path (doesn't exist on 23.11 anymore) to just 'ping', relying on PATH. For check_ping, this doesn't work as PATH is ignored. Instead, we can hard-code the full path to ping during build time.

In addition to that, this change includes a comment about additional checks added by us which we don't use in platform code. We can probably remove them in 24.05.

postInstall is also overridden in a safe way now, appending to it if it already exists upstream.

PL-132185

## Release process

Impact:

Changelog:

* sensu/monitoring-plugins: fix `check_ping` (PL-132185).

- [x] PR has internal ticket

## Security implications

  - keep our monitoring working (restores functionality we had on 23.05)
   - tested on affected dev machine that the problem is fixed and ping statistics are reported
   - automated test now checks if check_ping works

### DIFF
--- a/pkgs/overlay.nix
+++ b/pkgs/overlay.nix
@@ -203,10 +203,44 @@ in {
 
   mysql = super.mariadb;
 
-  monitoring-plugins = super.monitoring-plugins.overrideAttrs(_: rec {
-    postInstall = ''
-    cp plugins-root/check_dhcp $out/bin
-    cp plugins-root/check_icmp $out/bin
+  monitoring-plugins = let
+    binPath = lib.makeBinPath (with self; [
+      (placeholder "out")
+      "/run/wrappers"
+      coreutils
+      gnugrep
+      gnused
+      lm_sensors
+      net-snmp
+      procps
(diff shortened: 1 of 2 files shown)


pkgs/overlay.nix
```diff
@@ -203,10 +203,44 @@ in {
...
+    # program and not a script like other checks), so check_ping needs to
+    # be compiled with the full path.
+    postPatch = ''
+      substituteInPlace po/Makefile.in.in \
+        --replace /bin/sh ${self.runtimeShell}
+
+      sed -i configure.ac \
+        -e 's|^DEFAULT_PATH=.*|DEFAULT_PATH=\"${binPath}\"|'
+
+      configureFlagsArray+=(
```
osnyx, line 232: Could be done as a Nix attr as well, but okay for me.

osnyx (approved):

PR #918, opened 2024-02-15
roles.gitlab: add HSTS headers

osnyx:
Add HTTP strict-transport-security headers to nginx config. Enabled by default for default VHost configs, defaults to 60 minutes expiration for now. Configurable via new `flyingcircus.roles.gitlab.hsts` options. Warning: Also applies to subdomains.

Includes a NixOS smoke test, also manually checked in a test deployment with `curl --hsts`.

PL-132164

## Release process

Impact:
- [gitlab] added default: Gitlab roles in default VHost configuration now ship HTTP Strict-Transport-Security headers
  - Once a browser has visited gitlab once, this information persists locally for 1 hour
  - during this time, gitlab and all of its subdomains can only be visited via HTTPS

Changelog:
- gitlab: introduce `flyingcircus.roles.gitlab.hsts`
  - enabled by default for all gitlab installations with `isDefaultVhost`
  - enables HTTP Strict-Transport-Security headers

- [x] PR has internal ticket

## Security implications

  - security by default: improve HTTPS effectivity by shipping HSTS headers, enable by default
  - no known regressions 
  - include basic smoke test for that change
  - [x] manually tested on staging gitlab, checked header parsing with `curl --hsts`

### DIFF
--- a/nixos/roles/gitlab.nix
+++ b/nixos/roles/gitlab.nix
@@ -53,6 +53,35 @@ in
         '';
       };
 
+      hsts = {
+        enable = lib.mkOption {
+          # GitLab not being the default vhost is a heuristic for uncommon setups
+          default = cfg.isDefaultVhost;
+          type = types.bool;
+          description = ''
+            Enable HTTP Strict-Transport-Security.
+            This causes browsers to only connect via HTTPS on subsequent visits.
+            Applies to all subdomains as well.
+            When disabled, the headers are still being sent with a max-age of 0,
+            effectively clearing up previously set HSTS times in browsers.
+          '';
+        };
+        maxAge = lib.mkOption {
+          type = types.ints.positive;
+          description = "Expiration time of HSTS caching in browser (unit: seconds)";
+          default = 3600;
+          defaultText = ''
+            3600 seconds (1 hour)
+            Note: We intend to massively increase that value in a following release.
+          '';
+        };
+        preload = lib.mkOption {
+          type = types.bool;
+          description = "Allow including this into the HSTS preloading list shipped with browsers.";
+          default = false;
+        };
+      };
+
       generateSecrets = mkOption {
         type = types.bool;
         description = ''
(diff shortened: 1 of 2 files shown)


nixos/roles/gitlab.nix
```diff
@@ -53,6 +53,30 @@ in
...
+          description = ''
+            Enable HTTP Strict-Transport-Security headers.
+            This causes browsers to only connect via HTTPS on subsequent visits.
+            Applies to all subdomains as well.
+          '';
+        };
+        maxAge = lib.mkOption {
+          type = types.ints.positive;
+          description = "Expiration time of HSTS caching in browser (unit: seconds)";
+          default = 63072000;
```
dpausp: I think it's better to start with a very low default (an hour?) to check if everything works correctly. Then we can increase the default or change it per VM.

nixos/roles/gitlab.nix
```diff
@@ -300,5 +335,11 @@ in
     };
   })
 
+  (lib.mkIf (cfg.enable && cfg.hsts.enable) {
+    assertions = [{
```
dpausp, line 342: maybe just a warning when we decrease the default to a short period (as mentioned above)?

dpausp (changes requested):

osnyx:
I had initially consulted https://infosec.mozilla.org/guidelines/web_security.html#http-strict-transport-security for information on HSTS and implicitly assumed there was a minimum required max-age. But after directly consulting [the RFC6797](https://www.rfc-editor.org/rfc/rfc6797#page-16) on this, this assumption turned out to be wrong. So let's start with a short max-age for now.

Additionally, I found

>    Additionally, server implementers should consider employing a default
>    max-age value of zero in their deployment configuration systems.
>    This will require deployers to willfully set max-age in order to have
>    UAs enforce the HSTS Policy for their host and will protect them from
>    inadvertently enabling HSTS with some arbitrary non-zero duration.

which motivated me to always ship HSTS headers, but with a max-age of 0 when being disabled.

dpausp (approved):

PR #928, opened 2024-02-26
Ensure docker firewall rules are deleted when role is removed.

sysvinit:
The docker roles adds a number of iptables chains in order to link the platform-provided firewall with the rules managed by the docker daemon. However, when the docker role is removed, these extra chains are not properly removed, as the commands for removing them are also removed from the firewall management scripts. As the additional docker-related chains contain references to other NixOS-managed chains, reloading the firewall after removing the docker role will fail, as the references from the docker chains will prevent the NixOS chains from being deleted and re-created properly.

This change introduces extra commands to the firewall management script when the docker role is *not* enabled, in order to ensure that the docker-related firewall chains are not present which might interfere with the rest of the firewall scripts.

PL-131699

## Release process

Impact:

Changelog: The firewall should now continue to function properly after the Docker role is removed (PL-131699).

- [x] PR has internal ticket

## Security implications

  - The platform-provided firewall should continue to function correctly when host configuration and the set of enabled roles is changed.
  - Manually verified on a test VM. When disabling the Docker role with this change applied, the firewall is reloaded and the system rebuild process exit without errors.

### DIFF
--- a/nixos/roles/docker.nix
+++ b/nixos/roles/docker.nix
@@ -2,6 +2,7 @@
 
 let
   fclib = config.fclib;
+  cfg = config.flyingcircus.roles.docker;
 in
 {
   options = {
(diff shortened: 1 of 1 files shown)


nixos/roles/docker.nix
```diff
@@ -11,40 +12,53 @@ in
...
+        ip46tables -F fc-docker 2>/dev/null || true
+        ip46tables -X fc-docker 2>/dev/null || true
+        # End FC docker rules (1100)
+      '';
+    })
+
+    (lib.mkIf (!cfg.enable) {
+      networking.firewall.extraStopCommands = lib.mkOrder 1100 ''
+        # FC docker rules (1100) -- ensure absence of docker firewall configuration
+        iptables -D FORWARD -j DOCKER-USER 2>/dev/null || true
```
dpausp: I think we should not remove DOCKER-USER which is created by docker. Users might use docker without our role and add custom config in this chain. Can we just remove the rules we add, as above in `extraStopCommands` for the other case?

dpausp (changes requested):

nixos/roles/docker.nix
ctheune: Hmm? Why are our scripts creating DOCKER-USER in the first place, then?

sysvinit: > Can we just remove the rules we add, as above in extraStopCommands for the other case?

Fixed in [04f2249](https://github.com/flyingcircusio/fc-nixos/pull/928/commits/04f224937c854f7baad86a305773c1d37689edaa)?

> Hmm? Why are our scripts creating DOCKER-USER in the first place, then?

Normally the DOCKER-USER chain is created by the docker daemon when it configures its iptables rules, however the docker daemon is only started *after* the NixOS firewall is started. The fc-docker chain which controls access between containers and other hosts in the resource group needs to be called from the DOCKER-USER chain, so both need to be created in the NixOS firewall scripts.

dpausp (approved):

PR #932, opened 2024-03-06
agent check free disk before upgrading, sensu check

dpausp:
request_update now guesses if the system to be prepared will fit on
the Nix store partition. It uses the size of the current system plus
some extra space as worst-case estimation. This should leave at least
some free space after preparing the new system.

Often nixpkgs updates change core dependencies, so adding the size of the
current system when preparing a new one is not uncommon.

This worst-case estimation will also prevent smaller updates from being installed, even though they'd perfectly fit into available disk space. This creates a new challenge for ensuring that a machine does not miss any updates, we have to make sure to watch the fc-agent check
which will become red when there's not enough space.

Additions:

- Successful system builds now show the closure size of the new system.
- `fc-manage check` now:
  - displays current system closure size as info.
  - warns when less than twice the system size plus some extra space is free.
  - errors when less than the system size plus some extra space is free.
  -  now shows system and state version

## Release process

Impact:

Changelog:

- agent: check free disk space before preparing a system update. At least the size of the current system plus a fixed value must be available. A Sensu check warns before the limit is reached and becomes critical if updates cannot run anymore. The default fixed value of 5 GiB can be configured using the `flyingcircus.agent.diskKeepFree` option (PL-131859).

- [x] PR has internal ticket

## Design notes

- [x] Provide a feature toggle if the change might need to be adjusted/reverted quickly depending on context. Consider whether the default should be `on` or `off`. Example: rate limiting. => min free space (5 GiB) can be configured and set to 0 (but the size of the current system is still considered). If all fails, the system can still be updated manually using fc-manage. 
- [ ] All customer-facing features and (NixOS) options need to be discoverable from documentation. Add or update relevant documentation such that hosted and guided customers can understand it as well. => document the new option before the release

## Security implications

  - Don't fill disks to 100% to prevent downtime and maybe even data corruption
  - Delaying updates for too long is a security risk. The low disk situation must be monitored so that we can identify machines that cannot run updates. 
  - checked on a test VM that
    - an update is prepare when enough space is available (with default diskKeepFree) 
    - preparing the update is rejected when not enough space available (diskKeepFree = 15, system size 2.6, free space 17.0)
    - sensu check is critical when updates are suspended
    - sensu check warns when disk space is getting low

### DIFF
--- a/pkgs/fc/agent/fc/maintenance/maintenance.py
+++ b/pkgs/fc/agent/fc/maintenance/maintenance.py
@@ -8,6 +8,7 @@ TODO: better logging for created activites.
 import os.path
 import shutil
 import typing
+from pathlib import Path
 from typing import Iterable, Optional
 
 import fc.util
@@ -20,6 +21,7 @@ from fc.maintenance.activity.reboot import RebootActivity
 from fc.maintenance.activity.update import UpdateActivity
 from fc.maintenance.activity.vm_change import VMChangeActivity
 from fc.maintenance.state import State
+from fc.util import nixos
 
 
 def request_reboot_for_memory(log, enc) -> Optional[Request]:
@@ -147,7 +149,7 @@ def request_reboot_for_kernel(
 
 
 def request_update(
-    log, enc, current_requests: Iterable[Request]
+    log, enc, config, current_requests: Iterable[Request]
 ) -> Optional[Request]:
     """Schedule a system update if the channel has changed and the
     resulting system is different from the running system.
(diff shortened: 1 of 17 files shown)


pkgs/fc/agent/fc/maintenance/maintenance.py
```diff
@@ -196,6 +198,24 @@ def request_update(
...
 
+    free_disk_gib = nixos.get_free_store_disk_space(log) / 1024**3
+    current_size_gib = (
+        nixos.system_closure_size(log, Path("/run/current-system")) / 1024**3
+    )
+    disk_keep_free = config.get("limits", "disk_keep_free", fallback=5)
+    free_disk_thresh = current_size_gib + disk_keep_free
+
+    if free_disk_gib < free_disk_thresh:
+        log.warning(
```
dpausp: error

pkgs/fc/agent/fc/util/nixos.py
```diff
@@ -358,6 +358,32 @@ def find_nix_build_error(stderr: str, log=_log):
...
+            args,
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+        size = result.stdout.split()[1]
+
+    except subprocess.CalledProcessError:
+        log.error("nix-path-info-failed", args=args, exc_info=True)
+        return
```
dpausp: raise a custom Exception instead?

pkgs/fc/agent/fc/maintenance/tests/test_maintenance.py
```diff
@@ -203,10 +210,15 @@ def test_request_update_unchanged_system_and_no_other_requests_skips_request(
         "fc.maintenance.maintenance.UpdateActivity.from_enc", from_enc_mock
     )
 
+    monkeypatch.setattr(
+        "fc.util.nixos.get_free_store_disk_space", lambda *a: 10 * 1024**3
```
osnyx, line 214: This monkeypatching *could* be put into a separate fixture that is just pulled in by all other tests and fixtures that run in an environment with enough free space. That reduces unnecessary code verbosity.

osnyx (approved):
Given the 2 changes you still intend to make, this looks good to me.

pkgs/fc/agent/fc/maintenance/tests/test_maintenance.py
dpausp: I thought about this but decided to keep it. Tests can be a more explicit and copying around two one-liners is ok, I think.

osnyx (approved):

PR #940, opened 2024-03-12
Reimplement HTTP service check program in Python for better redirect handling

sysvinit:
The `check_http` program in monitoring-plugins which is used for executing customer-provided HTTP service checks does not provide a way to check for specific status codes after following redirects. The only available behaviour is: (a) follow redirects, but then evaluate the resulting status code using the default built-in behaviour; (b) check the status code against the specified list of expected status codes, but do not follow redirects.

This change provides a minimal Python reimplementation of the HTTP check program in fc-sensuplugins. This script supports checking the status codes against an explicitly specified list both with and without following any redirects first.

The `fc-monitor` script has correspondingly been updated to generate sensu checks which use the Python script instead of the monitoring-plugins program.

PL-132208

## Release process

Impact: internal

Changelog:

- [x] PR has internal ticket

## Security implications

  - The redirect handling for customer-provided HTTP checks should observe the principle of least astonishment when combined with an explicit list of expected status codes.
  - Tested under global maintenance on services15. The script runs properly and check results are populated correctly in Sensu.

### DIFF
--- a/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
+++ b/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
@@ -87,7 +87,10 @@ def main():
         err("empty hostname in URL")
 
     session = requests.Session()
-    # match monitoring-plugins defaults
+    # default redirect depth limit is 10, which matches
+    # monitoring-plugins -- chrome and firefox have default limits of
+    # 20. the browsers also perform redirection loop detection,
+    # however we don't implement this yet.
     session.max_redirects = 10
 
     if url.username is not None:
--- a/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
+++ b/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
@@ -109,7 +109,7 @@ def main():
     except requests.ConnectionError as ex:
         critical(f"could not connect to remote host: {ex}")
     except requests.TooManyRedirects as ex:
-        warning("maximum redirection depth exceeded")
+        critical("maximum redirection depth exceeded")
     except requests.Timeout as ex:
         critical("request to remote host exceeded timeout: 10s")
 
@@ -119,7 +119,7 @@ def main():
                 f"request returned server error status code: {resp.status_code}"
             )
         elif resp.status_code >= 400:
-            warning(
+            critical(
                 f"request returned client error status code: {resp.status_code}"
             )
     elif resp.status_code not in args.expect:
--- a/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
+++ b/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
@@ -102,13 +102,13 @@ def main():
         session.mount("https://", SourceAddressAdapter(source_address))
 
     try:
-        resp = session.get(target, allow_redirects=args.follow)
+        resp = session.get(target, allow_redirects=args.follow, timeout=10)
     except requests.ConnectionError as ex:
         critical("could not connect to remote host")
     except requests.TooManyRedirects as ex:
         warning("maximum redirection depth exceeded")
     except requests.Timeout as ex:
-        critical("request to remote host timed out")
+        critical("request to remote host exceeded timeout: 10s")
 
     if args.expect is None:
         if resp.status_code >= 500:
--- a/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
+++ b/pkgs/fc/sensuplugins/fc/sensuplugins/http.py
@@ -16,12 +16,12 @@ from requests_toolbelt.adapters.source import SourceAddressAdapter
 
 def err(msg):
     print(msg, file=sys.stderr)
-    sys.exit(1)
+    sys.exit(2)
 
 
 def unknown(msg):
     print(f"HTTP UNKNOWN - {msg}")
-    sys.exit(1)
+    sys.exit(3)
 
 
 def warning(msg):
@@ -31,7 +31,7 @@ def warning(msg):
 
 def critical(msg):
     print(f"HTTP CRITICAL - {msg}")
-    sys.exit(1)
+    sys.exit(2)
 
 
 def ok(msg):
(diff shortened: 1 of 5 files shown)


pkgs/fc/sensuplugins/fc/sensuplugins/http.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    if source_address is not None:
+        session.mount("http://", SourceAddressAdapter(source_address))
+        session.mount("https://", SourceAddressAdapter(source_address))
+
+    try:
+        resp = session.get(target, allow_redirects=args.follow)
+    except requests.ConnectionError as ex:
+        critical("could not connect to remote host")
```
ctheune: adding exception detail here would be great, or at least the specific connection error message

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
```diff
@@ -0,0 +1,131 @@
...
+    except requests.TooManyRedirects as ex:
+        warning("maximum redirection depth exceeded")
```
ctheune: I think this could/should be critical.

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
```diff
@@ -0,0 +1,131 @@
...
+    except requests.Timeout as ex:
+        critical("request to remote host timed out")
```
ctheune: adding exception detail here would be great, or if we can at least state the specific timeout threshold that was crossed

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
```diff
@@ -0,0 +1,131 @@
...
+
+    if args.expect is None:
+        if resp.status_code >= 500:
+            critical(
+                f"request returned server error status code: {resp.status_code}"
+            )
+        elif resp.status_code >= 400:
+            warning(
```
ctheune: this should be critical, too

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
sysvinit: This intentionally matches the behaviour of monitoring-plugins (https://github.com/monitoring-plugins/monitoring-plugins/blob/v2.3.5/plugins/check_http.c#L1599-L1602). Should we keep this behaviour?

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
sysvinit: This is likewise behaviour adopted from monitoring-plugins (https://github.com/monitoring-plugins/monitoring-plugins/blob/v2.3.5/plugins/check_http.c#L1217-L1221). Do we want to keep this behaviour too?

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
sysvinit: `requests` doesn't actually set any timeouts by default, which I missed when I skimmed through the documentation earlier this week, so currently this doesn't actually do anything. But yes, I'll add timeout configuration and put that into the message.

sysvinit: Fixed in [29b7899](https://github.com/flyingcircusio/fc-nixos/pull/940/commits/29b78996ebd2da0f16565d6363d9eacb29d910fe).

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
sysvinit: Fixed in [39ac04b](https://github.com/flyingcircusio/fc-nixos/pull/940/commits/39ac04b0256307c4a9dc61bd0c4b9d08e9284798).

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
ctheune: nope. let's make this critical.

ctheune: So, browsers have two behaviours that we can detect and predict whether this page will end up as an error in a user's browser:

1. They follow a limited number of redirects, before displaying an error page. Chrome and Firefox seem to be at 20 by default. IE8 supposedly was limiting it to 10.
2. They seem to detect redirection loops, so that could be an improvement for the future.

I'd be happy if we stay with 10 as a limit for now and make a note about this as well as the potential improvement to detect loops.

ctheune: So, browsers have two behaviours that we can detect and predict whether this page will end up as an error in a user's browser:

1. They follow a limited number of redirects, before displaying an error page. Chrome and Firefox seem to be at 20 by default. IE8 supposedly was limiting it to 10.
2. They seem to detect redirection loops, so that could be an improvement for the future.

I'd be happy if we stay with 10 as a limit for now and make a note about this as well as the potential improvement to detect loops.

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
ctheune: In general a 400 error code indicates an error on the client side. However: in monitoring we create specific requests that are supposed NOT to be in error and thus a change on the server side (e.g. a URL going away) can cause an error that otherwise would correctly be an error on the client side. In this case it becomes: either your monitoring is out of date, or we accidentally broke something on the server side. We can't distinguish this and should err on the side of caution and consider this critical.

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
sysvinit: See [88b4e75](https://github.com/flyingcircusio/fc-nixos/pull/940/commits/88b4e75496fa600d2d6641655a06b096faf1932f) and [48b0f76](https://github.com/flyingcircusio/fc-nixos/pull/940/commits/48b0f764d57285d3463383612333e3a647ce217a).

pkgs/fc/sensuplugins/fc/sensuplugins/http.py
sysvinit: Fixed in [48b0f76](https://github.com/flyingcircusio/fc-nixos/pull/940/commits/48b0f764d57285d3463383612333e3a647ce217a).

ctheune (approved):

PR #944, opened 2024-03-12
[21.05] ceph osds: prevent them from being stopped by fc-blockdev restarts

osnyx:
This is a less-intrusive alternative to #943.

This fixes the issue of all OSD service units being shut down and staying down when fc-blockdev.service was stopped (and started) again by a system switch.
The main issue was that stops of fc-blockdev propagated to stop the individual OSD units, but nothing ensured that they come up again. Only fc-ceph-osds-all.service was pulled in by the multi-user.target and stayed active all the time. By reducing the dependency type towards fc-blockdev, we prevent the osds-all unit from being stopped in the first place.

One side-effect of this is though that OSDs will start even if
fc-blockdev.service fails.

PL-132304 PL-132303

## Release process

Impact:
- internal only

Changelog:
- fix all OSDs being shutdown and staying down when `fc-blockdev.service` changes

- [x] PR has internal ticket

## Design notes

## Security implications

  - fix an issue that stopped services at unattended upgrades
  - systems have to transition from a functioning state to another functioning state before and after a maintenance upgrade 
  - manually reproduce the underlying cause
    - manual service stops
    - NixOS-triggered service stops
  - [ ] automated tests still pass
  - [x] manually verified the system unit behaviour on a dev host with
    - [x] manual start/stops of fc-blockdev
    - [x] NixOS-induced start/stops of fc-blockdev
    - [x] manually verify upgrading from a previous state works
      - disclaimer: mostly looked at the service stops and (re)starts done; the network on dev hosts currently breaks at switch time

### DIFF
--- a/nixos/roles/ceph/osd.nix
+++ b/nixos/roles/ceph/osd.nix
@@ -10,6 +10,13 @@ let
 
   cephPkgs = fclib.ceph.mkPkgs role.cephRelease;
 
+  osdServiceDeps = rec {
+    # Ceph requires the IPs to be properly attached to interfaces so it
+    # knows where to bind to the public and cluster networks.
+    wants = [ "network.target" "fc-blockdev.service" ];
+    after = wants;
+  };
+
   defaultOsdSettings = {
     # Assist speedy but balanced recovery
     osdMaxBackfills = 2;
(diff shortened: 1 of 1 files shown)


ctheune (commented):
This reflects the dependency much better. fc-blockdev is an optimization and I'd rather the service keeps running with sub-optimal settings and also doesn't get restarted as those changes really apply on the fly.

One question: does this change itself trigger the bug when the 'requires' is still active when rolling it out (and we still have the trigger around for the reason that fc-blockdev goes down?)

osnyx:
> One question: does this change itself trigger the bug when the 'requires' is still active when rolling it out (and we still have the trigger around for the reason that fc-blockdev goes down?)

No, I explicitly tested the upgrade from the current state (see security considerations), modulo the current brokenness of our dev network of course.

ctheune (approved):

PR #947, opened 2024-03-15
[21.05] ceph: workaround for 8ab330b2d48dc88a5b9698d191b75e0086fdcc0d

osnyx:
Unfortunately the changes in 8ab330b2d48dc88a5b9698d191b75e0086fdcc0d are not sufficient for preventing the shutdown of OSDs, as the decision
at switch-to-configuration time on which dependent units to stop when
fc-blockdev is stopped relies on the unit state _before_ the switch. As
long as the `requires` relation between OSDs and fc-blockdev is still
there in that state, OSDs will still be stopped and do not come up again
automatically.

This utilises a trick of using a `systemctl restart` instead of a stop
and start in switch-to-configuration, as the restart is already done
based on the new unit file and its dependency structures.
Also, changing this modifies the unit file itself and already triggers a
restart of fc-blockdev.service by itself, making sure the changes are
persisted.

This _might_ be removed in the future again, as its main purpose is
making sure the transition of fc-ceph-osd@.service dependencies happens
cleanly without pulling down the osds.

PL-132304

## Release process

Impact:
- internal only

Changelog:
- *prevents* automatical restarts of fc-blockdev, causing changed blockdev setting to only be applied at the next boot

- [x] PR has internal ticket

## Design notes

## Security implications

  - analysed system behaviour during rollout for regression
  - design workaround to keep ceph OSDs available
  - must not introduce any new known new regressions 
  - [x] manually tested on a host
  - [x] manually tested direct update from de4b8e543cc02d6f1106918417ee22bf2316c149 (prev commit of the `requires` change)  to this commit in staging

### DIFF
--- a/nixos/services/ceph/server.nix
+++ b/nixos/services/ceph/server.nix
@@ -124,6 +124,16 @@ in
       environment = {
         PYTHONUNBUFFERED = "1";
       };
+
+      # workaround for properly getting rid of the RequiredBy = fc-ceph-osd@…
+      # done in 8ab330b2d48dc88a5b9698d191b75e0086fdcc0d without stopping these
+      # dependant units one last time at switch-to-configuration time based on
+      # the *previous* unit state. This *restarts* fc-blockdev already with its
+      # current dependency declarations.
+      #
+      # The main purpose of this is to manage the transition from before the
+      # mentioned commit, this *may* be removed in the future at some point again.
+      stopIfChanged = false;  # do a restart with the new unit file instead of stop-start
     };
 
     boot.kernel.sysctl = {


nixos/services/ceph/server.nix
```diff
@@ -124,6 +124,14 @@ in
...
         PYTHONUNBUFFERED = "1";
       };
+
+      # temporary workaround, can be removed after this has successfully reached
+      # production:
+      # Prevent a restart of the unit at system switch, as this causes OSDs to
+      # restart if the system state _before_ the switch is still earlier than
+      # 8ab330b2d48dc88a5b9698d191b75e0086fdcc0d
+      stopIfChanged = false;
+      restartIfChanged = false;
```
ctheune: stopIfChanged must be "false" and restart must be "true", then we're good.

ctheune (changes requested):

ctheune (approved):

PR #951, opened 2024-03-25
[21.05] Improve resilience of system network state under upgrade conditions

sysvinit:
This change refactors the configuration for setting up VXLAN interfaces (once again), and ensures that systemd unit dependencies are correctly set in platform-provided units upon units provided by NixOS's scripted interface configuration automation. This should ensure that the host network configuration is more resilient under some upgrade conditions which have previously led to the loss of network connectivity. 

In particular, channel updates may cause a large number of network-related systemd units to be rebuilt, especially if some core system dependency has changed in the new channel version. When activating the first configuration based upon the new channel, units related to setup of the VXLAN overlay may be indirectly stopped due to dependencies on other units, particularly `PartOf=` dependencies. If these units are not explicitly picked up by switch-to-configuration, then they may not be restarted and leave the network in a broken state. Upstream-provided units for virtual network interfaces (e.g. bridge interfaces) have a number of dependencies set to ensure that they are properly restarted during upgrades, so the solution is to ensure that corresponding/matching dependencies are set on the platform-provided units for configuration of dummy and VXLAN interfaces.

Additionally, the unit for zebra has been updated to ensure that it is restarted when the addresses assigned to the `ul-loopback` interface are changed. The addresses on this interface are used as source addresses for routes inserted into the kernel by zebra, and if these addresses disappear then zebra will remove the routes and will not try to reinsert them without manual intervention.

PL-132305

## Release process

Impact:

Changelog: internal

- [x] PR has internal ticket

## Design notes

## Security implications

  - The upgrade process in and of itself should not break the host network configuration as a side-effect.
  - Manually verified on a dev host by toggling the most recent 23.05 channel update and then ensuring that the network doesn't break on the next switch-to-configuration.

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -466,7 +466,9 @@ in
               reload = ''
                 IFACE=${iface.layer2device}
 
-                # Set underlay address for virtual interface ${iface.layer2device}
+                # Set underlay address for virtual interface ${iface.layer2device}.
+                # Note that changing the VNI or destination port after the interface
+                # has been created is not supported.
                 ip link set $IFACE type vxlan local ${fclib.underlay.loopback}
 
                 # Set MTU and layer 2 address
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -396,12 +396,19 @@ in
               wantedBy = [ "multi-user.target" ];
               before = wantedBy;
               path = [ pkgs.nettools pkgs.procps fclib.relaxedIp ];
-              stopIfChanged = false;
+              reloadIfChanged = true;
               script = ''
                 IFACE=ul-loopback
 
                 # Create virtual interface underlay
                 ip link add $IFACE type dummy
+
+                ip link set $IFACE mtu ${toString fclib.underlay.mtu}
+
+                ${sysctlSnippet}
+              '';
+              reload = ''
+                IFACE=ul-loopback
 
                 ip link set $IFACE mtu ${toString fclib.underlay.mtu}
 
@@ -433,7 +440,7 @@ in
               requires = [ "network-addresses-ul-loopback.service" ];
               after = requires;
               partOf = requires;
-              stopIfChanged = false;
+              reloadIfChanged = true;
               path = [ pkgs.nettools pkgs.procps fclib.relaxedIp ];
               script = ''
                 IFACE=${iface.layer2device}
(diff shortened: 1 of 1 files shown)


nixos/platform/network.nix
```diff
@@ -367,23 +368,17 @@ in
             }
           )) bridgedInterfaces) ++
         (lib.optionals (!isNull fclib.underlay)
+          # loopback dummy device
           ([(lib.nameValuePair
-            "fc-lldp-to-altnames"
+            "ul-loopback-netdev"
```
ctheune, line 393: I'm assuming the naming convention here and the next onw (x-netdev and network-addresses-x) are to conform with upstream naming schema?

ctheune (approved):
Ok, so this is extremely involved and it's hard to read the diff.

If you like, we can walk through this together, otherwise ... I'm fine with taking this as is.

nixos/platform/network.nix
sysvinit: Yes, precisely. I've noted this in the commit message in 8878dd2.

nixos/platform/network.nix
```diff
@@ -434,6 +441,21 @@ in
 
                 ${sysctlSnippet}
               '';
+              reload = ''
+                IFACE=${iface.layer2device}
+
+                # Set underlay address for virtual interface ${iface.layer2device}
+                ip link set $IFACE type vxlan local ${fclib.underlay.loopback}
```
ctheune, line 472: we should re-apply all settings identical to the add statement above. maybe also make a statement here to keep this in sync (or can we share the code?)

ctheune (changes requested):
Reviewed this together in Jitsi, looks good. There's a small adjustment needed for the vxlan interface reload.

nixos/platform/network.nix
sysvinit: I've tested this just now, and it's not possible to change the VNI or the destination port for a VXLAN interface after it's been created -- the only thing we can actually change is VTEP IP address (as I have here already).

Is it worth keeping the VXLAN interface reload functionality? Or should we drop this seeing as we can only change a subset of the relevant settings at runtime? (I think there's a tradeoff between how likely it is for the loopback address to change at runtime vis-à-vis how likely it is for our global, static VLAN numbers to change...)

ctheune: Yeah. I'm fine with this. Maybe mention this in the comment?

sysvinit: Fixed in [ef6cba1](https://github.com/flyingcircusio/fc-nixos/pull/951/commits/ef6cba11b056d292ddf41d9ff34371c36eef6c46).

ctheune:
I'd be happy to merge this, but Hydra isn't giving feedback?

sysvinit:
Hydra seems to have run, everything looks green now.

ctheune (approved):

PR #957, opened 2024-04-04
nginx: use copytruncate for modsecurity log rotation

osnyx:
Because modsecurity is not re-opening its logfile after rotation and continues to write into the same file descriptor, we need to use `copytruncate`.
Better handling of that situation is stuck upstream for several years. https://github.com/owasp-modsecurity/ModSecurity-nginx/issues/121

We use the presence of `/var/log/modesc_*.log` as a heuristic for modsecurity being enabled, these files are now rotated with copytruncate.
All other nginx logs are still rotated by moving and reloading.
Note that, due to overlapping wildcard matches, this specific case got a higher logrotate match priority and needs an `ignoreduplicates`.

`copytruncate` is non-atomic and might loose some logs written between
copying and the truncation being done.

PL-132296

## Release process

Impact: -

Changelog:
- nginx, modsecurity: Fix improper rotation and continuous growth of `/var/log/nginx/modsec_audit.log`
  - log files of the pattern `/var/log/nginx/modsec_*.log` are now rotated via copying and then truncating the open log
  - all nginx modsecurity logfiles need to follow the naming scheme `modsec_*` to not accidentally grow unrotated. This is the naming convention suggested by our nginx modsecurity configuration examples.
  - Logs written directly at the time of the rotation might be lost.

- [x] PR has internal ticket

## Design notes

## Security implications

  - [x] work around an (upstream) issue with modsecurity to prevent disk from filling up with logs (avaialbility)
  - [x] rotation needs to be tested, overlap of logrotate rules
  - [x] manually tested rotation on testvm, including modification of dates and stamps

### DIFF
--- a/nixos/services/nginx/default.nix
+++ b/nixos/services/nginx/default.nix
@@ -469,16 +469,27 @@ in
         inherit virtualHosts;
       };
 
-      services.logrotate.settings = {
-        "/var/log/nginx/*.log" = {
+      services.logrotate.settings = let
+      commonRotate = {
           rotate = cfg.rotateLogs;
           create = "0644 ${nginxCfg.masterUser} nginx";
           su = "${nginxCfg.masterUser} nginx";
+        };
+        in {
+        "/var/log/nginx/modsec_*.log" = {
+          # need higher prio, because more-specific match.
+          # Our platform header options use priority 900, we need to chose a
+          # higher number here for using them.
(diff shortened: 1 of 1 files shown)


osnyx:
We need to decide whether this also needs a NixOS tests. I do have an idea on how to build one, but this would further delay this PR.

ctheune (approved):

PR #963, opened 2024-04-10
[21.05] devhost: replace hydra eval with metadata endpoint, various improvements (PL-132065)

PhilTaken:

relates to [PL-132065]

## Release process

Impact:

This PR implements a new subcommand "provision" for `fc-devhost` that takes a channel url and an image url instead of a hydra eval (see subcommand "ensure") to provision a new devhost vm. Functionally, the new subcommand is identical to the existing subcommand "ensure".
Additionally, it changes the destroy subcommand to accept an arbitrary number of vm names that will all be deleted.
The `destroy` subcommand is also aliased to `rm` similarly to `list` being aliased to `ls`.

Changelog:

- implement a new subcommand "provision" for `fc-devhost` that takes a channel url and an image url instead of a hydra eval (see subcommand "ensure") to provision a new devhost vm
- change the `destroy` subcommand to take an arbitrary (n >= 1) amount of vm names to destroy
- set the program name in `argparse` for more readable error messages that don't include a nix store hash

- [x] PR has internal ticket

## Design notes

## Security implications

We now allow users to provide explicit image URLs to bootstrap VMs from. Theoretically those could be arbitrary OS images including Windows, memtest, or some malware that then get access to the internet via NAT.

However, the VM architecture itself already provides similar vectors, like becoming root and spawning custom kernels. This is the whole point of running VMs anyway.

Functionality tested manually with @leona-ya on largo00.

### DIFF
--- a/nixos/roles/devhost/fc-devhost.py
+++ b/nixos/roles/devhost/fc-devhost.py
@@ -1,6 +1,7 @@
 import argparse
 import datetime
 import fcntl
+import hashlib
 import ipaddress
 import json
 import os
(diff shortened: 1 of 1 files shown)


leona-ya:
Blocked by https://gitlab.flyingcircus.io/flyingcircus/fc.directory/-/merge_requests/377

leona-ya (approved):

ctheune (approved):

PR #974, opened 2024-04-17
[21.05] ceph: add new `passive` option to disable daemons and timers

ctheune:
This is a knob for operators where we want to temporarily, but reliably disable all Ceph daemons and timers to perform maintenance but do not lose Ceph config and can run reboots and nixos-rebuild without accidentally turning the daemons back on.

Re PL-132400

## Release process

Impact:

none

Changelog:

* Add ability to switch Ceph servers into a `passive` mode during manual maintenances to avoid daemons activating automatically on boot and during agent runs. (PL-132400)

- [x] PR has internal ticket

## Design notes

This is a feature toggle ;)

## Security implications

Avoid downtimes.

manually tested

### DIFF
--- a/nixos/roles/ceph/osd.nix
+++ b/nixos/roles/ceph/osd.nix
@@ -195,6 +195,8 @@ in
       flyingcircus.services.ceph.cluster_network = head fclib.network.stb.v4.networks;
 
       systemd.services.fc-ceph-osds-all = rec {
+        enable = ! config.flyingcircus.services.ceph.server.passive;
+
         description = "All locally known Ceph OSDs (via fc-ceph managed units)";
         wantedBy = [ "multi-user.target" ];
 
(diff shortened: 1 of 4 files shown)


nixos/roles/ceph/osd.nix
```diff
@@ -223,6 +225,8 @@ in
       } // osdServiceDeps;
 
       systemd.services."fc-ceph-osd@" = rec {
+        enable = ! config.flyingcircus.services.ceph.server.passive;
```
osnyx, line 228: I'm not fully certain what `enable` causes for services that we only want to pull in by other depending units. Need to check that.

osnyx: This is fine, the default is `true` anyways, setting it to `false` just equals a mask.

osnyx (approved):

PR #976, opened 2024-04-19
Pl 127635 stabilize network reconfiguration

ctheune:

## Release process

Impact:

* All network interfaces will be restarted, causing a short service interruption of a few seconds. (PL-127635)

Changelog:

* We have rebuilt a major part of the network configuration logic to allow more fine grained online network reconfiguration in the future and to make it a lot more robust when making changes like changing IP addresses, renaming interfaces  and shifting between more complex architectures like tagged networks, bridge setups and even EVPN-VXLAN environments.

- [x] PR has internal ticket

## Design notes

## Security implications

Ensure availability by improving robustness.

Relying on the existing automated tests. Also extensively tested on cartman06 (Ceph) while developing and tested on cartman37 and kyle05 and tstest21 to check production rollout impact. In online scenarios this causes about 10-15 seconds of interruption in EVPN-VXLAN situations, less than 5 seconds in non-EVPN-VXLAN setups and VMs.

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -376,6 +376,8 @@ in
                 # Look up the desired interface's current name by mac address
                 interface="$(ip -j -d link show | jq -r '.[] | select(.linkinfo?.info_kind? == null) | select(.address == "${iface.mac}") | .ifname')"
 
+                # XXX properly handle missing MAC address and cause this unit to fail.
+
                 if [[ "$interface" != "${iface.link}" ]]; then
                   echo "Interface is currently known as $interface ..."
 
@@ -393,6 +395,7 @@ in
                     # but an altname, try to delete that as well.
                     # This looks funny, but we can use the altname to look up the
                     # link to remove the altname ...
+                    # don't do this if the name is already gone
                     ip l property del altname ${iface.link} dev ${iface.link} || true
                     counter=$((counter+1))
                   done
@@ -460,6 +463,7 @@ in
                   # Try, don't care if it fails. We check the success on the
                   # next loop.
                   ip l set ${iface.link} name eth$counter || true
+                  # don't do this if the name is already gone
                   ip l property del altname ${iface.link} dev ${iface.link} || true
                   counter=$((counter+1))
                 done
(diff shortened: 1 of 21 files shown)


nixos/platform/network.nix
```diff
@@ -357,63 +350,115 @@ in
...
+                exec 4</var/lock/interface-rename.lock
+
+                echo "Renaming to ${iface.link} ..."
+
+                echo "Acquiring interface renaming lock ..."
+                flock -x -w 60 4 || exit 1
+                echo "Got interface renaming lock ..."
+
+                # Look up the desired interface's current name by mac address
+                interface="$(ip -j -d link show | jq -r '.[] | select(.linkinfo?.info_kind? == null) | select(.address == "${iface.mac}") | .ifname')"
```
sysvinit, line 379: Another thought which comes to mind: what happens if there is no interface with the desired MAC address at all? In the world of `.device` units, this would cause dependents to fail with a dependency error (after a timeout?) - presumably this script would just exit with an error somewhere.

(Handling this better could reasonably be kicked down the road to the discussed Python implementation...)

nixos/platform/network.nix
```diff
@@ -532,13 +574,31 @@ in
...
+
+          (map (iface: (lib.nameValuePair
+            "${iface.interface}-netdev"
+            {
+              after = [ "network-pre.target" "network-addresses-${iface.link}.service" ];
+              # I'd love to bind those to zebra, but that creates a cycle.
+              bindsTo = [ "network-addresses-${iface.link}.service" ];
+            }
+          )) vxlanInterfaces) ++
+
```
sysvinit: I originally wrote these overrides because NixOS was incorrectly generating dependencies on `.device` units for virtual interfaces where this makes no sense.

As this branch is using a patched version of nixpkgs which gets rid of device unit dependencies completely, these overrides specify redundant dependencies (e.g. the generated unit file for `brfe-netdev.service` has a duplicated After= dependency on `network-addresses-vxfe.service`).

Both of those overrides can be removed, if we're not making any further local customisations to dependencies for these interfaces.

sysvinit: Same applies to the override at line 521 for `network-addresses-ul-loopback.service`, this can also be removed.

sysvinit (changes requested):

nixos/platform/network.nix
ctheune: will do, made a note in the script so i won't forget during python conversion.

nixos/platform/network.nix
ctheune: yup, those are right. reviewed together.

PR #997, opened 2024-05-04
Remove deprecated postgresql ensurePermissions

laalsaas:
Replaces deprecated option `postgresql.ensureUsers.*.ensurePermissions` with a custom systemd unit that ensures database privileges for k3s. This was the only place where we used the option.

PL-132355

## Release process

Impact:

Changelog:

- [x] PR has internal ticket

## Design notes

## Security implications

  - This test should not change the behaviour compared to previous releases.
  - I locally changed the postgress-version for k3s to 16 and ran the test cases. All tests passed.

### DIFF
--- a/nixos/roles/k3s/server.nix
+++ b/nixos/roles/k3s/server.nix
@@ -304,10 +304,30 @@ in {
       ensureDatabases = [ "kubernetes" ];
       ensureUsers = [ {
         name = "root";
-        ensurePermissions = {
-          "DATABASE kubernetes" = "ALL PRIVILEGES";
-        };
       } ];
+    };
+
+    systemd.services.kubernetes-ensure-db-permissions = {
+      description = "Ensure the root user has all permissions on kubernetes db";
+      wantedBy = [ "k3s.service" ];
+      before = [ "k3s.service" ];
+      requires = [ "postgresql.service" ];
+      script = ''
+        PSQL="${config.services.postgresql.package}/bin/psql --port=${toString config.services.postgresql.port}"
+        for i in {1..10}; do
+          $PSQL -d postgres -c "" 2> /dev/null && break
+          sleep 0.5
+
+          if [[ $i -eq 10 ]];then
+            echo "couldn't establish connection to postgres"
+            exit 1
+        fi
+        done
+          $PSQL -tAc 'GRANT ALL PRIVILEGES ON DATABASE kubernetes TO "root"'
+      '';
+      serviceConfig = {
+        type = "oneshot";
+      };
     };
 
     flyingcircus.services.sensu-client = {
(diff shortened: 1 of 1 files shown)


nixos/roles/k3s/server.nix
```diff
@@ -304,12 +304,32 @@ in {
       ensureDatabases = [ "kubernetes" ];
       ensureUsers = [ {
         name = "root";
-        ensurePermissions = {
-          "DATABASE kubernetes" = "ALL PRIVILEGES";
-        };
       } ];
     };
 
+    systemd.services.kubernetes-ensure-db-permissions = {
```
leona-ya: Is there a reason why to use a custom systemd service instead of `services.postgresql.ensureUsers.*.ensureDBOwnership`?

EDIT: I see. The username is differently. Still the other comments apply. This might be something we want to have in upstream nixos, but might also not (this is a huge discussion topic unfortunately).

nixos/roles/k3s/server.nix
```diff
@@ -304,12 +304,32 @@ in {
...
+        for i in {1..10}; do
+          $PSQL -d postgres -c "" 2> /dev/null && break
+          sleep 0.5
+
+          if [[ $i -eq 10 ]];then
+            echo "couldn't establish connection to postgres"
+            exit 1
+        fi
+        done
+          $PSQL -tAc 'GRANT ALL PRIVILEGES ON DATABASE kubernetes TO "root"'
```
leona-ya, line 326: This will most likely also not work on PostgreSQL 15+. The user needs to have permissions on the `public` schema too or ownership of the DB. See https://github.com/NixOS/nixpkgs/issues/216989 for more information.

nixos/roles/k3s/server.nix
```diff
@@ -304,12 +304,32 @@ in {
...
         name = "root";
-        ensurePermissions = {
-          "DATABASE kubernetes" = "ALL PRIVILEGES";
-        };
       } ];
     };
 
+    systemd.services.kubernetes-ensure-db-permissions = {
+      description = "Ensure the root user has all permissions on kubernetes db";
+      wantedBy = [ "k3s.service" ];
```
leona-ya, line 312: I think this should also have a 
```nix
Before = [ "k3s.service" ]
```

leona-ya (dismissed):

nixos/roles/k3s/server.nix
```diff
@@ -304,12 +304,33 @@ in {
       ensureDatabases = [ "kubernetes" ];
       ensureUsers = [ {
         name = "root";
-        ensurePermissions = {
-          "DATABASE kubernetes" = "ALL PRIVILEGES";
-        };
       } ];
     };
 
+    systemd.services.kubernetes-ensure-db-permissions = {
```
dpausp: better follow the naming scheme for existing services: `fc-k3s-ensure-db-permissions`

dpausp (changes requested):

dpausp (approved):

PR #998, opened 2024-05-06
[21.05] RadosGW: selectively re-enable connection tracking

sysvinit:
In #988 and #989 connection tracking was disabled on the STO and STB networks, to reduce the overhead in the kernel connection table. This however breaks RadosGW clients using the STO interface (e.g. the rgw.local reverse proxy on the KVM servers), as the RGW hosts use iptables REDIRECT rules to steer incoming connections on port 7480 to the radosgw process listening on port 80 (as a workaround for another problem). The REDIRECT target relies on conntrack in order to rewrite the port numbers on incoming and outgoing packets correctly, and disabling conntrack renders these rules ineffective.

This commit selectively re-enables conntrack in the RGW role on the STO interface, so that only connections to and from the RGW process are tracked. This allows the REDIRECT rules to function correctly again.

PL-132519

## Release process

Impact: internal.

Changelog: none.

- [x] PR has internal ticket

## Design notes

## Security implications

  - No specific requirements, this change fixes the integration of Ceph in the platform downstream of other changes made as part of the VXLAN integration.
  - Manually verified in DEV. Previously non-functional RGW access now works over the STO network from a KVM host again with this change applied to an RGW host.

### DIFF
--- a/nixos/roles/ceph/rgw.nix
+++ b/nixos/roles/ceph/rgw.nix
@@ -129,30 +129,37 @@ in
       networking.firewall.extraCommands =
         let
           srv = fclib.network.srv;
-        in
-        ''
-          set -x
-          # Accept traffic from S3 gateways from within the SRV network.
-          ip46tables -w -t nat -N fc-nat-pre
+          sto = fclib.network.sto;
+        in lib.mkMerge [
+          (lib.mkOrder 700 ''
+            # Ensure that conntrack is enabled for RGW connections using port redirects
+            iptables -w -t raw -A fc-raw-prerouting -i ${sto.interface} -p tcp --dport 7480 -j RETURN
+            iptables -w -t raw -A fc-raw-output -o ${sto.interface} -p tcp --sport 80 -j RETURN
+          '')
+          (''
+            set -x
+            # Accept traffic from S3 gateways from within the SRV network.
+            ip46tables -w -t nat -N fc-nat-pre
 
-        '' + (lib.concatMapStringsSep "\n"
-          (net: ''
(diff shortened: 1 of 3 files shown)


nixos/roles/ceph/rgw.nix
```diff
@@ -129,30 +129,37 @@ in
...
-          (net: ''
-            iptables -A nixos-fw -i ${srv.interface} -s ${net} -p tcp --dport 80 -j ACCEPT
-            # PL-130368 Fix S3 presigned URLs
-            iptables -t nat -A fc-nat-pre -p tcp --dport 7480 -j REDIRECT --to-port 80
+          sto = fclib.network.sto;
+        in lib.mkMerge [
+          (lib.mkOrder 700 ''
+            # Ensure that conntrack is enabled for RGW connections using port redirects
+            iptables -w -t raw -A fc-raw-prerouting -i ${sto.interface} -p tcp --dport 7480 -j RETURN
+            iptables -w -t raw -A fc-raw-output -o ${sto.interface} -p tcp --sport 80 -j RETURN
```
ctheune: this is v4 only, is that on purpose?

ctheune: are we only suppressing connection tracking on v4?

ctheune (changes requested):
generally looks good, the v4/v6 asymmetry makes me a bit uneasy

nixos/roles/ceph/rgw.nix
sysvinit: > are we only suppressing connection tracking on v4?

Yes, that's correct. When you turned off the connection tracking you only did so for v4.

sysvinit: Fixed in 3cfa511

ctheune (approved):

PR #1009, opened 2024-05-22
[21.05] Disable resource group firewall rules on routers

sysvinit:
Normally, hosts will rebuild and reload their firewall rules when the SRV addresses in ENC for other hosts in the same resource group change -- this is intended to ensure that only hosts in the same resource group may reach each other over SRV. This is however less desirable on routers, as firewall reloads briefly cause all incoming packets to be dropped, which interrupts connectivity for other hosts.

This commit introduces a new NixOS option to allow the firewall rules restricting access to only the hosts in the same resource group to be disabled, replacing it with rules which allow all connections on the SRV interface. This rule is then enabled in the router role to ensure that new IP addresses in ENC does not cause firewall reloads. In the case where the RG rules are disabled, I've added a hard-coded exception to deny access to Telegraf, as the location statshost is already whitelisted, and Telegraf should not otherwise be world-readable.

PL-132581

## Release process

Impact: internal.

Changelog: none.

- [x] PR has internal ticket

## Design notes

  - This option is for internal platform use only, and is hence only lightly documented.

## Security implications

  - Reducing frequency of firewall reloads on routers improves service availability.
  - Potentially sensitive metrics endpoints of platform-internal infrastructure should not be world-readable.
  - Tested with a test KVM server and test router in DEV.
  - No changes in the KVM host firewall; router firewall is left open on the SRV interface, however Telegraf port remains protected.

### DIFF
--- a/nixos/platform/firewall.nix
+++ b/nixos/platform/firewall.nix
@@ -192,7 +192,7 @@ in
              then (lib.optionalString (rgRules != "")
                "# Accept traffic within the same resource group.\n${rgRules}\n\n")
              else ''
-               # Accept traffic on the SRV interface
+               # Accept traffic on the SRV interface, but disallow arbitrary access to Telegraf
                ip46tables -A nixos-fw -i ${fclib.network.srv.interface} -p tcp --dport 9126 -j nixos-fw-refuse
                ip46tables -A nixos-fw -i ${fclib.network.srv.interface} -j nixos-fw-accept
              ''
--- a/nixos/platform/firewall.nix
+++ b/nixos/platform/firewall.nix
@@ -80,6 +80,12 @@ in
         0 (emerg), 1 (alert), 2 (crit), 3 (error), 4 (warning), 5 (notice), 6 (info), 7 (debug)
       '';
     };
+    enableSrvRgFirewall = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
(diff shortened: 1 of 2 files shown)


dpausp (approved):

nixos/platform/firewall.nix
```diff
@@ -181,9 +187,16 @@ in
...
-            (lib.mkOrder 1200 (lib.optionalString
-              (rgRules != "")
-              "# Accept traffic within the same resource group.\n${rgRules}\n\n"))
+          (lib.mkOrder 1200
+            (if cfg.firewall.enableSrvRgFirewall
+             then (lib.optionalString (rgRules != "")
+               "# Accept traffic within the same resource group.\n${rgRules}\n\n")
+             else ''
+               # Accept traffic on the SRV interface
+               ip46tables -A nixos-fw -i ${fclib.network.srv.interface} -p tcp --dport 9126 -j nixos-fw-refuse
```
osnyx, line 196: This might benefit from a comment explaining what we're refusing here.

sysvinit: Fixed in [fa37aef](https://github.com/flyingcircusio/fc-nixos/pull/1009/commits/fa37aef93c60c7b1e57f7164ac38d0adc75e7e48)?

osnyx: Fine for me.

PR #995, opened 2024-05-03
[21.05] ceph: clean up mon/mgr/rgw dependencies

ctheune:
Required turning the mon and mgr units into simple units.

Fixes PL-132473

@flyingcircusio/release-managers

## Release process

Impact:

n/a

Changelog:

n/a

### PR release workflow (internal)

- [x] PR has internal ticket
- [x] internal issue ID (PL-…) part of branch name
- [x] internal issue ID mentioned in PR description text
- [x] ticket is on Platform agile board
- [x] ticket state set to *Pull request ready*
- [ ] if ticket is more urgent than within the next few days, directly contact a member of the Platform team

## Design notes

- [x] Provide a feature toggle if the change might need to be adjusted/reverted quickly depending on context. Consider whether the default should be `on` or `off`. Example: rate limiting.
- [x] All customer-facing features and (NixOS) options need to be discoverable from documentation. Add or update relevant documentation such that hosted and guided customers can understand it as well.

## Security implications

- [x] [Security requirements](https://wiki.flyingcircus.io/System_Development_Guideline#Security_requirement_principles_and_testing) defined? (WHERE)

Limit unnecessary downtimes.

- [x] Security requirements tested? (EVIDENCE)

Manually tested on cartman06, relying on existing test suite.

PR #1021, opened 2024-06-06
[21.05] Fix fc-ceph-mon autostart after boot

osnyx:
In https://github.com/flyingcircusio/fc-nixos/pull/995 we added a start precondition of a mountpoint existing to fc-ceph-mon.service:

`ConditionPathIsMountPoint = "/srv/ceph/mon/ceph-${config.networking.hostName}";`

Unfortunately, the actual mounting of the mon volume to that path is done by exactly that service that is prevented to start by that condition.

The main reason to introduce this condition was just to make the tests less noisy when a test VM started for the first time, without having a mon volume created yet. We can simply remove that condition at the cost of having more red herrings in the test output.

This PR additionally:
- adds a test for ceph service autostarts after a reboot
- slows down the restart intervals of failed mons and mgrs
- adapts the test code to the new service structure

PL-132667
## Release process

Impact: internal only
- ceph mon and mgr will restart

Changelog:
- fix autostart of fc-ceph-mon.service after boot

- [x] PR has internal ticket

## Design notes

## Security implications

  - fix a known regression
  - adds a regression test (which unfortunately fails when being run by Hydra)
  - servers shall boot into a service-ready state automatically
  - [x] deliberately ignore the test failure of the regression test occuring only in hydra, as manual tests succeeded
  - [x] previously existing automated tests still pass
  - [x] manually tested on a host in dev cluster (including reboot)

### DIFF
--- a/nixos/roles/ceph/mon.nix
+++ b/nixos/roles/ceph/mon.nix
@@ -156,7 +156,11 @@ in
         serviceConfig = {
           Type = "simple";
           ExecStart = " ${cephPkgs.fc-ceph}/bin/fc-ceph mon activate --as-systemd-unit";
+          # try to restart after 5s for 6 attempts, afterwards wait for a minute between attempts
           Restart = "always";
+          RestartSec = "5s";
+          RestartSteps = 6;
+          RestartMaxDelaySec = "1min";
         };
       };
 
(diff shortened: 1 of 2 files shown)


osnyx:
Unfortunately, the "Verify all services are up after a reboot"
reproducibly fails when run by Hydra. It does _not_ fail when running
the test manually via SSH, also a manual test of rebooting a dev host
works fine.

As mon activation is definitely broken as of now, this changeset is an
improvement. I propose merging this without the regression test and
observing how it performs in staging.

The hydra test failure is being dealt with in PL-132687 separately.

ctheune (approved):
The tests which were commented out are weird as they only break in hydra. This needs further investigation, but we also need to get back into the state where mons start automatically. The tests work when run manually.

PR #996, opened 2024-05-03
[FC-36891] fix varnish monitoring

PhilTaken:
The default monitoring check for the varnish service relied on Varnish to run on a specific address.
Since this address can be changed through nix, the monitoring check needs to be adjusted to correctly infer the host and port that varnish is running on.

@flyingcircusio/release-managers

FC-36891

## Release process

Impact: -

Changelog:
- varnish: monitoring checks now support customised listening addresses and ports

### PR release workflow (internal)

- [x] PR has internal ticket
- [x] internal issue ID (PL-…) part of branch name
- [x] internal issue ID mentioned in PR description text
- [ ] ticket is on Platform agile board
- [ ] ticket state set to *Pull request ready*
- [ ] if ticket is more urgent than within the next few days, directly contact a member of the Platform team

## Design notes

- [ ] Provide a feature toggle if the change might need to be adjusted/reverted quickly depending on context. Consider whether the default should be `on` or `off`. Example: rate limiting.
- [ ] All customer-facing features and (NixOS) options need to be discoverable from documentation. Add or update relevant documentation such that hosted and guided customers can understand it as well.

## Security implications

- [x] [Security requirements](https://wiki.flyingcircus.io/System_Development_Guideline#Security_requirement_principles_and_testing) defined? (WHERE)
  - [x] the varnish monitors the correct address when not using the default
- [x] Security requirements tested? (EVIDENCE)

```
phil@test01 ~ $ sensu-client-show-config | jq '.checks.varnish_http'
{
  "command": "check_http -H 172.22.57.135 -p 8080 -c 10 -w 3 -t 20 -e HTTP",
  "interval": 60,
  "notification": "varnish port 8080 HTTP response",
  "standalone": true,
  "timeout": null,
  "ttl": null,
  "warnIsCritical": false
}

phil@test01 ~ $ cat /etc/local/nixos/varnish_http.nix
{ lib, ... }:
{
  services.varnish.http_address = lib.mkForce "172.22.57.135:8080";
}

phil@test01 ~ $
```

PR #1035, opened 2024-06-21
Fix varnish monitoring 2.0

PhilTaken:
the default monitoring check relied on varnish to run on a specific address.
since this address can be changed, the sensu check needs to be adjusted to correctly infer the host and port that varnish is running on

This pr is based on https://github.com/flyingcircusio/fc-nixos/pull/996 but fixes an issue where an incorrect address was inferred when multiple addresses were passed to varnish.

FC-36891

## Release process

Impact:

Changelog:

- [x] PR has internal ticket

## Design notes

## Security implications

  - [x] the varnish monitors the correct address when not using the default

```
phil@test01 ~ $ sensu-client-show-config | jq '.checks.varnish_http'
{
  "command": "/nix/store/4c1bx0qnfylqxqr4gi72ws0snvgg3y0q-check-varnish-http",
  "interval": 60,
  "notification": "varnish port 8008 HTTP response",
  "standalone": true,
  "timeout": null,
  "ttl": null,
  "warnIsCritical": false
}

phil@test01 ~ $ cat /nix/store/4c1bx0qnfylqxqr4gi72ws0snvgg3y0q-check-varnish-http
#!/nix/store/516kai7nl5dxr792c0nzq0jp8m4zvxpi-bash-5.2p32/bin/bash
ADDRS=$(/nix/store/m7agwiiz62q5nq7fhsj6vgd4b86i6rqi-varnish-7.4.3/bin/varnishadm debug.listen_address | awk '/([0-9.]+\.)+/ { print $2":"$3; }')
for ADDR in $ADDRS; do
  host=$(echo $ADDR | cut -d ":" -f 1)
  port=$(echo $ADDR | cut -d ":" -f 2)

  echo "checking host '$host' on port '$port'"
  /nix/store/j6r4xyhlr11wk7pad33y1f18c0wmj8zj-monitoring-plugins-2.3.5/bin/check_http -H $host -p $port -c 10 -w 3 -t 20 -e HTTP
done


phil@test01 ~ $ sudo -u sensuclient /nix/store/4c1bx0qnfylqxqr4gi72ws0snvgg3y0q-check-varnish-http
checking host '172.22.57.135' on port '8008'
HTTP OK: Status line output matched "HTTP" - 520 bytes in 0.001 second response time |time=0.000553s;3.000000;10.000000;0.000000;20.000000 size=520B;;;0;
checking host '127.0.0.1' on port '8008'
HTTP OK: Status line output matched "HTTP" - 526 bytes in 0.000 second response time |time=0.000360s;3.000000;10.000000;0.000000;20.000000 size=526B;;;0;

phil@test01 ~ $
```

### DIFF
--- a/nixos/roles/webproxy.nix
+++ b/nixos/roles/webproxy.nix
@@ -63,7 +63,16 @@ in
         };
         varnish_http = {
           notification = "varnish port 8008 HTTP response";
-          command = "check_http -H localhost -p 8008 -c 10 -w 3 -t 20 -e HTTP";
+          command = "${pkgs.writeShellScript "check-varnish-http" ''
+            ADDRS=$(${cfg.package}/bin/varnishadm debug.listen_address | awk '/([0-9.]+\.)+/ { print $2":"$3; }')
+            for ADDR in $ADDRS; do
+              host=$(echo $ADDR | cut -d ":" -f 1)
+              port=$(echo $ADDR | cut -d ":" -f 2)
+
+              echo "checking host '$host' on port '$port'"
+              ${pkgs.monitoring-plugins}/bin/check_http -H $host -p $port -c 10 -w 3 -t 20 -e HTTP
+            done
+          ''}";
         };
       };
 
--- a/nixos/services/varnish/default.nix
+++ b/nixos/services/varnish/default.nix
@@ -120,6 +120,12 @@ in {
     http_address = mkOption {
       type = types.str;
       default = "*:8008";
+      description = ''
+        The http address for the varnish service to listen on.
+        Unix sockets can technically be used for varnish, but are not currently supported on the FCIO platform due to monitoring constraints.
+        Multiple addressess can be specified in a comma-separated fashion in the form of `address[:port][,address[:port][...]`.
+        See `varnishd(1)` for details.
+      '';
     };
     virtualHosts = mkOption {
       type = types.attrsOf (types.submodule ({ name, config, ... }: {


nixos/services/varnish/default.nix
```diff
@@ -76,9 +76,18 @@ in {
     http_address = mkOption {
       type = types.str;
       default = "*:8008";
+      description = ''
```
osnyx, line 123: Adding a proper option description is good. We might want to document the approach of how to add multiple listen addresses here as well.

nixos/roles/webproxy.nix
```diff
@@ -63,9 +63,29 @@ in
           command = "${cfg.package}/bin/varnishadm status";
           timeout = 180;
         };
-        varnish_http = {
-          notification = "varnish port 8008 HTTP response";
-          command = "check_http -H localhost -p 8008 -c 10 -w 3 -t 20 -e HTTP";
+        varnish_http = let
+          # the plattform module passes multiple addresses in the form of '<address>:<port> -a <address>:<port> -a <address>:<port>'
```
osnyx: The host platform module also just provides a string field, so it depends on what the user enters there.

What you've written is one approach of how to pass multiple addresses, but according to [varnish docs](http://varnish-cache.org/docs/7.5/users-guide/command-line.html) another one is comma-separation of addresses like `-a '0.0.0.0:8080,[::]:8081'`. This needs to be taken into consideration as well when processing the string, quotation marks may be involved as well.

nixos/roles/webproxy.nix
```diff
@@ -63,9 +63,29 @@ in
...
+          first_address = builtins.elemAt (lib.splitString " " cfg.http_address) 0;
```
osnyx: The question here is what we want to achieve with our monitoring:

- monitor the varnish server process and whether it works -> it is sufficient to reach varnish via *any* of its multiple listen addresses, in this implementation we always take the first one
- monitor the listen address configuration as well: is varnish actually reachable via all configured IP-port configurations? Did we mis-type something, did one of the IPs somehow change, etc…
  - in that case, we need to run the http check against __all__ the configured listen addresses, not just the first one.

osnyx (commented):
While the code in general looks plausible, there are some general questions and decisions we need to take before investigating the details makes sense.

Happy to discuss this with you if needed or I misunderstood something.

osnyx:
rebased onto latest dev branch to fix a test failure in docs.

osnyx:
We also noticed that the docs regarding monitoring are outdated. It'd be really neat if you updated that part to reflect the current state. Can be a brief overview of what monitoring is happening, no need for deep details.

https://doc.flyingcircus.io/roles/fc-24.05-production/webproxy.html

nixos/roles/webproxy.nix
osnyx: After discussing this, we decided that adding a check for each of the parsed configured listen addresses is necessary, as each of them can serve a meaningful service.

nixos/roles/webproxy.nix
osnyx: We are aware that the current way of parsing does not cover all potential cases. Configuring one or multiple addresses with this inline string was a bad idea anyways, but we have to deal with it for the time being.

We dedice to leave the parsing as is, unexpected address formatting will fail at eval time. We will discover failing evals in our staging environment and then decide if this is bearable or we'll roll back the PR.

PhilTaken:
rebased to 24.05-dev, still in draft status

PhilTaken:
rebased to fc-24.05-dev and implemented a different approach of getting to the listen addresses after talking about it with @ctheune. this approach should be more stable but has the drawback of limiting the implementation to a single check for all listen addresses. the logic is a lot simpler however

nixos/roles/webproxy.nix
```diff
@@ -63,7 +63,16 @@ in
         };
         varnish_http = {
           notification = "varnish port 8008 HTTP response";
-          command = "check_http -H localhost -p 8008 -c 10 -w 3 -t 20 -e HTTP";
+          command = "${pkgs.writeShellScript "check-varnish-http" ''
```
ctheune, line 66: i think there's some sudo magic missing here.

ctheune (changes requested):

nixos/services/varnish/default.nix
PhilTaken: https://github.com/flyingcircusio/fc-nixos/pull/1035/files#diff-8853157fe79ca26de07ebefb6906f31f01516e5d7df621eb158402a52aa2b220R126 

is this ok?

osnyx: What about the format `'<address>:<port> -a <address>:<port> -a <address>:<port>'`? IIRC this is what an earlier version of the PR even expected as a baseline, so I'm rather certain we have that format in some setups.

But if the parser supports both formats, I'm fine with only suggesting the better looking one here in the docs.

ctheune: Yeah that second form is allowed by varnish but is pretty much a hack in this context, let's not mention this here.

ctheune (approved):

PR #1048, opened 2024-07-02
always restart gitlab-mailroom

laalsaas:
PL-131890

## Release process

Impact:

Changelog:

- [x] PR has internal ticket

## Design notes

## Security implications

  - This should make incoming mail on gitlab more reliable, so problems like the one in the ticket don't occur.
  - ran the gitlab test locally, with a slightly modified config to activate the gitlab-mailroom service
  - verified that it was restarted multiple times

### DIFF
--- a/nixos/roles/gitlab.nix
+++ b/nixos/roles/gitlab.nix
@@ -268,7 +268,7 @@ in
     # Needed for Git via SSH.
     users.users.gitlab.extraGroups = [ "login" ];
 
-    systemd.services.gitlab-mailroom.serviceConfig.Restart = lib.mkForce "always";
+    systemd.services.gitlab-mailroom.serviceConfig.Restart = lib.mkOverride 90 "always";
   })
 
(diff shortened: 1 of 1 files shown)


dpausp (approved):

nixos/roles/gitlab.nix
```diff
@@ -267,6 +267,8 @@ in
 
     # Needed for Git via SSH.
     users.users.gitlab.extraGroups = [ "login" ];
+
+    systemd.services.gitlab-mailroom.serviceConfig.Restart = lib.mkForce "always";
```
dpausp: `lib.mkOverride 90` would be a better option for such cases to allow others to use mkForce, but it's acceptable here to use mkForce.

dpausp:
```suggestion
    systemd.services.gitlab-mailroom.serviceConfig.Restart = lib.mkOverride 90 "always";
```

PR #1086, opened 2024-08-21
devhost: cleanup unused VMs (PL-132737)

leona-ya:
PL-132737

## Release process

Impact:

- devhost VMs will be shut down after 14 days and deleted after 31 days now if they don't get a deployment in the mean time

Changelog:

- devhost VMs will be shut down after 14 days and deleted after 31 days now if they don't get a deployment in the mean time (PL-132737)

- [x] PR has internal ticket

## Design notes

## Security implications

  - This is according the process as discussed internally and with customers
  1. VMs created before this PR should not be stopped and deleted directly, but their countdown should then start.
  2. Data should not be lost 
 
  - Change tested on largo00
  1. VMs that were created before the `last_deploy_date` config attribute was introduced - i.e. before this PR arrived on machines - get this attribute with the first cleanup run and a value of `now`
  2. devhost VMs are intended to be short-lived, so by concept they should not have any valuable data. Also by shutting down the VM before it gets deleted, users of the VM get informed when they want to use it

### DIFF
--- a/nixos/roles/devhost/fc-devhost.py
+++ b/nixos/roles/devhost/fc-devhost.py
@@ -42,6 +42,33 @@ def check_if_nbd_device_is_used(number):
         return f.read() != "0"
 
 
+def write_nix_file(nix_file_path, cfg):
+    # Nixify the alias list
+    nix_aliases = " ".join(map(lambda x: f'"{x}"', cfg["aliases"]))
+    with open(nix_file_path, mode="w") as f:
+        f.write(
+            textwrap.dedent(
+                f"""\
+        # DO NOT TOUCH!
+        # Managed by fc-devhost
+        {{ ... }}: {{
+          flyingcircus.roles.devhost.virtualMachines = {{
+            "{cfg['name']}" = {{
+              enable = {"true" if cfg['online'] else "false"};
+              id = {cfg['id']};
+              memory = "{cfg['memory']}";
+              cpu = {cfg['cpu']};
+              srvIp = "{cfg['srv-ip']}";
+              srvMac = "{cfg['srv-mac']}";
+              aliases = [ {nix_aliases} ];
+            }};
+          }};
+        }}
+        """
+            )
+        )
+
+
 def generate_enc_json(cfg, channel_url):
     return json.dumps(
         {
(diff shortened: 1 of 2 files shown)


nixos/roles/devhost/fc-devhost.py
```diff
@@ -317,7 +325,21 @@ def ensure(
...
                 os.rename(self.image_file_tmp, self.image_file)
-            else:
+
+            # Make sure the VM is now online, even if was previously offline
+            run("fc-manage", "-v", "-b")
+
+            fcntl.flock(self.lockfile, fcntl.LOCK_UN)
+            # Wait for the VM to get online
+            print("Waiting for VM to become pingable ...")
+            while True:
```
ctheune, line 335: FYI - nothing to change right away: In fc.qemu and other places we have timeout helpers for loops like this.

nixos/roles/devhost/fc-devhost.py
```diff
@@ -375,6 +385,37 @@ def list_vms(self, long_format, user=None, location=None):
...
         fcntl.flock(self.lockfile, fcntl.LOCK_EX)
+
+        print("Cleaning up the devhost now.")
+        vm_shut_down = False
+        for vm_cfg in list_all_vm_configs():
+            if "last_deploy_date" not in vm_cfg:
+                vm_cfg["last_deploy_date"] = (
+                    datetime.datetime.utcnow().isoformat()
+                )
+                with open(CONFIG_DIR / f"{vm_cfg['name']}.json", mode="w") as f:
```
ctheune, line 396: FYI - nothing to do now: There's a pattern around (also fc.qemu) that provides a safe "write and replace" approach that ensures atomicity. This pattern here is fragile in crash/power-off situations and can leave defunct files behind.

nixos/roles/devhost/fc-devhost.py
```diff
@@ -375,6 +385,37 @@ def list_vms(self, long_format, user=None, location=None):
...
+                    f.write(json.dumps(vm_cfg))
+
+            if datetime.datetime.fromisoformat(vm_cfg["last_deploy_date"]) < (
```
ctheune, line 399: Small pattern nitpick: computing the age and comparing it is a tiny bit easier to read than comparing two absolute dates.

ctheune (approved):

PR #1088, opened 2024-08-26
[21.05] Sensu monitoring for EVPN control plane

sysvinit:
This change adds a sensu check script for detecting sync issues between the EVPN control plane (i.e. the FRR daemons) and the kernel, along with supporting infrastructure.

The check script itself is `check_rib_integrity.py` which loads and compares either the IPv4 unicast RIB (for monitoring of the underlay network routing) or the EVPN RIB (for monitoring of the overlay network configuration, i.e. mostly MAC addresses). If there are mismatches between the state in either the kernel or FRR, then the script will return a critical result. In case of the check being critical, the script will output a list of problems detected (e.g. which IP addresses or MAC addresses are in the kernel but not FRR and vice versa, or which destinations for which the kernel is missing nexthops) in a space-separated format which should be straightforwardly machine-readable.

Given that this script has to deal with relatively complex system network state, I've additionally expanded our FRR test suite to check that the sensu script detects the relevant faults correctly. In order to properly simulate an EVPN environment, I've written a `ping-on-tap.py` script, which opens a tap interface and then respond to ARP and ICMP echo on that interface. The test VMs then run one or more instances of this script so the kernel can dynamically learn MAC addresses from the python script on the tap interfaces, which are then exported to its peers by FRR.

I also realised that we never actually added the FRR tests to the default Hydra tests (as far as I can tell), so I've fixed that as well.

PL-132595

## Release process

Impact: internal.

Changelog: none.

- [x] PR has internal ticket

## Design notes

## Security implications

  - This change adds monitoring for the VXLAN overlay network in order to improve observability when the control plane and data plane get out of sync.
  - Tested manually on a dev host.
  - Additional Hydra tests to automatically verify check script functionality.

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -784,11 +784,31 @@ in
         in
           "/run/wrappers/bin/sudo ${pkgs.fc.check-link-redundancy}/bin/check_link_redundancy ${links}";
       };
+      rib_integrity_ipv4 = {
+        notification = "Kernel network state has problems with underlay network routes";
+        interval = 300;
+        command = let
+          args = lib.concatMapStringsSep " " (p: "-p " + p) fclib.underlay.subnets;
+        in
+          "sudo -g frrvty ${pkgs.fc.check-rib-integrity}/bin/check_rib_integrity check-unicast-rib ${args}";
+      };
+      rib_integrity_evpn = {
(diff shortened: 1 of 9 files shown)


ctheune (approved):
LGTM

PR #1094, opened 2024-08-29
varnish: handle change in default stateDir

laalsaas:
PL-132901
## Release process

Impact:

Changelog:
The default value for stateDir has changed upstream, this migration
falls back to the old default in the reload script. This can be reverted
as soon as all VMs have been rebooted.

- [x] PR has internal ticket

## Design notes

## Security implications

  - varnish shouldn't crash due to performance issues, this is why the stateDir is moved in the first place
  - however, running instances of varnish should continue to use the old stateDir until the next reboot/service restart
  - tested changing the stateDir on pntest00, reloading went successfully

### DIFF
--- a/nixos/services/varnish/default.nix
+++ b/nixos/services/varnish/default.nix
@@ -126,7 +126,12 @@ in {
       reloadIfChanged = true;
       restartTriggers = [ cfg.extraCommandLine vcfg.package cfg.http_address ];
       reload = ''
-        vadm="${vcfg.package}/bin/varnishadm -n ${vcfg.stateDir}"
+        if [ -d "${vcfg.stateDir}" ]; then
+          statedir="${vcfg.stateDir}"
+        else
+          statedir="/var/spool/varnish/${config.networking.hostName}" #temporary migration
+        fi
+        vadm="${vcfg.package}/bin/varnishadm -n $statedir"
         cat ${commandsfile} | $vadm
 
         coldvcls=$($vadm vcl.list | grep " cold " | ${pkgs.gawk}/bin/awk {'print $5'})


ctheune (changes requested):
We need to coordinate this with @PhilTaken and @Ma27 - we're touching varnish and especially the reload logic, from multiple pointsn ow.

Also, as the stateDir is configurable in the upstream module, I'd like to see this properly managed in upstream, that changes in the stateDir have a proper migration path.

laalsaas:
> We need to coordinate this with @PhilTaken and @Ma27 - we're touching varnish and especially the reload logic, from multiple points now.

sure, is there a feature branch to have a look at?

> Also, as the stateDir is configurable in the upstream module, I'd like to see this properly managed in upstream, that changes in the stateDir have a proper migration path.

I think this isn't so much of a problem upstream, since upstream doesn't have `reloadOnChange = true` for this unit it just restarts, which means no _reload_ logic is needed, the stateDir should be discarded on restart.

ctheune:
> I think this isn't so much of a problem upstream, since upstream doesn't have `reloadOnChange = true` for this unit it just restarts, which means no _reload_ logic is needed, the stateDir should be discarded on restart.

Ah, that's good to know. I wonder whether we want to port the reload logic upstream then?

leona-ya (approved):
To me this seems quite fine. This is a temporary migration that can be removed once all VMs had a reboot.

Upstreaming our reload script is a later, kinda unrelated task. Also nixpkgs currently doesn't need any stateDir handling based on e.g. state version as (as @laalsaas mentioned) it's deleted on every systemd service restart.

ctheune (approved):

PR #1095, opened 2024-09-02
[21.05] Monitoring and alerting for kernel neighbour cache state

sysvinit:
This change introduces a new script, which uses `lnstat(1)` from iproute2 to monitor the state of the kernel neighbour cache tables for both IPv4 and IPv6. The script provides metrics to Telegraf in order to monitor the neighbour table state over time, and also provides a Sensu check for detecting when the number of entries in the cache has exceeded a threshold, and for when the table has overflowed.

The monitoring and alerting is currently only enabled on routers. The Sensu check is intentionally configured to run every minute, instead of the usual every ten minutes, as problems with the neighbour cache filling up can cause significant problems very suddenly.

PL-132950

## Release process

Impact: internal.

Changelog: none.

- [x] PR has internal ticket

## Design notes

## Security implications

  - Overflowing neighbour caches can cause outages and denial of service, so we should proactively monitor for kernel states which may cause problems.
  - Telegraf metrics verified with curl on a test host.
  - Sensu monitoring verified on a test host with an active load test to simulate an overflowing neighbour cache.

### DIFF
--- a/pkgs/fc/neighbour-cache-monitor/neighbour-cache-monitor.py
+++ b/pkgs/fc/neighbour-cache-monitor/neighbour-cache-monitor.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import subprocess
+import sys
+
+CACHE_TYPES = {
+    "ipv4": "arp_cache",
+    "ipv6": "ndisc_cache",
+}
+
+
+def warning(msg):
+    print(f"WARN - {msg}")
+    sys.exit(1)
+
+
+def critical(msg):
+    print(f"CRITICAL - {msg}")
+    sys.exit(2)
+
+
+def ok(msg):
+    print(f"OK - {msg}")
+    sys.exit(0)
+
+
+def read_cache_stats(cache_type):
+    # XXX: lnstat unconditionally returns exit code 1, even on
+    # success.
+    proc = subprocess.run(
+        [
+            "lnstat",
+            "-f",
+            cache_type,
+            "-j",
+            "-c1",
+        ],
+        capture_output=True,
+    )
+    return json.loads(proc.stdout.decode("utf-8"))
+
+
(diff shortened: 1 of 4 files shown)


pkgs/fc/neighbour-cache-monitor/neighbour-cache-monitor.py
```diff
@@ -0,0 +1,158 @@
...
+        data = fh.read()
+
+        # only attempt to decode json if the file is non-empty. file
+        # is empty on first creation.
+        state = None
+        if len(data) > 0:
+            state = json.loads(data.decode("utf-8"))
+
+        fh.seek(0)
+        fh.truncate()
```
ctheune, line 89: This is an unsafe write: truncate and write can both end up with partial data in the file. This is in itself not a problem if we catch this when loading data. The two options are: make the write safe (write to a tmpfile and rename) or catch errors and start with an empty state. In this situation I'm recommending catching the error and starting fresh.

ctheune: Because this was ambiguous (and everything that can happen will happen): keep the optimistic write path(s) but respond to errors when parsing the json and just start fresh.

ctheune (changes requested):
LGTM, small change and then this can go out.

pkgs/fc/neighbour-cache-monitor/neighbour-cache-monitor.py
ctheune: That's one part that I hate with Python's json decoding: JSON already specifies UTF-8 but the json module operates on (unicode) strings so we have to handle parsing ourselves.

There is a tiny crash situation still left in here: the data might be an invalid UTF-8 byte sequence and that will fail with a different error that we won't recover from. A general except would have been fine here.

ctheune (approved):

PR #1097, opened 2024-09-04
[21.05] zebra restart tuning and keepalived integration

sysvinit:
This change raises the startup backoff limits for zebra. The system default restart settings allow zebra to restart up to five time in 60 seconds, however in some circumstances where zebra has crashed due to the system being overloaded, it may take longer for the conditions causing zebra to crash to clear, so this change raises the limit to 20 times in 120 seconds.

Additionally, when zebra crashes on the primary router in a location, this should ideally cause the router to demote itself and let the secondary take over. This change adds a new check script to the keepalived configuration in all locations to monitor the liveness of zebra.

PL-132950

## Release process

Impact: internal.

Changelog: none.

- [x] PR has internal ticket

## Design notes

## Security implications

  - Improve site reliability and service availability with better integration between FRR and keepalived.
  - Load test in DEV used to artificially cause zebra crashes and test keepalived demotion behaviour.


ctheune (changes requested):
Generally fine, little adjustment: add a 10 sec restartsec option (and increase the startlimitinterval accordingly to 320s)

ctheune (approved):

PR #1099, opened 2024-09-10
Add basic role and integration for Grafana Loki

sysvinit:
This change introduces minimal, bare-bones platform integration for Grafana Loki, intended to serve as a basis for further development.

This includes:
- A role which can be assigned to VMs, which installs and provisions an instance of Loki. Currently this is statically configured to use filesystem-backed storage.
- Automatic configuration of promtail, a log-shipping client for Loki, when a Loki server is discovered in the same resource group. Currently only a single Loki server is supported. promtail is configured to read the system journal in JSON format, as this reads annotations and metadata which would otherwise be lost if reading only the plain journal messages. The Loki query language LogQL supports decoding JSON log messages with the `json` filter.
- Initial integration with Grafana in the existing statshost role. Hosts with the statshost-master role for a per-RG Grafana instance which also have the loki role enabled will have a data source added to allow Grafana to read and query data from the local Loki instance.

## Release process

Impact: internal. 

Changelog: none. This role is still under development, and is not yet intended for public use.

- [x] PR has internal ticket

## Design notes

  - The log retention time is intentionally configurable, as this may differ between use cases and applications.
  - No documentation yet, this is a development preview which is subject to change.

## Security implications

This is a basic first implementation, we're reviewing and considering overall security measures in PL-133025.

At this point we're limiting retention to 30 days by default (overridable per loki instance). We are relying on the SRV firewall as usual and find unauthed/cleartext submission acceptable for now.

To reduce impact if attackers have access on a non-loki machine, we're limiting access to the loki API to selected endpoints.

  - Integrations tested incrementally in a development environment to ensure correct functioning.
  - [x] data separation is handled with regular dev/whq/rzob and also RG separation for customer involvement
  - [x] attack surface area minimisation: no auth used, but covered by RG firewalling, service only on SRV
  - [x] loki runs as separate user

### DIFF
--- a/nixos/platform/default.nix
+++ b/nixos/platform/default.nix
@@ -34,6 +34,7 @@ in {
     ./monitoring.nix
     ./network.nix
     ./packages.nix
+    ./promtail.nix
     ./shell.nix
     ./static.nix
     ./syslog.nix
--- a/nixos/platform/promtail.nix
+++ b/nixos/platform/promtail.nix
@@ -1,0 +1,51 @@
+{ lib, config, ... }:
+
+let
+  enc = config.flyingcircus.enc;
+  fclib = config.fclib;
+
+  # XXX support multiple loki servers. the upstream docs note: "It is
+  # generally recommended to run multiple Promtail clients in parallel
+  # if you want to send to multiple remote Loki instances."
+  lokiServer = fclib.findOneService "loki-collector";
+in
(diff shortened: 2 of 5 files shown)


ctheune (commented):
That looks like a great simple start.

ctheune (approved):

PR #1124, opened 2024-10-16
Trigger cold reboots upon changes in KVM environment parameters

sysvinit:
This change adds support to fc-agent for a new mechanism in fc.qemu which signals selected boot-time parameters from the KVM environment, similar to the existing qemu binary generation handling. The corresponding side in fc.qemu will seed the parameters once at boot and then periodically write them into /run at runtime through the qemu guest agent. If fc-agent detects changes, then it will schedule a cold reboot in a maintenance window so the KVM host can carry out changes which require a restart of the qemu process.

PL-131857

## Release process

Impact:

Changelog:
- VMs will now automatically schedule reboots during maintenance when changes in KVM parameters such as CPU flags or VM backing storage (i.e. SSD or HDD) are detected (PL-131857).

- [x] PR has internal ticket

## Design notes

## Security implications

  - This is an extension of the existing qemu binary generation upgrade mechanism, where the hypervisor signals the state of the guest-external KVM environment through the qemu guest agent.
  - Additional unit tests, in particular to test compatibility with the previous qemu binary generation handling mechanism, and the upgrade path from the existing code to the new code in fc.qemu.
  - Verified manually in DEV.

### DIFF
--- a/pkgs/fc/agent/fc/maintenance/maintenance.py
+++ b/pkgs/fc/agent/fc/maintenance/maintenance.py
@@ -5,6 +5,8 @@ system state. All maintenance activities start their life here in this module.
 TODO: better logging for created activites.
 
 """
+
+import json
 import os.path
 import shutil
 import typing
@@ -22,6 +24,10 @@ from fc.maintenance.activity.update import UpdateActivity
 from fc.maintenance.activity.vm_change import VMChangeActivity
 from fc.maintenance.state import State
 from fc.util import nixos
+
+QEMU_STATE_DIR = "/var/lib/qemu"
+KVM_SEED_DIR = "/tmp/fc-data"
+RUNTIME_DIR = "/run"
 
 
 def request_reboot_for_memory(log, enc) -> Optional[Request]:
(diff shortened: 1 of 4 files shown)


pkgs/fc/agent/fc/maintenance/maintenance.py
```diff
@@ -23,6 +24,10 @@
 from fc.maintenance.state import State
 from fc.util import nixos
 
+# If the value of these parameters in the ENC data change, then a cold
+# reboot should be scheduled
+ENC_PARAMETERS_FOR_COLD_REBOOT = ["cpu_model", "rbd_pool"]
```
ctheune: As discussed: maintaining the parameters on the agent running within the VM is prone to us not being able to manage infrastructure aspects sufficiently independent from the guest agent. Meaning: if we decide that further parameters need to be able to trigger cold reboots, then we need to release updates to older platforms. Please let the inner agent just use whatever data is passed into the guest-properties file to make a decision for a cold boot.

pkgs/fc/agent/fc/maintenance/maintenance.py
```diff
@@ -56,53 +61,107 @@ def request_reboot_for_cpu(log, enc) -> Optional[Request]:
...
+    ):
+        msg = "Cold restart because the Qemu binary environment has changed"
+        return Request(RebootActivity("poweroff"), comment=msg)
+
+    # Schedule maintenance if ENC parameteres related to the KVM
+    # environment have changed.
+    if booted_properties:
+        current_properties = enc["parameters"]
+        changed_properties = []
+        for key in ENC_PARAMETERS_FOR_COLD_REBOOT:
```
ctheune: - newly added keys should likely trigger a reboot
- keys that go away maybe should not trigger a reboot?
- when making changes from the earlier comment this needs to exclude the binary generation OR decide that binary generation can also be downgraded and make the old/new style paths more separate (and then we can retire the old path in 24.11 or so)
- i think it's fine if we schedule gratuitous reboots to introduce this feature

ctheune (changes requested):

sysvinit:
After further discussion we've decided to revise some of the logic in here, and instead of using ENC as the source of the desired state of the world, we should use an extra properties file fed into the VM by the qemu guest agent from the KVM layer. This means we're no longer limited to only comparing values inside ENC, and means we can detect new guest property values being added by the hypervisor at runtime which were not present at boot time.

pkgs/fc/agent/fc/maintenance/maintenance.py
sysvinit: Fixed in [25e49b3](https://github.com/flyingcircusio/fc-nixos/pull/1124/commits/25e49b35f3b49b6e605f9fe2fcd7e8e3cf375a44)

pkgs/fc/agent/fc/maintenance/maintenance.py
sysvinit: Addressed in [25e49b3](https://github.com/flyingcircusio/fc-nixos/pull/1124/commits/25e49b35f3b49b6e605f9fe2fcd7e8e3cf375a44).

ctheune (approved):

PR #1165, opened 2024-11-13
[FC-41616] overlay: add patchelf-venv

Ma27:
This patch adds a `--add-rpath-and-shrink` option which performs `--add-rpath` and `--shrink-rpath` in a single write to not leak space (`--add-rpath` increases the ELF header size, `--shrink-rpath` doesn't reduce it when necessary).

See also https://github.com/flyingcircusio/batou_ext/pull/199 & FC-41616.

__Note:__ this is a draft until I know whether this has a chance of getting upstreamed. Would still be interested in your opinion on the change itself and the patch in https://github.com/flyingcircusio/patchelf/commit/6ffde887d77275323c81c2e091891251b021abb3.patch

## Release process

- [ ] PR has internal ticket

## Design notes

## Security implications

### DIFF
--- a/changelog.d/20241113_135856_FC-41616-patched-patchelf_scriv.md
+++ b/changelog.d/20241113_135856_FC-41616-patched-patchelf_scriv.md
@@ -1,0 +1,22 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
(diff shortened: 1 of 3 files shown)


Ma27:
OK no idea how to delete the `just a test` thingy here, I tried to submit a request against GitHub to make sure the token is still correct while investigating PL-133144.

Ma27:
I guess I'll close this.
We use the override in-place in batou_ext on numerous places and the package is so small that I don't recall anybody complaining. The correct call is to upstream this anyways. Either you get to it or I'll try to find some free-time. Let's see...

PR #1169, opened 2024-11-17
fix postgres tests and remove eol postgresql12

laalsaas:
Since postgresql sets ProtectSystem=strict, all directories must exist before the service is started. Also all directories where write access is needed must be specified.

PL-133170

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - we need the postgresql test to pass. 
  - we have to run supported software versions
  - the postgresql test passes now 
  - postgresql12 is now removed as it's eol


nixos/services/postgresql/default.nix
```diff
@@ -167,6 +167,7 @@ in {
 
       systemd.services.postgresql.serviceConfig = {
         Restart = "always";
+        ReadWritePaths = [ "/nix/var/nix/gcroots/per-user/postgres" ];
```
laalsaas, line 170: to me, this raises the question, why exactly we are setting gcroots „manually“ and if we _really_ need to do that. maybe @dpausp can provide some insight here.

osnyx:
You forgot to remove the postgres12-related entries from important packages as well, I just added that to mitigate eval errors in a more recent nixpkgs.

nixos/services/postgresql/default.nix
osnyx: This is done to keep around old postgresql versions during postgres updates. At start, each postgres verrsion registers itself in the postStart phase as a gcroot. Our fc-postgresql tooling cleans these gcroots after a successful update.

osnyx (approved):

PR #1175, opened 2024-11-25
[21.05] Replace dhcpd from isc-dhcp with Kea

sysvinit:
Forward porting hardware support to 24.11 is currently blocked on isc-dhcp, which is end-of-life and has been removed from Nixpkgs in more recent versions. This change replaces dhcpd with Kea, the designated successor project, which is supported by Nixpkgs upstream.

This includes fc-agent support for generating static lease configuration based on data from the directory. This integration has a few correctness fixes in comparison with the isc-dhcp integration: the NTP server option advertised in DHCP is now set to the subnet router address to match the NixOS host configuration, twinned networks are handled properly to avoid multiple definitions of overlapping subnets, and the hostnames advertised in DHCPv6 are now fully-qualified domain names, as DHCPv6 does not have separate options for hostname and domain name.

PL-133205

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - The new DHCP server implementation should match the behaviour of the old DHCP server implementation.
  - Kea is actively maintained by upstream and has security support.
  - Tested on a router in DEV. Manually verified that both BIOS and UEFI clients can properly load iPXE, and that the iPXE loader can properly load the boot menu script.

### DIFF
--- a/changelog.d/20241126_174743_PL-133205-dhcp-port-to-kea_scriv.md
+++ b/changelog.d/20241126_174743_PL-133205-dhcp-port-to-kea_scriv.md
@@ -1,0 +1,20 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+
+### NixOS XX.XX platform
+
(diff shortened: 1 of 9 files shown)


osnyx (approved):
I wonder whether the `fc-dhcpd` command is still required. Still merging this, if we can get rid of it feel free to provide a follow-up cleanup PR.

ctheune (commented):
late to the party, but just went through it with molly to get a rough understanding what's going on.

PR #1191, opened 2024-12-09
[FC-42163] fc-postgresql: support upgrades with custom extensions

Ma27:
Upgrading postgresql to a new major with fc-postgresql fails when using custom extensions, defined via `services.postgresql.extraPlugins`.

This patch modifies the relevant agent code to use a postgresql package WITH all needed plugins for upgrades:

* for automatic upgrades, these extensions are discovered automatically.
* for manual upgrades, the user must specify those manually with `--extension-names`.

The original use-case of a customer was to upgrade postgresql to a newer version with `postgresqlPackages.anonymizer` installed (FC-42089). This is a little trickier since it requires `anon` to be listed in `shared_preload_libraries`. For this, all `shared_preload_libraries` are now extracted from `postgresql.conf` and passed to `pg_upgrade` via CLI. This code is based on the assumption that the old `postgresql.conf` is generated by Nix and thus ignores all edge-cases, the postgresql config format may have.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Be able to upgrade postgresql, even with custom extensions.
  - Anonymization (`pg_anon` is used) is kept during upgrades.
  - Implemented an integration test for both automatic and manual upgrades, based on the original one.
  - Verified a manual upgrade (12 -> 14 with inactive anonymization, 14 -> 15 with active anonymization) on test68.

### DIFF
--- a/nixos/services/postgresql/default.nix
+++ b/nixos/services/postgresql/default.nix
@@ -64,6 +64,16 @@ let
   collationVerifierScript = pkgs.writers.writePython3Bin "verify-collations"
       {} (builtins.readFile ./verify-collations.py);
 
+  mkExtensionNamesForAutoUpgrade = postgresqlMajor:
+    let
+      availableExtensions = lib.mapAttrs'
+        (attrName: package: lib.nameValuePair (lib.getName package) attrName)
+        (builtins.removeAttrs pkgs."postgresql${toString postgresqlMajor}Packages" [ "recurseForDerivations" ]);
+      pluginByPname = package: lib.optionals (availableExtensions?${lib.getName package}) [
+        availableExtensions.${lib.getName package}
+      ];
+    in
+    lib.concatMap pluginByPname (config.services.postgresql.extraPlugins config.services.postgresql.package.pkgs);
 in {
   options = with lib; {
 
@@ -159,7 +169,7 @@ in {
       '';
 
       systemd.services.postgresql.postStart = ''
-        ln -sfT ${postgresqlPkg} ${upstreamCfg.dataDir}/package
+        ln -sfT ${config.services.postgresql.package.withPackages config.services.postgresql.extraPlugins} ${upstreamCfg.dataDir}/package
         ln -sfT ${upstreamCfg.dataDir}/package /nix/var/nix/gcroots/per-user/postgres/package_${cfg.majorVersion}
       '' + (lib.optionalString (lib.versionAtLeast cfg.majorVersion "15") ''
         ${collationVerifierScript}/bin/verify-collations ${upstreamCfg.dataDir}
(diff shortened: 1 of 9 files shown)


nixos/services/postgresql/default.nix
```diff
@@ -64,6 +64,16 @@ let
...
+  mkExtensionNamesForAutoUpgrade = postgresqlMajor:
+    let
+      availableExtensions = lib.mapAttrs'
+        (attrName: package: lib.nameValuePair (lib.getName package) attrName)
+        (builtins.removeAttrs pkgs."postgresql${toString postgresqlMajor}Packages" [ "recurseForDerivations" ]);
+      pluginByPname = package: lib.optionals (availableExtensions?${lib.getName package}) [
+        availableExtensions.${lib.getName package}
+      ];
+    in
+    lib.concatMap pluginByPname (config.services.postgresql.extraPlugins config.services.postgresql.package.pkgs);
```
osnyx, line 76:
```nix
      extraPlugins = mkOption {
        type = with types; coercedTo (listOf path) (path: _ignorePg: path) (functionTo (listOf path));
        default = _: [];
```

`config.services.postgresql.extraPlugins` can be either a function or a list, according to the nixpkgs definition. It looks like the code above can only handle the *function* case.

tests/postgresql/upgrade-with-extension.nix
```diff
@@ -0,0 +1,280 @@
...
+            (testlib.fcConfig { net.fe = false; })
+          ];
+
+          flyingcircus.roles.postgresql12.enable = lib.mkDefault true;
+          services.postgresql = {
+            extraPlugins = ps: with ps; [ anonymizer ];
+            settings.shared_preload_libraries = lib.mkForce "auto_explain, pg_stat_statements, anon";
+          };
+
+          specialisation = {
```
osnyx, line 95: This boilerplate of postgres version definitions is mostly duplicated from the main postgresql update tests. The last times the postgres versions have been adapted there, this turned out to be a brittle and annoying manual task – whis is now also necessary twice.
Hence, refactoring the tests to be modular and share some code should be the goal. This can be postponed though until when the autoupgrade tests are generally refactored.

nixos/services/postgresql/default.nix
Ma27: It's always a function: the second argument of `coercedTo` transform the list (first type) into a function (second type).

tests/postgresql/upgrade-with-extension.nix
Ma27: > This can be postponed though until when the autoupgrade tests are generally refactored.

Does this mean there was work on this already?
But yeah, I'd prefer to postpone this for now. I considered using the upgrade test for that, but splitting it off made it easier to iterate than having to invent a structure to embed these testcases into the existing ones on the fly.

Regarding modular tests: just in case you haven't seen it already, tests are also just modules which makes it relatively straightforward to write test helpers for a class of tests. See e.g. https://github.com/NixOS/nixpkgs/blob/master/nixos/tests/nextcloud/default.nix.

osnyx: No, there was no proper work on refactoring so far, only on manually updating the versions in the test before. And that was when it became apparent that this just incurs a lot of manual labour.

osnyx (approved):

PR #1192, opened 2024-12-09
[PL-133251] fix varnish reload

PhilTaken:
varnish reload was broken before when there was more than 0 cold vcls since the $vadm variable is empty. changing it to ${vadm} properly templates the path to the varnish admin binary using nix.

this does not immediately break varnish but leaves cold vcls lying around. a restart will remove all cold vcls anyways and a consecutive reload would also mark the system as failed but load new vcls as intended besides discarding cold vcls

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - fixes reloads when there is > 0 cold vcls
  - tested on test vm

### DIFF
--- a/changelog.d/20241209_153315_phil-fix-varnish-reload_scriv.md
+++ b/changelog.d/20241209_153315_phil-fix-varnish-reload_scriv.md
@@ -1,0 +1,15 @@
+<!--
+A new changelog entry.
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+-->
+
+### Impact
+
+
+### NixOS XX.XX platform
+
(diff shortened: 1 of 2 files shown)


ctheune (changes requested):
Please improve the changelog.

Also: how did this happen and wasn't caught during development?

PhilTaken:
This is an edge case that only occurs when there are cold vcls to discard. This happened in 9f10ecc9a888f603bfd3df914466fd5d244d207f when vadm was changed from a shell variable to a nix variable for better reuse in another script. Since bash doesn't inform about undefined variables and only substitutes them with an empty string this was not caught immediately

ctheune:
As discussed yesterday: please try to add a test case for this - it's been broken once already and caused issues.

Also: the branch name doesn't include the ticket number and the changelog entry doesn't either.

ctheune (changes requested):
see my other comment: branch name, ticket number  in changelog, and test case

PR #1200, opened 2024-12-11
[23.11] fix changelog.sh

dhnasa:
PL-133226

## Release process

- n/a

- [x] PR has internal ticket

## Design notes

  - n/a
  - n/a

## Security implications

  - bugfix
  - manually

### DIFF
--- a/changelog.d/scriv.ini
+++ b/changelog.d/scriv.ini
@@ -1,6 +1,6 @@
 [scriv]
 format = md
 entry_title_template =
-categories = Impact, NixOS platform
+categories = Impact, NixOS XX.XX platform
 output_file = changelog.d/CHANGELOG.md.tmp
 skip_fragments = CHANGELOG.*
--- a/changelog.sh
+++ b/changelog.sh
@@ -2,8 +2,20 @@
 
 set -e
 
-base="./changelog.d"
+if [ -z "$EDITOR" ]; then
+  echo "Error: \$EDITOR is not set or empty"
+  exit 1
+fi
 
-new_item="${base}/$(date '+%Y%m%d_%H%M%S')_$(git rev-parse --abbrev-ref HEAD)_scriv.md"
-cp "${base}/new_fragment.md.j2" ${new_item}
-$EDITOR $new_item
+base="./changelog.d"
+branch=$(git rev-parse --abbrev-ref HEAD)
+
+new_item="${base}/$(date '+%Y%m%d_%H%M%S')_${branch/\//-}_scriv.md"
+template="${base}/new_fragment.md.j2"
+
+cp "$template" "$new_item"
+
+if ! $EDITOR "$new_item" || diff -q "$template" "$new_item" > /dev/null; then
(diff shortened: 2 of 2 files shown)


changelog.d/20241211_144107_PL-133226-fix-changelog.sh_scriv.md
```diff
@@ -0,0 +1,16 @@
...
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### NixOS XX.XX platform
+
+- Internal: Fix changelog.sh script (PL-133226)
```
ctheune: This doesn't require a changelog entry. Also, the changelog is a typical non-helpful entry, which is also an indicator of: just don't do it, as it's basically noise for the reader.

changelog.d/scriv.ini
```diff
@@ -1,6 +1,6 @@
 [scriv]
 format = md
 entry_title_template =
-categories = Impact, NixOS platform
+categories = Impact, NixOS XX.XX platform
```
ctheune, line 4: I guess this was a typo previously?

changelog.sh
```diff
@@ -6,4 +6,9 @@ base="./changelog.d"
 
 new_item="${base}/$(date '+%Y%m%d_%H%M%S')_$(git rev-parse --abbrev-ref HEAD)_scriv.md"
 cp "${base}/new_fragment.md.j2" ${new_item}
-$EDITOR $new_item
+if [ -z "$EDITOR" ]; then
+  echo "\$EDITOR is not set or empty"
+  echo "edit $new_item manually"
```
ctheune: Ah, no, I'd like this to be the other way around. Ask people to set their editor or refuse to do work. I don't want people to implement different workflows here.

ctheune (changes requested):

changelog.sh
sysvinit: It would be nice if the script could check the return status of the editor process and discard the changelog if it returned non-zero. This would be more consistent with other tools that spawn editors like git (e.g. when spawning an editor to write a commit message, git will abort creating the commit if the commit message file is unchanged or if the editor exits non-zero).

ctheune (changes requested):
See comments.

changelog.d/20241211_144107_PL-133226-fix-changelog.sh_scriv.md
dhnasa: done

changelog.d/scriv.ini
dhnasa: Forgot to update this when updating the template headers. (Shouldn't have made a difference as unknown categories are ordered last)

changelog.sh
dhnasa: done

ctheune (approved):

PR #1201, opened 2024-12-12
rabbitmq: add sensu check

laalsaas:

PL-132311
## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - admins should know which feature flags to enable _before_ they upgrade again, when required feature flags aren't enabled, updates can lead to a breakage.
  - deployed the check to pntest00 with rabbitmq without all stable feature flags enabled -> check failed
  - enabled all feature stable flags -> check succeeded.

### DIFF
--- a/nixos/services/rabbitmq/default.nix
+++ b/nixos/services/rabbitmq/default.nix
@@ -283,20 +283,31 @@ in {
 
     flyingcircus.services = {
 
-      sensu-client.checks.rabbitmq-alive = {
-        notification = "rabbitmq amqp alive";
-        command = ''
-          ${pkgs.sensu-plugins-rabbitmq}/bin/check-rabbitmq-amqp-alive.rb \
-            -u fc-sensu -w ${cfg.listenAddress} -p ${sensuPassword}
-        '';
-      };
+      sensu-client.checks = {
+        rabbitmq-alive = {
+          notification = "rabbitmq amqp alive";
+          command = ''
+            ${pkgs.sensu-plugins-rabbitmq}/bin/check-rabbitmq-amqp-alive.rb \
+              -u fc-sensu -w ${cfg.listenAddress} -p ${sensuPassword}
+          '';
+        };
 
-      sensu-client.checks.rabbitmq-node-health = {
-        notification = "rabbitmq node healthy";
-        command = ''
-          ${pkgs.sensu-plugins-rabbitmq}/bin/check-rabbitmq-node-health.rb \
-            -u fc-sensu -w ${cfg.listenAddress} -p ${sensuPassword}
-        '';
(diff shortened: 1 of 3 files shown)


nixos/services/rabbitmq/default.nix
```diff
@@ -283,20 +283,31 @@ in {
...
+          command = ''
+            ${pkgs.sensu-plugins-rabbitmq}/bin/check-rabbitmq-node-health.rb \
+              -u fc-sensu -w ${cfg.listenAddress} -p ${sensuPassword}
+          '';
+        };
+
+        rabbitmq-feature-flags-enabled = {
+          notification = "rabbitmq all stable feature flags enabled";
+          command = ''
+            ! /run/wrappers/bin/sudo -u rabbitmq ${pkgs.rabbitmq-server}/bin/rabbitmqctl list_feature_flags state stability |
```
osnyx: It's a bit a matter of taste. but I'd like to avoid hard-coded absolute paths when introducing new things.
`sudo` is in PATH in the check env.

osnyx (changes requested):
As long as it is not too complex, please add a corresponding test in our plattform rabbitmq tests. The other sensu checks are tested there as well.
We'd like to see changes in the output we're parsing here.

laalsaas:
I modified the test to make sure the check succeeds. I don't test for the check failing when rabbitmq is down, since this isn't a check for the "immediate status" of rabbitmq.

When rabbitmq is down, the rabbitmqctl command will fail. however, since the exit code of the entire statement is negated, the entire check will succeed. I don't think it is strictly neccessary, that the check fails, as explained above.

tests/rabbitmq.nix
```diff
@@ -31,6 +31,7 @@ in {
     sensuOpts = "-u fc-sensu -w ${ipv4} -p ${testlib.derivePasswordForHost "sensu"}";
     amqpAliveCheck = "${pkgs.sensu-plugins-rabbitmq}/bin/check-rabbitmq-amqp-alive.rb ${sensuOpts}";
     nodeHealthCheck = "${pkgs.sensu-plugins-rabbitmq}/bin/check-rabbitmq-node-health.rb ${sensuOpts}";
+    featureFlagCheck = "! sudo -u rabbitmq ${pkgs.rabbitmq-server}/bin/rabbitmqctl list_feature_flags state stability |${pkgs.gnugrep}/bin/grep stable | ${pkgs.gnugrep}/bin/grep  disabled";
```
osnyx, line 34: This is basically code duplication, prone to diverting between test and actual check. We can instead pull in the actual code via the NixOS module system. I'll do that in a follow-up commit.

PR #1240, opened 2025-01-16
fc-slurm check: don't crash on non-"slurm-node"s

osnyx:
When executing `fc-slurm check` on a host that does not have the `slurm-node` or `slurm-controller` role, but is e.g. a `slurm-external-dependency`, the `fc-slurm check` command used to crash due to no available node data.

This implements a fallback warning message to better understand the situation and, if this was about to happen on an actual slurm cluster node, better debug the available cluster information.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - user-facing commands shall not crash on common szenarios with invalid data
  - provide helpful debugging information in case of unexpected data
  - increase test coverage for newly-added functionality
  - [x] new code covered by a unit test
  - [x] NixOS tests still pass

### DIFF
--- a/pkgs/fc/agent/fc/util/slurm.py
+++ b/pkgs/fc/agent/fc/util/slurm.py
@@ -4,6 +4,7 @@ import time
 from collections import Counter
 from enum import Enum
 from functools import reduce
+from itertools import chain
 from typing import NamedTuple, Optional
 
 import pyslurm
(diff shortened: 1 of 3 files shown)


pkgs/fc/agent/fc/util/slurm.py
```diff
@@ -733,16 +734,25 @@ def check(log, hostname) -> CheckResult:
...
 
-    if hostname in pyslurm.node().get():
+    if hostname in slurm_nodes:
         results.append(check_node(log, hostname))
 
-    return reduce(CheckResult.merge, results)
+    if not results:
+        return CheckResult(
+            warnings=[
+                f"No data available for this node `{hostname}`. Is this a `slurm-node`?\n"
```
ctheune, line 760: The question is unclear. Does this indicate it should run on slurm-nodes or should not run on slurm nodes?

On a more general node: why would this be executed or be installed on non-slurm-nodes?

ctheune (changes requested):

pkgs/fc/agent/fc/util/slurm.py
osnyx: `fc-slurm` is in-path on hosts with the role `slurm-external-dependency`, because they are also managed by our platform slurm (global) maintenance logic. But theses nodes are neither a slurm agent nor a controller, hence the actual slurm cluster has no metadata about this host.

On the affected host, `fc-slurm check` ist **not** automatically invoked by a sensu check, so in the normal case you need to manually invoke that command to see it – as of now – crash. Nonetheless I decided to adopt this warning style that lists which node data would have been available, because that might be helpful if the check also fails on a proper slurm node due to incorrect data being returned from the slurm controller. We can then spot in the warning what data is missing – instead of just observing a crash.

ctheune: Ack. Please add that as a comment there.

ctheune (changes requested):

ctheune (approved):

PR #1242, opened 2025-01-17
rabbitmq: auto-enable feature flags after upgrade

laalsaas:
Autmatically enable all feature flags after an upgrade for single-node rabbitmq clusters.

PL-132311

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - in constellation where this is possible (single-node clusters), all available feature flags should be enabled automatically.
  - tested manually on pntest00
  - wrote and ran tests

### DIFF
--- a/tests/rabbitmq.nix
+++ b/tests/rabbitmq.nix
@@ -1,29 +1,44 @@
 import ./make-test-python.nix ({ lib, pkgs, testlib, ... }:
 let
   # Default IP automatically assigned in NixOS tests. Must not be changed here.
-  ipv4 = "192.168.1.1";
+  ipv4 = "192.168.3.1";
+  nodeDefaults = {
+    flyingcircus.roles.rabbitmq.enable = true;
+
+    flyingcircus.encServices = [
+      {
+        address = "machine.gocept.net";
+        location = "test";
+        password = "baz";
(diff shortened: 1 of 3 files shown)


tests/rabbitmq.nix
```diff
@@ -1,29 +1,37 @@
 import ./make-test-python.nix ({ lib, pkgs, testlib, ... }:
 let
   # Default IP automatically assigned in NixOS tests. Must not be changed here.
-  ipv4 = "192.168.1.1";
+  ipv4 = "192.168.3.1";
+  nodeDefaults = {
+    imports = [ (testlib.fcConfig { }) /*../nixos ../nixos/roles*/ ];
```
osnyx: `testlib.fcConfig` takes an optional `id` argument, defaulting to 1. This is used to derive node's IP addresses, MACs and other network-related config.

The current re-usage of `nodeDefaults` causes both nodes to have equal addresses. This currently happens to work, as we do not test any network interactions between individual nodes, but nonetheless isn't really clean and *might* cause issues at some point. => I prefer to have different IDs per node.

osnyx (changes requested):

osnyx (approved):
I took the liberty of adjusting the test myself.

PR #1245, opened 2025-01-22
[24.11] Fix IPv6 autoconfiguration

sysvinit:
Due to refactoring and merges, the network code in the platform did not correctly ensure that IPv6 autoconfiguration is disabled on all interfaces by default. This change fixes this regression, and introduces some tests to ensure that we catch unintentional changes to this behaviour again in the future.

Additionally, I've adjusted the network configuration code to allow the VM or hardware configuration profile to be selectable using a NixOS option. This allows the hardware profile to be used in Hydra test VMs so we can test EVPN-related network functions without having to rewrite all the config manually by hand, and I've also included a test fixture which will create a VM acting as a mock datacentre VXLAN switch.

PL-133360

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - IPv6 autoconfiguration should be disabled by default: this ensures that hosts will use only IP addresses explicitly assigned to them in the directory, and prevents traffic being stolen by rogue hosts sending router advertisements.
  - New Hydra tests to check that the desired configuration is really applied
  - Manually verified on both hardware and in VMs.

### DIFF
--- a/tests/ipv6-autoconfig.nix
+++ b/tests/ipv6-autoconfig.nix
@@ -1,0 +1,114 @@
+import ./make-test-python.nix ({ pkgs, testlib, ... }:
+let
+  assertZeroSysctl = pkgs.writeScriptBin "assert_zero_sysctl" ''
+    set -eu
+    sysctl="$1"
+
+    value="$(${pkgs.procps}/bin/sysctl -n -b "$sysctl")"
+
+    [ "$value" == "0" ]
+  '';
+
+  makePhysicalHost = { id, links }: { lib, config, ... }:
+    let
+      testNodeId = config.virtualisation.test.nodeNumber;
+    in {
+      imports = [
+        (testlib.fcConfig {
+          inherit id;
+          net.mgm = true;
+          net.ul = true;
+          extraEncParameters = {
+            inherit id;
+            interfaces.fe.policy = "vxlan";
+            interfaces.srv.policy = "vxlan";
+            interfaces.ul = {
+              policy = "underlay";
+              nics = map (link: {
+                mac = "52:54:00:12:${lib.toLower (lib.toHexString link)}:0${toString testNodeId}";
+                external_label = "phys/${toString link}";
+              }) links;
+            };
+          };
+        })
+      ];
+
+      # use the hardware networking config profile
+      flyingcircus.networking.physicalHostNetworking = true;
+      # extra underlay network links
+      virtualisation.vlans = links;
+
+      services.fail2ban.enable = false;
+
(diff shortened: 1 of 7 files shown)


tests/ipv6-autoconfig.nix
```diff
@@ -0,0 +1,101 @@
...
+            "temp_valid_lft",
+            "temp_prefered_lft",
+            "addr_gen_mode",
+        ]
+
+        machine.wait_for_unit("multi-user.target")
+
+        with subtest("testing ipv6 autoconf configuration on ethsrv"):
+            for sysctl in sysctls:
+                machine.succeed(f"sysctl net.ipv6.conf.ethsrv.{sysctl} | grep 0")
```
ctheune: I'm a bit frustrated by the switch from Python to Bash ... ;) Specifically the grep feels like a shotgun.

If we need to stay within bash then this would be a bit more exact: `[ "$(sysctl  net.ipv6.conf.ethsrv.accept_ra -n -b)" == "1" ]` ... BUT ... it's terrible to read IMHO ... 

I'll let you decide.

ctheune (approved):
Looks good to me - I'm trusting the fix (I'm a bit hazy of what exactly was wrong here) as you mentioned that this has happened before, fixed up the learnings from the various incarnations and wrote tests.

tests/ipv6-autoconfig.nix
sysvinit: On second thought, the sysctls which handle temporary SLAAC-generated addresses take times in seconds, so this would match if there's a 0 digit in the number of seconds. I think a better solution might be to write a script which asserts that the sysctl is 0, and then invoke that in the test.

sysvinit:
> Looks good to me - I'm trusting the fix (I'm a bit hazy of what exactly was wrong here)

In the implementation of the network code we have a distinction between interfaces which are backed by some hardware network device, and those which are backed by a synthetic VXLAN kernel network device. The existing code only generates the systemd units which disable IPv6 autoconfig for the VXLAN-backed "virtual" links. This means that the physical links are left with the default behaviour (which is to enable IPv6 autoconfig, accept router advertisements, etc).

https://github.com/flyingcircusio/fc-nixos/blob/8e6c7d2d430bd4cbd88276703d61d2f088aa83ce/nixos/platform/network.nix#L551-L555

This is missing the creation of the systemd units also for `ethernetLinks`. This used to be there back when we were developing VXLAN support, however it seems that it got erroneously removed at some point during a refactoring.

ctheune (approved):

PR #1246, opened 2025-01-22
Initial platform support for layer-3 routed networks (L3-VNI)

sysvinit:
This change adds initial support for using EVPN-VXLAN as a transport for layer 3 routed-to-the-host networking. Networks which have the `routed` flag set in the ENC data will have an additional VRF device will be created which will be configured as the parent interface for other devices associated with that network (bridge device, VXLAN device).

PL-133324

## Release process

- [x] ~~Created changelog entry using `./changelog.sh`~~ => internal change.

- [x] PR has internal ticket

## Design notes

## Security implications

  - This is part of a new feature which is still currently in development. However, this is gated behind an ENC flag per network which is currently not set by the directory, and which otherwise defaults to a no-op relative to existing behaviour.
  - Tested on various hardware hosts in DEV.

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -7,21 +7,29 @@ let
 
   fclib = config.fclib;
 
-  # XXX there's a lot of conditionals here
-  interfaces = filter
-    (i: i.vlan != "ipmi" && i.vlan != "lo")
+  # Reminder:
+  # * interfaces are those things that we configure IP addresses on.
+  # * links are things that carry MAC addresses and provide (physical) transport to a network
+  # see lib/network.nix for the full nomenclature.
+
+  # Some interfaces aren't managed through this code. For example: IPMI
+  # is configured using the ipmitool elsewhere.
+  managedInterfaces = filter
+    (i: ! (elem i.policy [ "unmanaged" "null" ]) &&
+        ! (elem i.vlan ["ipmi" "lo"]))
     (lib.attrValues fclib.network);
 
-  managedInterfaces = filter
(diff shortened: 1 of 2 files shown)


nixos/platform/network.nix
```diff
@@ -81,7 +85,15 @@ let
...
 
-   quoteLabel = replaceStrings ["/"] ["-"];
+  quoteLabel = replaceStrings ["/"] ["-"];
+
+  vrfName = iface: "vrf${iface.vlan}";
+  vrfTable = iface: let
+    # the routing tables 0, 253, 254, and 255 are reserved by the
+    # kernel. this base number is chosen as an arbitrary offset for
+    # numbering kernel routing tables for vrfs
+    tableBase = 1000;
```
ctheune: HMMM! This is interesting, we should check back with our vlan/vxlan/... numbering scheme and whether that can fit without shifting the values ...?

ctheune (changes requested):
We should review this together - I can't quite follow all the code and have the impression that we might want to pay attention to comments and factoring to keep it readable/understandable for the future us's.

nixos/platform/network.nix
sysvinit: That's not a bad idea actually. We definitely have space in the numbering scheme as it stands at present, so we could reserve those four IDs and then map the virtual network ID one-to-one to kernel tables.

ctheune: agreed. add assertions and update the docs.

ctheune (approved):

PR #1255, opened 2025-01-29
[24.11] Remove fc-kea from activation scripts

sysvinit:
As a workaround for problems when upgrading from 21.05, extra calls to the fc-kea script which generates parts of the Kea DHCP configuration were added to the activation scripts. However, the activation scripts may be executed without network connectivity (especially when performing large system updates which restart lots of network-related systemd units) -- this means that fc-kea will fail when attempting to connect to the directory, which will cause the system activation to fail. This is a failure mode observed when attempting to upgrade routers from 21.05 to 24.11 automatically, as due to changes in FRR the EVPN control plane will be stopped completely before the activation scripts are run.

This change removes fc-kea from the activation scripts.

PL-133387

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Reliability and robustness of upgrade paths.
  - Tested successfully on the staging router which had become stuck after failing to activate 24.11 for the first time.

### DIFF
--- a/nixos/roles/router/kea.nix
+++ b/nixos/roles/router/kea.nix
@@ -93,9 +93,6 @@ let
     ];
   };
 
-  # The script is embedded in two locations: before the start of the
-  # kea server, to ensure that we have a known-good configuration and in
-  # the agent to update it continuously from the directory data.
   configUpdateScript = pkgs.writeShellScript "update-kea-config" ''
       # Generate DHCP server configuration
       ${pkgs.fc.agent}/bin/fc-kea -4 -L ${location} -o /etc/nixos/localconfig-dhcpd4.json -e ipmi ${lib.concatMapStringsSep " " (net: "-n ${net}") dhcpNetworks}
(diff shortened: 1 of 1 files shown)


osnyx:
So when will the old 21.05 kea config be replaced by the new 24.11 kea config? Either when the periodic fc-agent.service is run, or also when activating the primary specialisation (as kea DHCP is only needed on the primary router)?

sysvinit:
> So when will the old 21.05 kea config be replaced by the new 24.11 kea config? Either when the periodic fc-agent.service is run, or also when activating the primary specialisation (as kea DHCP is only needed on the primary router)?

The first time when fc-agent.service runs under the new system. There exists the possibility that a failover could cause a router to be promoted to primary before this happens, which would cause kea to fail on startup. We already had this problem previously, and `fc-kea` was added to the activation scripts to try and fix this, but as we've seen this causes different issues.

osnyx (approved):

PR #1274, opened 2025-02-06
[PL-133327] nixos/platform: use `nix.settings` for connect-timeout

Ma27:
@osnyx this can be backported to 24.05 & 23.11. For 21.05 I'll need to check.

PL-133327

As discussed, it might be useful to have some kind of feature-flag for the connect-timeout since modifying declarations of `nix.extraOptions` is hard.

Instead, decided to just use `nix.settings` here, that way the timeout can be adjusted with primitives of the module system and without an additional option.

## Release process

- [ ] Created changelog entry using `./changelog.sh`: Followup for #1256, no behavioral change.

- [x] PR has internal ticket

## Design notes

  - The option `nix.settings.connect-timeout` is now used. This can overriden to any value with `mkForce`.
  - `nix.settings` has documentation already.

## Security implications

  - Maintaining Availability: Being able to quickly change the timeout in case of any issues.
  - Activated this change on a test VM. Then, confirmed that I can change the timeout with
```nix
{ lib, ... }: {
  nix.settings.connect-timeout = lib.mkForce 23;
}
```

### DIFF
--- a/nixos/platform/default.nix
+++ b/nixos/platform/default.nix
@@ -298,7 +298,6 @@ in {
         http-connections = 2
         log-lines = 25
         experimental-features = nix-command flakes fetch-closure
-        connect-timeout = 1
       '';
 
       settings = {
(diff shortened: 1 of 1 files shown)


nixos/platform/default.nix
```diff
@@ -298,7 +298,6 @@ in {
         http-connections = 2
         log-lines = 25
         experimental-features = nix-command flakes fetch-closure
```
osnyx, line 300: We can also move the other extraOptions over to settings.
It's okay not to do this in this PR, because this keeps the change down to its intended minimal core, and it remains backportable.

osnyx (approved):

PR #1275, opened 2025-02-06
[21.05][PL-133327] nixos/platform: make `connect-timeout` configurable

Ma27:
~~__Note:__ this requires testing on a 21.05 machine.~~

PL-133327

On 21.05 we don't `nix.settings` yet, so a custom option it is. Same idea as in #1274.

## Release process

- [ ] Created changelog entry using `./changelog.sh`: Followup for https://github.com/flyingcircusio/fc-nixos/pull/1256, no behavioral change.

- [x] PR has internal ticket

## Design notes

  - The option nix.settings.connect-timeout is now used. This can overriden to any value with mkForce.
  - New option with description got added.
 
## Security implications

- [x] [Security requirments](https://wiki.flyingcircus.io/System_Development_Guideline#Security_requirement_principles_and_testing) defined? (WHERE)
  - Maintaining Availability: Being able to quickly change the timeout in case of any issues.
- [x] Security requirements tested? (EVIDENCE):
  - Activated on a test VM
  - Confirmed that overriding works via
```nix
{ lib, ... }: {
flyingcircus.nix.connectTimeout = lib.mkForce 23;
}
```

### DIFF
--- a/nixos/platform/default.nix
+++ b/nixos/platform/default.nix
@@ -48,6 +48,14 @@ in {
 
   options = with lib.types; {
 
+    flyingcircus.nix.connectTimeout = mkOption {
+      default = 1;
+      type = types.int;
+      description = ''
+        Corresponds to the `connect-timeout` setting of `nix.conf`.
+      '';
+    };
+
     flyingcircus.activationScripts = mkOption {
       description = ''
         This does the same as system.activationScripts,
@@ -235,7 +243,7 @@ in {
       extraOptions = ''
         fallback = true
         http-connections = 2
-        connect-timeout = 1
+        connect-timeout = ${toString cfg.nix.connectTimeout}
       '';
     };
 


nixos/platform/default.nix
```diff
@@ -48,6 +48,14 @@ in {
 
   options = with lib.types; {
 
+    flyingcircus.nix.connectTimeout = mkOption {
```
osnyx, line 51: Inventing a new option here that won't exist in any future release will break the unattended update in maintenance path, as the configuration will then only be valid for one of the platform versions.
Why not explicitly backport the option name `nix.settings.connect-timeout`, that will then be handled by generator-backed settings in future NixOS versions?

osnyx (changes requested):

Ma27:
As discussed in PL-133327, we don't really need this anymore.

PR #1284, opened 2025-02-12
fc-collect-garbage: fix user switching issues, logging

dpausp:

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - user-specific ignore files have to be effective to avoid scanning unwanted files and directories which also could be a security risk
  - permissions for fc-userscan should be limited but scanning was always run as root to be able to scan files with differing UID from container file systems.
  - scripts should log important actions, specifically which gcroots are deleted/created by fc-userscan
  - checked on a test VM that manual an timer-triggered garbage collection run as expected and that a stamp file is written for automated runs 
  - added some Python tests to improve code coverage for garbage collection code in the agent package


nixos/platform/collect-garbage.nix
```diff
@@ -6,7 +6,7 @@
 with builtins;
 
 let
-  cfg = config.flyingcircus;
+  cfg = config.flyingcircus.agent;
   fclib = config.fclib;
   log = "/var/log/fc-collect-garbage.log";
```
dpausp: this is actually a stamp file which is read by sensu

pkgs/fc/agent/fc/util/logging.py
```diff
@@ -534,51 +534,67 @@ def format_exc_info(logger, name, event_dict):
...
+
     # The invocation ID is normally set by systemd when the script is called
     # from a systemd unit.
     invocation_id = os.environ.get("INVOCATION_ID")
     if invocation_id:
         formatted_dt = datetime.now().strftime("%Y-%m-%dT%H_%m_%S")
         cmd_log_file_name = (
-            logdir / f"fc-agent/{formatted_dt}_build-output"
-            f"_{invocation_id}.log"
+            logdir / f"{formatted_dt}_{log_name}_{invocation_id}.log"
```
dpausp, line 578: we have to clean up old log files, they are not included in the current systemd tmpfiles rules

nixos/platform/collect-garbage.nix
dpausp: stamp file is back now, but it's explicitly named as such instead of calling it a log file. Also, it's placed under /var/lib/fc-collect-garbage now to make clear that it belongs to the fc-collect-garbage service

dpausp:
Doesn't work as expected. A new idea is to replace fc-userscan by ripgrep and move other features to the calling Python code.

osnyx:
Rebased to match new pre-commit restrictions after #1375.

pkgs/fc/agent/fc/manage/collect_garbage.py
```diff
@@ -28,100 +30,194 @@
...
-            preexec_fn=lambda: os.setresuid(user.pw_uid, 0, 0),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
         )
-        rc = p.wait()
-        log.debug("userscan-result", rc=rc)
-        return_codes.append(rc)
 
+        # FIXME: unused variable, will this be used in the future?
```
osnyx, line 209: @dpausp ruff complained when rebasing, please review whether this is still needed.

ctheune:
this is dead, jim

PR #1291, opened 2025-02-18
[FC-42506] roles/lamp: make interpreters of all pools available globally

Ma27:
It's a common pattern in a lot of projects to expose `php-{name}` (or `{name}-php`) which wraps the interpreter used by the phpfpm pool.

This patch adds this behavior to the platform and updates the docs accordingly.

Existing projects are still able to upgrade with `php-{name}` in `systemPackages` since the priority of each wrapper package is set to low.

@frlan may also be interested in this.

## Release process

- [x] PR has internal ticket

## Design notes

The wrapper packages use `lib.lowPrio` to avoid conflicts with existing things in `systemPackages`, so it's nothing that interferes with existing configurations.

## Security implications

  - Provide the same PHP (same version, same config) used by the application for operators on the console and other background services of the application written in PHP.
  - Activated on a test VM.
  - Made sure that this doesn't break the build if such a thing was implemented already.
  - Checked that the correct phpfpm configuration is used, even if `services.phpfpm.pools.${name}` has manual overides.

### DIFF
--- a/changelog.d/20250218_125457_fc-24.11-dev_scriv.md
+++ b/changelog.d/20250218_125457_fc-24.11-dev_scriv.md
@@ -1,0 +1,21 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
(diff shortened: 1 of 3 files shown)


frlan:
LGTM

osnyx (approved):

PR #1292, opened 2025-02-18
logrotate: relax hardening of service to fix postrotate scripts

osnyx:
Logrotate was broken for several services lika haproxy, ceph, or mysql: the postrotate script was not properly executed, causing the log stream not being switched over to the current file.

This happened due to hardening of the systemd service being increased in NixOS 24.11. This hardening is a bit too tight for our platform, so we need to relax some aspects again.

This change alone does not fix an incorrect log file descriptor assignment of running services. As it is going to be rolled out together with a kernel update, the resulting system reboot will make services start from a fresh state again.

PL-133439

## Release process

- [x] PR has internal ticket

## Design notes

  -  user logrotate services not affected

## Security implications

  - fix a regression in log rotation to avoid machines running out of disk space
  - test coverage might need to be extended as this regression went uncaught
  - must not cause any known regressions
  - [ ] automated tests still pass
  - [x] tested the change manually on affected machines with haproxy, ceph, mysql
  - [x] follow-up ticket to extend test case coverage

### DIFF
--- a/changelog.d/20250218_163706_PL-133439-logrotate-unhardening_scriv.md
+++ b/changelog.d/20250218_163706_PL-133439-logrotate-unhardening_scriv.md
@@ -1,0 +1,18 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+
+### NixOS XX.XX platform
+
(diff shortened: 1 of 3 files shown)


changelog.d/20250218_163706_PL-133439-logrotate-unhardening_scriv.md
```diff
@@ -0,0 +1,22 @@
...
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- A bullet item for the Impact category.
```
ctheune: superfluous

ctheune (dismissed):

PR #1312, opened 2025-02-28
routers: fix pmacctd traffic accounting bind to interface

osnyx:
Fix a regression in pmacctd after it was updated from 1.7.5 (NixOS 21.05) to 24.11 (1.7.9), where the `interface` config key was removed. The same functionality can now be configured with `pcap_interface`. The change can be narrowed down to pmacct-1.7.7, revision 8583beaadc4b8d559b84edc0f7273e59aea842d8, but was never properly documented by the upstream project.

This had the effect that pmacctd ignored that config option and fell back to accounting *all* interfaces. As a result, traffic was accounted 3 times (incoming physical, vx… bridge).
As we are running 2 instances of this service, originally designated to account only fe and srv each, we get 2*3 = 6x.

Additionally to updating the config key name, our NixOS test now checks for any warning issued by pmacctd, and also verifies that log output confirms binding to the correct interface.

PL-133497

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - is a hotfix for a regression => include regression test
  - must not impact remaining router functionality
  - [x] extended test suite to ensure absence of warnings (negative match) and logging of correct interface binding (positive match)
  - [x] rolled out to a development router
  - [x] on a testvm utilising that router, downloaded file of known size, and
    - verified it is only accounted on the correct interface
    - verified the accounted size is correct
    - done for ethsrv and ethfe

### DIFF
--- a/changelog.d/20250228_165417_PL-133497-pmacctd-interface-hotfix_scriv.md
+++ b/changelog.d/20250228_165417_PL-133497-pmacctd-interface-hotfix_scriv.md
@@ -1,0 +1,17 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+
+### NixOS XX.XX platform
+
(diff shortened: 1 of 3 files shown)


ctheune (approved):
The assertions are superfluous. Lets keep the train going, though. Please remove them next week.

PR #1320, opened 2025-03-04
[PL-133484] nixos/platform: change default nix to 2.18 (prod) or 2.25 (non-prod)

Ma27:
Nix 2.24's inefficient Git tarball cache is a major problem for VMs[1] with low IOPS (default is 250). The way forward is

* switch back to Nix 2.18 on all production VMs.
* roll out Nix 2.25 as designated successor to all non-production VMs (approach derived from the verification kernel topic from eba8be36bd4c5490dbf23ed9265f44fea92e3e42).

The challenge here is that `flyingcircus.agent.package` is supposed to be modified by downstream consumers, e.g. for Slurm support. Hence, the option isn't modified directly, but each agent package is post-processed via `apply` to use `config.nix.package`.

A nice side-effect of this is that setting `nix.package` also changes the Nix used by the agent, so `nix.package` behaves as I'd expect it. Please note that this also means that setting `nix.package` to e.g. Nix 2.26 implies a rebuild of the agent now.

I decided against overriding `pkgs.nix` with an overlay since there are a bunch of packages out there that explicitly require a specific Nix version, so the potential fallout from that is higher than modifying `nix.package`.

Additionally I changed the usage of Nix 2.24 in the following places:

* The PATH of `fc-collect-garbage.service` doesn't have a Nix at all anymore. The agent package already prefers its own Nix, so this had no effect at all.

* The agent isn't built against 2.24 in pkgs/fc: I see no reason to do that since there's zero usage of this. It's now built with Nix 2.18 since that's what the majority of all VMs will use for now.

  The variant with 2.25 is also built by Hydra because of the VM-test, so staging VMs don't have to build their own agent.

* The sensu check-env uses the default Nix as well.

[1] See PL-133484 for measurements with Nix versions and related
    upstream bugs.

## Release process

- [x] PR has internal ticket

## Design notes

  - 2.18/2.25 can be toggled with a newly introduced option. An arbitrary Nix version can be selected with `nix.package` (that also changes the Nix used by the agent now).

## Security implications

  - Availability: Nix 2.24 has a major performance regression regarding the unpacking/caching of tarballs which is especially a problem on VMs with low IOPS.
  - Measurements from PL-133484 show that Nix 2.25 and Nix 2.18 are equivalent in terms of performance and are faster than 2.24. Hence, 2.18 is being rolled out as known-good version for production and 2.25 on staging to discover potential problems with it before switching the entire platform to 2.25.

### DIFF
--- a/nixos/platform/nix.nix
+++ b/nixos/platform/nix.nix
@@ -1,0 +1,35 @@
+{ config, lib, pkgs, ... }:
+
+let
+  production = lib.attrByPath [ "parameters" "production" ] "" config.flyingcircus.enc;
+
+  nixPackage = if config.flyingcircus.nix.useUnstableNix
+    then pkgs.nixVersions.nix_2_25
+    else pkgs.nixVersions.nix_2_18;
+in {
+  options.flyingcircus = {
+    nix.useUnstableNix = lib.mkOption {
+      default = production == false;
+      defaultText = lib.literalExpression ''production == false'';
+      type = lib.types.bool;
+      description = ''
+        Whether to use a known stable Nix (i.e. 2.18) or a
+        newer, potentially unstable version (i.e. 2.25).
+      '';
+    };
+
+    # The option is defined in `<fc-nixos/nixos/platform/agent.nix>`.
+    # This injects a function that makes sure that the agent uses the correct
+    # Nix version.
+    #
+    # It's not feasible to modify `config.flyingcircus.agent.package` for this,
+    # since downstream consumers may do that already, e.g. for slurm support.
+    agent.package = lib.mkOption {
+      apply = package: package.override { nix = config.nix.package; };
+    };
+  };
+
+  config = {
+    nix.package = nixPackage;
+  };
+}
--- a/pkgs/fc/default.nix
+++ b/pkgs/fc/default.nix
@@ -3,8 +3,10 @@
 rec {
   recurseForDerivations = true;
 
-  agent = pythonPackages.callPackage ./agent {};
-  agentWithSlurm = pythonPackages.callPackage ./agent { enableSlurm = true; };
+  agent = pythonPackages.callPackage ./agent {
+    nix = pkgs.nixVersions.nix_2_18;
+  };
+  agentWithSlurm = agent.override { enableSlurm = true; };
 
   blockdev = callPackage ./blockdev {};
 
--- a/changelog.d/20250304_120935_PL-133484-nix-version-changes_scriv.md
+++ b/changelog.d/20250304_120935_PL-133484-nix-version-changes_scriv.md
@@ -1,0 +1,24 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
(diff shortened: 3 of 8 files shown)


nixos/platform/nix.nix
```diff
@@ -0,0 +1,36 @@
...
+  production = lib.attrByPath [ "parameters" "production" ] "" config.flyingcircus.enc;
+
+  nixPackage = if config.flyingcircus.nix.useUnstableNix
+    then pkgs.nixVersions.nix_2_25
+    else pkgs.nixVersions.nix_2_18;
+in {
+  options.flyingcircus = {
+    nix.useUnstableNix = lib.mkOption {
+      default = (location == "dev") || (location == "whq") || (production  == false);
+      defaultText = lib.literalExpression ''(location == "dev") || (location == "whq") || (production  == false)'';
```
Ma27: @osnyx you mentioned that you'd prefer 2.18 on staging first. This commit is the state I'd like to see in production at some point.

Would it make sense to you to push another commit on top that sets the default to `false` and adjusts the test accordingly? We could just revert the commit then when this can go to production.

EDIT: follow-up question: should the changelog reflect that? If so, we should probably write down the dates when this will change in production.

osnyx: Another suggestion would be to also split whq and dev machines depending on whether those are production or not. At least on the VM side we have both prod and non-prod machines in dev and WHQ as well.
```suggestion
      defaultText = lib.literalExpression ''production  == false'';
```

This obviously reduces the number of hosts exposed to Nix 2.25 again, especially for hardware (almost all of our infra is marked as production). But combining this with *push another commit on top*, we could go with my suggestion for this release cycle, and possibly move to the one you suggested later on.

osnyx: As we have 2 releasetest VMs per branch, I also suggest configuring one of them to use the stable nix for the time being.

osnyx (changes requested):
LGTM apart from my remark on the useUnstableNix defaults.

pkgs/fc/default.nix
```diff
@@ -3,8 +3,10 @@
 rec {
   recurseForDerivations = true;
 
-  agent = pythonPackages.callPackage ./agent {};
-  agentWithSlurm = pythonPackages.callPackage ./agent { enableSlurm = true; };
+  agent = pythonPackages.callPackage ./agent {
+    nix = pkgs.nixVersions.nix_2_18;
```
ctheune, line 7: This one I don't understand. Isn't this undermining all the option stuff?

ctheune (changes requested):

pkgs/fc/default.nix
Ma27: No, this gets the correct Nix version with the `override` in `nix.nix`.

I also changed this here since I don't see a reason to build three variants of the agent, i.e. one with 2.24 (the default nix), and the variants with 2.25/2.18.

osnyx: Maybe to elaborate a bit further: I also stumbled upon this, but then understood that this part is necessary for when the agent is just used as a Nix package in a context where the NixOS module system is not evaluated. E.g. when installing it imperatively as a package, pulling it into a nix-shell, etc. also transitively via other packages.

@Ma27 decided to give it a safe default, which then *in the NixOS module system context, may* be overridden to use another Nix, e.g. in the agent *service*.

osnyx (approved):

PR #1321, opened 2025-03-04
Add patch for FRR with regression fix for MAC mobility

sysvinit:
This change introduces a patch to our pinned FRR version which fixes a regression affecting MAC mobility (and hence VM migration) introduced between 10.0 and 10.1.

Depending on ordering between different hosts, a receiving KVM host may learn the MAC address of a migrating guest before the sending host stops announcing it. Due to the regression, bgpd would frequently decide that the remote route was a "better" route than the locally learned route and thus delete the latter, which would mean that when the remote route was withdrawn when the migration completed, bgpd would not announce a route for the locally learned MAC.

This would lead to routes for guest MAC addresses going missing, causing unnecessary flooding from the other hosts in the EVPN fabric and eventually to site-wide network performance degradation.

This regression has been fixed upstream and backported to the stable branch, but is yet to appear in a 10.1.x stable release.

PL-133422

## Release process

- [x] ~~Created changelog entry using `./changelog.sh`~~

- [x] PR has internal ticket

## Design notes

## Security implications

  - This is a recognised regression upstream, and while the fix has been merged it's not yet in a public release.
  - Passes the chaos monkey smoke test in dev (which the unpatched version does not pass).

### DIFF
--- a/pkgs/overlay.nix
+++ b/pkgs/overlay.nix
@@ -123,6 +123,12 @@ builtins.mapAttrs (_: patchPhps (fetchpatch {
 
     patches = [
       ./frr/0001-Don-t-throw-error-when-log-directory-already-exists.patch
+      # issue with best path selection erroneously deleting local mac
+      # addresses during incoming guest migration.
+      (fetchpatch {
+        url = "https://github.com/FRRouting/frr/commit/c3e264e27ac5b682c4647aeeaf4e61f09d5243f9.patch";
+        hash = "sha256-YUNyE2ZeW3q58E+fpFhmMGpp56eba8HUI7pcLLbQDTE=";
+      })
     ];
   });
 


pkgs/overlay.nix
```diff
@@ -123,6 +123,12 @@ builtins.mapAttrs (_: patchPhps (fetchpatch {
 
     patches = [
       ./frr/0001-Don-t-throw-error-when-log-directory-already-exists.patch
+      # issue with best path selection erroneously deleting local mac
+      # addresses during incoming guest migration.
+      (fetchpatch {
+        url = "https://github.com/FRRouting/frr/commit/c3e264e27ac5b682c4647aeeaf4e61f09d5243f9.patch";
```
osnyx, line 129: Note: `.patch` URIs are not always stable over longer times, better prefer `.diff` urls. One issue is that git(hub) auto-scales the lenth of the revision specifier with the growing number of commits in the repo. https://github.com/NixOS/nixpkgs/issues/257446#issuecomment-1736563091

For github, the following helps:
```suggestion
        url = "https://github.com/FRRouting/frr/commit/c3e264e27ac5b682c4647aeeaf4e61f09d5243f9.patch?full_index=1";
```

For gitlab, `.diff` URIs shall be preferred.

As this is only a temporary patch and will be pulled from the next upstream release instead, this is a purely informational note with no need to change the URI.

osnyx (approved):

PR #1329, opened 2025-03-10
overlay, php: build php82/83 with patched curl on hydra

osnyx:
Due to a regression, we need to override the curl used in php82 and php83 in several customer projects.
To ensure it is built by hydra and available in binary cache instead of having to build it locally, we temporarily add it to the list of packages.

FC-40100 FC-40106

## Release process

  - no, the packages are mainly designed for internal use 

- [ ] PR has internal ticket

## Design notes

## Security implications

  - [x] availability: reduce local system load of VMs for common package variations
  - [x] ensures packages are built by hydra
    - https://hydra.flyingcircus.io/build/4387591
    - https://hydra.flyingcircus.io/build/4387470

### DIFF
--- a/pkgs/overlay.nix
+++ b/pkgs/overlay.nix
@@ -44,6 +44,7 @@ builtins.mapAttrs (_: patchPhps phpLogPermissionPatch) {
   inherit (phps) php72 php73 php74 php80;
   # Import NixOS upstream PHPs.
   inherit (super) php81 php82 php83 php84;
+
 }
 //
 builtins.mapAttrs (_: patchPhps (fetchpatch {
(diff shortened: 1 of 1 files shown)


pkgs/overlay.nix
```diff
@@ -52,6 +53,37 @@ builtins.mapAttrs (_: patchPhps (fetchpatch {
...
+# Use curl 8.4.0 for PHP
+# FC-40106, FC-40100
+// lib.listToAttrs (map (v: lib.nameValuePair "php${v}-FC-40106" (
+  super."php${v}".override {
+          packageOverrides = final: prev: {
+            extensions = prev.extensions // {
+              curl = prev.extensions.curl.overrideAttrs (_: let
+                curl_8_4 = self.curl.overrideAttrs (_: {
+                  version = "8.4.0";
+                  name = "curl-8.4.0";
```
leona-ya: This should not be needed, as set be pname+version

osnyx: I have copied that code in verbatim from the affected customer project. But as removing this should not change the resulting derivation, i can adjust that as well.

leona-ya (approved):

PR #1337, opened 2025-03-12
percona83: bring back role to allow upgrading existing 24.05 VMs

osnyx:
We have decided that the percona84 role changes (PL-133082) will be developed on the unstable platform development branch, targeting NixOS 25.05. This *might* be backported to 24.11 at some point, but this is not certain.

To allow a timely upgrade of 24.05 VMs to 24.11, we instead also offer the EOL percona83 role on 24.11.

PL-133537

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - allow existing VMs to update to newer platform versions
  - We conciously add an end-of-life package version again, as allowing existing machines to at least update the rest of the surrounding platform already improves the overall security 
  - [x] added percona83 to percona role tests


pkgs/overlay.nix
```diff
@@ -452,11 +452,21 @@ builtins.mapAttrs (_: patchPhps (fetchpatch {
...
+    protobuf = self.protobuf_21;
+  };
 
   percona-xtrabackup_2_4 = super.callPackage ./percona-xtrabackup/2_4/2_4.nix {
     boost = self.boost159;
     openssl = self.openssl_1_1;
   };
+  # EOL, but we vendor it throughout the 24.11 release cycle
+  percona-xtrabackup_8_3 = self.callPackage ./percona-xtrabackup/8_3/innovation.nix { };
+  percona-xtrabackup_8_4 = super.percona-xtrabackup_8_4;
```
ctheune: is the 8.4 on purpose?

ctheune (changes requested):

pkgs/overlay.nix
osnyx: Not really. The 84 packages are already in nixpkgs anyways, but adding that name only here is inconsistent with other places. Removing it.

ctheune (approved):

PR #1339, opened 2025-03-13
devhost: improve error handling in image download

leona-ya:
PL-133539

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Should not interrupt ops
  - Bugfix, can only cause less breakage. Tested on largo00

### DIFF
--- a/nixos/roles/devhost/fc-devhost.py
+++ b/nixos/roles/devhost/fc-devhost.py
@@ -271,6 +271,7 @@ class Manager:
                     # Download the base image. We rename the file afterwards
                     # to ensure that the image is fully there.
                     r = requests.get(image_url)
+                    r.raise_for_status()
                     with open(vm_base_image_path_tmp, "wb") as f:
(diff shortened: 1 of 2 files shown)


nixos/roles/devhost/fc-devhost.py
```diff
@@ -271,6 +271,7 @@ def ensure(
                     # Download the base image. We rename the file afterwards
                     # to ensure that the image is fully there.
                     r = requests.get(image_url)
+                    r.raise_for_status()
```
osnyx, line 274: This then causes the deployment process to be aborted, right? Some retries could be desirable for a better UX, but this at lease prevents incorrect base image data from becoming persisted.

osnyx (approved):

nixos/roles/devhost/fc-devhost.py
leona-ya: Correct

PR #1341, opened 2025-03-14
fc.qemu improvements for test stability

ctheune:
- make the ceph roles easier to adapt to simpler network environments

- make it possible to disable the s3 proxy

- provide system-wide access to tools that are generally needed

- improve ceph setup routines

Re PL-133300

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

no security relevant changes

nothing to tests, improving overall test quality

### DIFF
--- a/nixos/roles/kvm.nix
+++ b/nixos/roles/kvm.nix
@@ -4,16 +4,23 @@ with builtins;
 
 let
   fclib = config.fclib;
-  role = config.flyingcircus.roles.kvm_host;
+  cfg = config.flyingcircus.roles.kvm_host;
   enc = config.flyingcircus.enc;
 
-  cephPkgs = fclib.ceph.mkPkgs role.cephRelease;
+  cephPkgs = fclib.ceph.mkPkgs cfg.cephRelease;
 
 in
 {
   options = {
     flyingcircus.roles.kvm_host = {
       enable = lib.mkEnableOption "Qemu/KVM server";
+      enableS3Proxy = lib.mkOption {
+        default = true;
+        example = false;
+        description = "Whether to enable the local S3 proxy.";
+        type = lib.types.bool;
+      };
+
       supportsContainers = fclib.mkDisableDevhostSupport;
       mkfsXfsFlags = lib.mkOption {
         type = with lib.types; nullOr str;
@@ -40,10 +47,17 @@ in
       cephRelease = fclib.ceph.releaseOption // {
         description = "Codename of the Ceph release series used by qemu.";
       };
+
+      network = lib.mkOption {
+        type = with lib.types; attrs; # an attrset from fclib.network.<xy>
+        default = fclib.network.sto;
+        description = "Network to use for migration";
+      };
+
     };
   };
 
-  config = lib.mkIf role.enable {
+  config = lib.mkIf cfg.enable {
 
     # Do not enable the watchdog for KVM hosts globally as we dealt with
     # way too many times.
@@ -51,7 +65,7 @@ in
 
     flyingcircus.services.ceph.client = {
       enable = true;
-      cephRelease = role.cephRelease;
+      cephRelease = cfg.cephRelease;
     };
 
     # toolpath for agent (fc-create-vm)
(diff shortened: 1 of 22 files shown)


leona-ya (approved):

nixos/services/ceph/client.nix
```diff
@@ -152,6 +150,32 @@ in
             Can override existing default setting values. Configuration keys like `mon osd full ratio` can alternatively be written in camelCase as `monOsdFullRatio`.
             '';
         };
+
+        mons = lib.mkOption {
+          type = with lib.types; listOf str;
+          default = (map
+              (mon: "${head (lib.splitString "." mon.address)}.${cfg.network.vlan}.${location}.ipv4.gocept.net")
```
leona-ya: doesn't eval, need to be

```suggestion
              (mon: "${head (lib.splitString "." mon.address)}.${cfg.client.network.vlan}.${location}.ipv4.gocept.net")
```

leona-ya (dismissed):

nixos/roles/kvm.nix
```diff
@@ -4,16 +4,23 @@ with builtins;
 
 let
   fclib = config.fclib;
-  role = config.flyingcircus.roles.kvm_host;
+  cfg = config.flyingcircus.roles.kvm_host;
```
osnyx, line 7: That's mainly a matter of taste, but okay, I am not going to die on that hill.
The distinction between `role` and `service` (latter often also `cfg`) is useful in cases where we have a role configuring a corresponding service module in the same file, so that the distinction between those 2 scopes is immediately clear. This is not strictly the case here, so feel free to change it like this.
Just some background on why the naming was like this and how I would not support generally changing this throughout the whole platform code.

nixos/services/ceph/client.nix
```diff
@@ -166,6 +190,7 @@ in
       }
     ];
 
+    flyingcircus.services.ceph.fc-ceph.enable = true;
```
osnyx, line 193: Why do ceph clients (e.g. backyservers, KVM hosts) require fc-ceph? My understanding so far was that fc-ceph is intended as a util to manage ceph cluster **hosts**. Right now this is limited by the intertwining of fc-ceph and fc-luks, but there is at least the *intention* to factor these two out.

pkgs/fc/qemu/default.nix
```diff
@@ -88,7 +91,7 @@ in
       py.pytest-cov
       py.pytest-timeout
       py.mock
-
+      fc-ceph
```
osnyx, line 94: Same as above, this pulls in `fc-qemu` into a KVM host, which is just a ceph client.

tests/kvm_host_ceph-nautilus.nix
```diff
@@ -99,39 +102,29 @@ let
...
-            echo "Pytest result code: $PYTESTRET"
-            if [ $PYTESTRET -eq 0 ] || [ $PYTESTRET -eq 5 ]; then
-              # 5 means no tests found, which might happen if we have options
-              true;
-            fi
-            '' else ""}
+            pytest -vv --cov-config=/etc/coveragerc --cov-append -c ${testPackage.src}/pytest.ini "$@" | tee /dev/console
 
+            # We're not looking at the exit code of the above pipe as the
+            # tee to /dev/kmsg sometimes causes a non-zero exit code ...
```
osnyx:
```suggestion
            # tee to /dev/console sometimes causes a non-zero exit code ...
```

tests/kvm_host_ceph-nautilus.nix
```diff
@@ -498,11 +384,12 @@ in
       result = show(host1, "fc-qemu --help")
       assert result.startswith("usage: fc-qemu"), "Unexpected help output"
 
-      result = show(host1, "fc-qemu ls")
-      assert "I simplevm              offline" in result, repr(result)
+      # host1.succeed("cp /etc/simplevm.cfg /etc/qemu/vm/")
```
osnyx, line 385: Deactivated because
> `# XXX the following tests should be migrated to fc.qemu at some point`

?

nixos/roles/kvm.nix
ctheune: Interesting. I wasn't quite aware of that ... interestingly `cfg` is for me the "this is the locally in this file/module relevant configuration" and nothing more. The role/service distinction we have ... well yeah ...

nixos/services/ceph/client.nix
ctheune: yeah, sure, that seems like a mistake.

pkgs/fc/qemu/default.nix
ctheune: No, these are the test dependencies. fc-qemu does call those in the test suite.

tests/kvm_host_ceph-nautilus.nix
ctheune: Aeh nope ... /dev/kmsg causes issues. /dev/console is fine. However, the comment is now superfluous. I still think looking at the report file is a good idea.

tests/kvm_host_ceph-nautilus.nix
ctheune: yeah, those are not very valuable tests and I'm out of time to port them all and they rely implicitly on data left over by the qemu test suite (which stopped working after the refactoring)

osnyx (dismissed):

osnyx (dismissed):
Eval finally succeeds. LGTM, as long as the tests still pass.

leona-ya (dismissed):

osnyx (approved):
Okay, my bad.

PR #1342, opened 2025-03-14
flyingcircus.services.rabbitmq: enable prometheus exporter, through telegraf

osnyx:
In preparation for the deprecation of the management_metrics_collection feature, we *additionally* expose rabbitmq metrics via the prometheus exporter.
To unify metrics discovery, we funnel that prometheus exporter through the local telegraf instance to collect all node metrics before providing them via a prometheus export again.

So far, the old metrics collection exists in parallal to provide a transition period for dashboard changes.

PL-133391

## Release process

- [x] PR has internal ticket

## Design notes

  - the rabbitmq prometheus exporter utilises and additional port number. That port can be customised via NixOS config if necessary 

## Security implications

  - ensure availability of metrics: provide a parallel metrics export in the face of an impending deprecation of the existing metrics export route
  - [x] no regressions: automated tests still pass
  - [x] did a quick plausibility check via curl on the telegraf and native rabbitmq exporter endpoint

### DIFF
--- a/nixos/services/rabbitmq/default.nix
+++ b/nixos/services/rabbitmq/default.nix
@@ -61,6 +61,17 @@ in {
         type = types.int;
       };
 
+      prometheusPort = mkOption {
+        default = 9125;
+        description = ''
+          Listening port for the RabbitMQ Prometheus exporter. Is exposed via
+          telegraf again, only change this value of the default port is used otherwise.
+          '';
+        type = types.port;
+        internal = true;
+        apply = builtins.toString;
+      };
+
       dataDir = mkOption {
         type = types.path;
         default = "/var/lib/rabbitmq";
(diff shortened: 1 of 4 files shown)


osnyx:
@ctheune I could still use some guidance for deciding on the metrics format used by telegraf:

> ### Metric Format Configuration
> 
> The `metric_version` setting controls how telegraf translates prometheus format
> metrics to telegraf metrics. There are two options.
> 
> With `metric_version = 1`, the prometheus metric name becomes the telegraf
> metric name. Prometheus labels become telegraf tags. Prometheus values become
> telegraf field values. The fields have generic keys based on the type of the
> prometheus metric. This option produces metrics that are dense (not
> sparse). Denseness is a useful property for some outputs, including those that
> are more efficient with row-oriented data.
> 
> `metric_version = 2` differs in a few ways. The prometheus metric name becomes a
> telegraf field key. Metrics hold more than one value and the field keys aren't
> generic. The resulting metrics are sparse, but for some outputs they may be
> easier to process or query, including those that are more efficient with
> column-oriented data. The telegraf metric name is the same for all metrics in
> the input instance. It can be set with the `name_override` setting and defaults
> to "prometheus". To have multiple metric names, you can use multiple instances
> of the plugin, each with its own `name_override`.
> 
> `metric_version = 2` uses the same histogram format as the [histogram
> aggregator](../../aggregators/histogram/README.md)
> 
> The Example Outputs sections shows examples for both options.
> 
> When using this plugin along with the prometheus_client output, use the same
> option in both to ensure metrics are round-tripped without modification.

nixos/services/rabbitmq/default.nix
```diff
@@ -251,6 +266,7 @@ in {
         rabbitmqctl list_users | grep guest && \
           rabbitmqctl delete_user guest
 
+        # TODO: can we remove the fc-telegraf user when having migrated to prometheus?
```
ctheune: yes.

ctheune (dismissed):

osnyx:
Regarding `metric_version`: As it is important to use the same metric_version as in `outputs.prometheus_client`, we continue to use the same default of `metric_version = 1` as used there.

ctheune (approved):

PR #1348, opened 2025-03-19
KVM: integration for guests with interfaces in layer 3 routed networks

sysvinit:
This change is the companion to flyingcircusio/fc.qemu#23. This change extends the KVM role to support guest VMs which have interfaces attached to layer 3 routed networks, implemented as EVPN-VXLAN L3VNIs.

- Provide fc.qemu with extra scripts for configuring guest interfaces connected to VRFs (as these are set up differently to interfaces connected to bridges) and specify the script paths in the configuration file.
- Assign virtual gateway addresses to the host side of the guest tap interfaces. These addresses are used as the default gateway in the guest network configuration in order to route through the point-to-point layer 2 connection between host and guest to reach other destinations.
- Reattach guest interfaces to VRF kernel device when the latter is restarted, analogue to the existing configuration for handling restarts of `brfe` and `brsrv`.
- Enable forwarding in order to provide layer 3 connectivity to guests on routed networks, with forwarding policy enforced by iptables.
- Provide DNS resolver services on the IPv4 virtual gateway address using DNAT in the host firewall to forward queries to the site-wide resolver address.

TODO: once the fc.qemu PR is merged, bump the version pin of the fc.qemu package here.

PL-133324

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - The KVM role now turns on IP forwarding globally on all interfaces (due to the kernel configuration interface for IPv6 forwarding). This has been a source of security problems in the past, however this change also includes a restrictive firewall to ensure that only guest traffic inside VRFs is forwarded.
  - Developed and tested on a pair of hosts in dev.

### DIFF
--- a/nixos/roles/kvm.nix
+++ b/nixos/roles/kvm.nix
@@ -304,6 +304,40 @@ in
 
     };
 
+    systemd.services.fc-qemu-reattach-vrf-taps = lib.mkIf (fclib.network ? pub) {
+      description = "Reattach all VM taps to VRF devices if needed.";
+
+      path = [
+        pkgs.jq
+        pkgs.iproute2
+      ];
+
+      script = ''
+        for interface in $(ip -j link show |  jq '.[] | .ifname' -r | egrep '^tpub'); do
+          echo "Ensuring attachment of $interface"
+          /etc/kvm/kvm-ifup-vrf $interface || true
+        done
+      '';
+
+      wantedBy = [ "multi-user.target" ];
+      bindsTo = [
+        "vrfpub-netdev.service"
+      ];
+      after = [
+        "vrfpub-netdev.service"
+      ];
+
+      # trigger a scrub when the vrf netdev changes to ensure that the
+      # host routes for the guest interfaces are also set up properly.
+      wants = [ "fc-qemu-scrub.service" ];
+      before = [ "fc-qemu-scrub.service" ];
+
+      serviceConfig = {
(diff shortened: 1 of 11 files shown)


osnyx:
needs **no** rebase after #1375

nixos/roles/kvm.nix
```diff
@@ -260,6 +308,40 @@ in
...
+        "vrfpub-netdev.service"
+      ];
+      after = [
+        "vrfpub-netdev.service"
+      ];
+
+      # trigger a scrub when the vrf netdev changes to ensure that the
+      # host routes for the guest interfaces are also set up properly.
+      wants = [ "fc-qemu-scrub.service" ];
+      before = [ "fc-qemu-scrub.service" ];
```
sysvinit, line 337: If a guest interface becomes detached from the VRF interface, then the routes in the VRF routing table for that interface will be deleted. Simply reattaching the interface won't make the guest behind the interface reachable again, but requires an ensure run to reinstall the routes.

While automatically performing a scrub if vrfpub is restarted seems like the correct action, I'm unsure whether it's necessary/a good idea to do this proactively outside of the regular ten-minutely timer.

ctheune (approved):
fine, @sysvinit did review my changes here.

ctheune:
Hydra isn't quite clean yet, but the KVM test works (we got a cached build for it anyway) and the other tests were clean before.

PR #1350, opened 2025-03-19
varnish: fix check-varnish-http check

osnyx:
The check failed because `awk` is not PATH of the sensuclient.
Replacing the old bash script with a less brittle
python-batteries-included script to reduce complexity.
And while we're at it, now also check the IPv6 addresses.

PL-133554 PL-133556

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - availability checks should work correctly
  - a partial check failure should not accidentally produce the outcome of an apparent success 
  - tested happy case on a test VM
  - added additional safety belt for the regression fixed. Manually tested that service belt by messing up the `ADDRESSES` gathering step in the script.

### DIFF
--- a/nixos/roles/webproxy.nix
+++ b/nixos/roles/webproxy.nix
@@ -61,18 +61,59 @@ in
           command = "${cfg.package}/bin/varnishadm status";
           timeout = 180;
         };
-        varnish_http = {
-          notification = "varnish port 8008 HTTP response";
-          command = "${pkgs.writeShellScript "check-varnish-http" ''
-            ADDRS=$(${cfg.package}/bin/varnishadm debug.listen_address | awk '/([0-9.]+\.)+/ { print $2":"$3; }')
-            for ADDR in $ADDRS; do
-              host=$(echo $ADDR | cut -d ":" -f 1)
-              port=$(echo $ADDR | cut -d ":" -f 2)
+        varnish_http = let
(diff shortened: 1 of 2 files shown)


nixos/roles/webproxy.nix
```diff
@@ -61,18 +61,59 @@ in
           command = "${cfg.package}/bin/varnishadm status";
           timeout = 180;
         };
-        varnish_http = {
+        varnish_http = let
+          check-varnish-http = pkgs.writers.writePython3 "check-varnish-http" {
+            flakeIgnore = [ "E501" ]; # ignore long lines
```
ctheune, line 66: i really really abhor this style of embedding here. if this is a separate standalone python script then it's much easier to achieve a proper feedback cycle during development. this is where code goes to die IMHO.

good enough for the fix now, it does make me wonder why we don't have better ready-made tooling at hand here. either we don't know nixpkgs well enough or there's something else fishy.

ctheune (approved):

nixos/roles/webproxy.nix
osnyx: Yes, I also considered moving this to a standalone file, but this would have required
1. another layer of wrapping to build a PATH with the executables we now instead just insert via string templating
2. having to look at yet another file at some different place to figure out that check script => fragmentation of code details

PR #1351, opened 2025-03-19
Ensure that IPv4 forwarding is disabled by default

sysvinit:
We enable `networking.nat.enable` in the platform in order to ensure that certain iptables change exist which our firewall integration relies upon. However, as an unintended side-effect this also enables IPv4 forwarding. This means that forwarding is enabled on hosts which are not intended to be configured as routers and don't have an appropriate firewall installed. This can lead to effects where unprivileged hosts may be able to configure their local routing tables to abuse more privileged hosts with forwarding enabled in order to access private networks or circumvent firewall restrictions.

This change ensures that IP forwarding is disabled by default, and only enabled in the roles where it is explicitly required.

PL-133557

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Only routers and VMs providing specific network functions should have IP forwarding enabled, as forwarding requires suitable firewall rules for access control.
  - Hydra tests for roles which require forwarding to be enabled.

### DIFF
--- a/changelog.d/20250319_173208_PL-133557-default-forwarding-without-firewall_scriv.md
+++ b/changelog.d/20250319_173208_PL-133557-default-forwarding-without-firewall_scriv.md
@@ -1,0 +1,21 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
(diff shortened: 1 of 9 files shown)


ctheune (approved):
good as a fast fix, i'd appreciate a test in the future

PR #1353, opened 2025-03-20
Add tests to check forwarding sysctls

sysvinit:
This change is a follow-up to #1351, which adds a NixOS to check if the default platform sysctl configuration correctly disables IP forwarding.

PL-133557

## Release process

- [x] ~~Created changelog entry using `./changelog.sh`~~ => not necessary

- [x] PR has internal ticket

## Design notes

## Security implications

  - This is a regression test which should compliment a previous change.
  - Hydra tests.

### DIFF
--- a/tests/testlib.nix
+++ b/tests/testlib.nix
@@ -258,7 +258,7 @@ rec {
 
   fcVlanIfaces = mapAttrs' (vlan: vid: {
     name = "eth${vlan}";
-    value = { vlan = vid; assignIP = true; };
+    value = { vlan = vid; assignIP = false; };
   });
 
   fcIPMap = listToAttrs (concatLists (mapAttrsToList (name: vid: [
--- a/tests/testlib.nix
+++ b/tests/testlib.nix
@@ -87,8 +87,10 @@ rec {
     location ? "test",
     secrets ? {},
     extraEncParameters ? {},
-  }: { config, ... }:
+  }: { lib, config, nodes, ... }:
   let
+    fclib = config.fclib;
+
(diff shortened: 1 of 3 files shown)


tests/testlib.nix
```diff
@@ -257,7 +257,7 @@ rec {
 
   fcVlanIfaces = mapAttrs' (vlan: vid: {
     name = "eth${vlan}";
-    value = { vlan = vid; assignIP = true; };
+    value = { vlan = vid; assignIP = false; };
```
ctheune, line 317: Is this intentional? This looks like it might break a lot of tests and Hydra seems to agree.

ctheune (changes requested):

tests/testlib.nix
sysvinit: Yes, it was intentional. I wasn't sure if there were many tests which relied on the assumption that `assignIP = true` is set, and really we ought not to use that if we're providing our own IP address configuration via the testlib ENC – this results in VM interfaces being assigned *both* the addresses generated by the NixOS test framework and our own testing library. In the new forwarding test in this PR this led to some correctness issues, as some of the IP addresses were assigned to more than one test VM, which was quite confusing. I can `lib.mkForce` the setting in the forwarding test and leave the other tests as they were though.

sysvinit: It turns out that `assignIP = true` also configures the /etc/hosts file, which is used for name resolution in a few tests. I've added extra handling for this and the tests which had problems now seem to pass.

ctheune (approved):

PR #1354, opened 2025-03-20
fclib.utils.python3BinFromFile: allow passing libraries and executables

osnyx:
Extend `python3BinFromFile` to optionally take a list of package
dependencies, whose executables are added to the PATH of the python
script. This facilitates writing more standalone python scripts that
utilise external binaries.
As an example, this is applied to the previously-inlined `check-varnish-http` script.

Also enables the usage of further `writePython3` options like
`libraries`.

Design decisions:
advantages:
- extend an established writer from nixpkgs and benefit from its already
  implemented capabilities
disadvantages:
- yet another wrapping indirection
- something like the `dependencies` arg of `buildPythonPackage` would
  provide an even better UX, but `wrapPythonPrograms` turned out to be
  too complex for this simple use case.

Follow-up for PL-133554.

## Release process

  - internal, do we need one? 

- [x] PR has internal ticket

## Design notes

## Security implications

  - facilitate writing check scripts using external dependencies, to better manage complexity 
  - [x] built some example python scripts with and without external dependencies
  - [x] verified that `check-varnish-http` works correctly on a test vm

### DIFF
--- a/nixos/roles/check-varnish-http.py
+++ b/nixos/roles/check-varnish-http.py
@@ -1,0 +1,42 @@
+import subprocess
+import sys
+
+ports = []
+
+for line in (
+    subprocess.check_output(["varnishadm", "debug.listen_address"])
+    .decode("ascii")
+    .splitlines()
+):
+    line = line.strip()
+    if not line:
+        continue
+    _, ip, port = line.split()
+    ports.append((ip, port))
+
+if not ports:
+    print("No listen_address reported by varnishadm")
+    sys.exit(2)
+
+STATUS = 0
+
+for ip, port in ports:
+    if ":" in ip:
+        ip_version = "-6"
+        print(f"Checking [{ip}]:{port} ...")
+    else:
+        ip_version = "-4"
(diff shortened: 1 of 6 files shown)


nixos/roles/check-varnish-http.py
```diff
@@ -0,0 +1,51 @@
...
+for ip, port in ports:
+    if ":" in ip:
+        ip_version = "-6"
+        print(f"Checking [{ip}]:{port} ...")
+    else:
+        ip_version = "-4"
+        print(f"Checking {ip}:{port} ...")
+    proc = subprocess.run(
+        [
+            "check_http",
```
ctheune: this is a good place to put a # fmt: off and make this more readable ;)

ctheune (changes requested):
great! the one formatting nit in the varnish check, please :)

leona-ya:
I'm relatively sure that I found a nixpkgs integrated alternative, even though it's not well documented. I'll try this.

ctheune (approved):

leona-ya (approved):

PR #1363, opened 2025-03-24
redis: remove ifd for password

leona-ya:
This change makes our gitlab test much more stable, as it doesn't change the password and thus the test derivation randomly.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Passwords shouldn't be more exposed than before: The password being in the nix store isn't good, tho changing that would mean a breaking change. Also our concept with using VMs per service accounts for that.
  - Functionality is ensured by NixOS tests 
  - see above

### DIFF
--- a/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
+++ b/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
@@ -1,0 +1,22 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- If you access `services.redis."".password` in your NixOS config, please change this to
+  `flyingcircus.services.redis.password`.
+
+### NixOS XX.XX platform
+
+- redis: restructure internal password handling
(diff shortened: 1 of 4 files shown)


changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
```diff
@@ -0,0 +1,22 @@
...
+-->
+
+### Impact
+
+
+### NixOS XX.XX platform
+
+- redis: restructure internal password handling
+  The password file /etc/local/redis/password now gets written as systemd ExecStartPre.
+  If you access `services.redis."".password` in your NixOS config, please change this to
```
ctheune: This would be good for the impact section.

ctheune: I'm generally wondering whether we want to deprecate the "store the password in the nix store" approach, at least for this instance.

ctheune (changes requested):

changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
leona-ya: I think it would be good to deprecate this approach, but at a later point? I can also do this here if you want.

ctheune (approved):

changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
ctheune: no, it's fine. lets keep it in mind.

PR #1366, opened 2025-03-25
Revert "redis: remove ifd for password"

leona-ya:

Reverts #1363 

With this change, both the sensu check and the telegraf input wouldn't work anymore when the password is auto-generated.


ctheune:
I've fast-tracked the revert to avoid this progressing to staging as requested by @leona-ya

PR #1367, opened 2025-03-25
services.redis: remove ifd for password

leona-ya:

Also adds a new option `flyingcircus.services.telegraf.environmentVariablesFromFile` that allows users to store
  passwords in files and using them as substituted environment variable in a telegraf input.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

   - Security: Improved. Passwords per default not stored in nix-store
  - Tested sensu and telegraf on dev VM 
  - Tested nixos password gets set when local file doesn't exist

### DIFF
--- a/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
+++ b/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
@@ -1,0 +1,25 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- If you access `services.redis."".password` in your NixOS config, please change this to
+  `flyingcircus.services.redis.password`.
+
+### NixOS XX.XX platform
+
+- redis: restructure internal password handling
+  The password file /etc/local/redis/password now gets written as systemd ExecStartPre. (PL-133653)
+
+- telegraf: Add new option `environmentVariablesFromFile` that allows users to store
+  passwords in files and using them as substituted environment variable in a telegraf input. (PL-133563)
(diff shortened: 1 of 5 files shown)


nixos/services/redis.nix
```diff
@@ -123,17 +112,32 @@ in
...
-    '';
+    systemd.services.redis.serviceConfig =
+      let
+        preStart = pkgs.writeShellScript "redis-prestart" (
+          ''
+            if [[ ! -e /etc/local/redis/password ]]; then (
+            umask 007;
+          ''
+          + lib.optionalString (cfg.password == null) ''
+            ${pkgs.apg}/bin/apg -a 1 -M lnc -n 1 -m 32 > /etc/local/redis/password
```
ctheune, line 123: If redis is used in the mode were the password is assigned (by the deployment) in the nixos configuration, then we should implement this in a convergent manner, i.e. keep updating it even if the file exists. If the pw is not given, then only generate once, though.

nixos/roles/gitlab.nix
```diff
@@ -143,7 +143,7 @@ in
         databaseCreateLocally = fclib.mkPlatform true;
         databasePasswordFile = "${cfg.secretsDir}/db_password";
         initialRootPasswordFile = "${cfg.secretsDir}/root_password";
-        redisUrl = "redis://:${config.services.redis.servers."".requirePass}@localhost:6379/";
+        redisUrl = "redis://:${config.flyingcircus.services.redis.password}@localhost:6379/";
```
ctheune, line 146: needs to be read from file

ctheune (changes requested):

changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
```diff
@@ -0,0 +1,25 @@
...
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- If you access `services.redis."".password` in your NixOS config, please change this to
+  `flyingcircus.services.redis.password`.
```
osnyx, line 17: Just for the record, as we're releasing this during an ongoing 24.11 release cycle: If you don't change this (or do this later), nothing breaks but you continue to get longer eval times due to IFD?

osnyx:
needs **no** rebase after #1375

nixos/roles/gitlab.nix
Ma27: While I've never touched the gitlab module, I just noticed that you can configure another redis like this: https://docs.gitlab.com/omnibus/settings/redis/#using-an-alternate-local-redis-instance

Since the format is Ruby, one could also allow to read the contents of a file inside.

PR #1371, opened 2025-03-26
nginx: make HTTPS sensu check work for vhost name != serverName

dpausp:
The assumption that server names are always determined by the attribute name of the vhost was wrong. The check now prefers to talk to the host identified by vhost.serverName, falling back to the attribute name.

Now, we also check if a vhost really uses SSL as there might be non-TLS but ACME-enabled vhosts which just serve the ACME challenge.

PL-131278

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - just a fix for non-standard nginx configs
  - check should prefer the serverName before using the attribute name of the vhost as fallback
  - check should exclude non-TLS vhost configs
  - checked of a test VM that standard config with serverName == attribute name still works and that a non-standard config with two attribute sets referring to the same serverName, one HTTP and the other HTTPS generates only one Sensu check

### DIFF
--- a/nixos/services/nginx/default.nix
+++ b/nixos/services/nginx/default.nix
@@ -83,7 +83,16 @@ let
     lib.filterAttrs (_: val: val ? emailACME && val.emailACME != null) cfg.virtualHosts
   );
 
-  acmeVhosts = (lib.filterAttrs (_: vhost: vhost.enableACME) nginxCfg.virtualHosts);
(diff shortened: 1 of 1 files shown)


osnyx (approved):
Some kind of test coverage would be neat as well, e.g. checkeing for some typical nginx configs whether this creates a check or not.
Still merging this, as it fixes the checks currently present in a released version.

PR #1355, opened 2025-03-20
codestyle-template: update

osnyx:
**precondition**: already based on https://github.com/flyingcircusio/codestyle-template/pull/11, merge that first

required manual interventions:
- keep nixfmt install in github CI
- remove ruff (too strict for fc-nixos, see PL-133414)

Follow-up for PL-131360

@flyingcircusio/release-managers

## Release process

- [x] Created changelog entry using `./changelog.sh`
  - no, entirely internal

## PR release workflow (internal)

- [x] PR has internal ticket
- [x] internal issue ID (PL-…) part of branch name
- [x] internal issue ID mentioned in PR description text
- [x] ticket is on Platform agile board
- [x] ticket state set to *Pull request ready*
- [x] if ticket is more urgent than within the next few days, directly contact a member of the Platform team

## Design notes

- [x] Provide a feature toggle if the change might need to be adjusted/reverted quickly depending on context. Consider whether the default should be `on` or `off`. Example: rate limiting.
- [x] All customer-facing features and (NixOS) options need to be discoverable from documentation. Add or update relevant documentation such that hosted and guided customers can understand it as well.

## Security implications

- [x] [Security requirements](https://wiki.flyingcircus.io/System_Development_Guideline#Security_requirement_principles_and_testing) defined? (WHERE)
  - update the internal enforced standard code checks to stay in sync with guidelines 
- [x] Security requirements tested? (EVIDENCE)
  - [x] automated tests still pass

PR #1375, opened 2025-03-27
introduce ruff pre-commit hook via codestyle-template

osnyx:
As a follow-up to #1355, I applied the same codestyle-template again. This introduces 2 major changes:

- somehow a line-length change from 79 -> 80 was missing the last time. This causes some reformatting.
- Introduce `ruff` hook: As this is not just a formatter but also a linter, some of the complaints were a bit more intricate. I applied them manually on a case-by-case base after checking context, or decided to skip them.

PL-133414

## Release process

  - no, internal only

- [x] PR has internal ticket

## Design notes

  - individual deviations from checks are marked with excludes or `# noqa`

## Security implications

  - [x] better monitoring of code quality and known problematic patterns by introducing the ruff linter
  - [x] not all suggestions are safe or sensible to apply
  - [x] changes scattered in many places of the codebase -> rely on regression tests: still passing
  - [x] suggested ruff changes evaluated and applied on a case-by-case basis


osnyx:
Marking this as a bit more urgent, because we need to limit the time window where this conflicts with other open PRs.
Depending on when this gets merged I can help with rebasing the pending PRs.

pkgs/fc/agent/fc/util/logging.py
```diff
@@ -1,6 +1,7 @@
 # This file is dual licensed under the terms of the Apache License, Version
 # 2.0, and the MIT License.  See the LICENSE file in the root of this
 # repository for complete details.
+# ruff: noqa: E741
```
ctheune: please leave it on

pkgs/fc/agent/fc/util/nixos.py
```diff
@@ -1,5 +1,7 @@
 """Helpers for interaction with the NixOS system"""
 
+# ruff: noqa: E741
```
ctheune: please remove

pkgs/fc/check-haproxy/fc/check_haproxy/haproxy.py
```diff
@@ -93,9 +93,9 @@ def filtered(self, prefix, records):
     def metrics(self, label, records):
         """Compute metrics for a set of `records`."""
         if label:
-            name = lambda metric: "{} {}".format(label, metric)
+            name = lambda metric: f"{label} {metric}"  # noqa: E731
         else:
-            name = lambda metric: metric
+            name = lambda metric: metric  # noqa: E731
```
ctheune: Move the if into the for loop and you don't need the callable binding

pkgs/fc/fix-so-rpath/fix-so-rpath.py
```diff
@@ -18,7 +18,7 @@ def run(args):
     if args.verbose:
         verbose = _verbose
     else:
-        verbose = lambda msg: None
+        verbose = lambda msg: None  # noqa: E731
```
ctheune:
```
def verbose(msg):
    pass
```

ctheune: Alternatively, make a _non_verbose method and assign the name in the if/else.

ctheune (changes requested):
Those cases where `noqa` was added are simple to fix. I don't want us to make excemptions for those.

pkgs/fc/check-haproxy/fc/check_haproxy/haproxy.py
osnyx: The lambda binding is also used below the for-loop again. but okay, I'll try.

ctheune (approved):

ctheune (dismissed):

osnyx:
I decided to merge without yet another approval, as that was just a technicality: The last semantical change was by @ctheune and had just a syntax error which I fixed. Everything else was properly reviewed.

PR #1381, opened 2025-03-31
[25.05] several test and eval fixes

osnyx:
Fix several eval issues and a few test failures.

PL-133559

## Release process

this goes against a development branch and piece by piece improves visibility of issues by fixing eval and tests.

## Security implications

  - none

### DIFF
--- a/pkgs/backy/default.nix
+++ b/pkgs/backy/default.nix
@@ -1,19 +1,25 @@
 {
   fetchFromGitHub,
-  fetchPypi,
+  lib,
+  stdenv,
   poetry2nix,
   lzo,
-  python310,
+  python312,
   mkShellNoCC,
   poetry,
   runCommand,
+  libiconv,
+  darwin,
+  rustPlatform,
+
 }@inputs:
 let
   src = fetchFromGitHub {
     owner = "flyingcircusio";
     repo = "backy";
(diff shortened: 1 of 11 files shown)


osnyx:
This has fixed eval at least. Still several tests and builds broken, but good to be merged as an intermediate step.

pkgs/backy/default.nix
```diff
@@ -1,19 +1,25 @@
...
   lzo,
-  python310,
+  python312,
   mkShellNoCC,
   poetry,
   runCommand,
+  libiconv,
+  darwin,
+  rustPlatform,
+
```
leona-ya, line 14: these are unused

leona-ya (commented):
otherwise lgtm

pkgs/backy/default.nix
osnyx: No, these are passed through as-is to https://github.com/flyingcircusio/backy/blob/9fa1a828495e14d3eb65000935022aef8083db9a/lib.nix . It's a bit crude as a construct, but I keep using the current way of backy packaging for now.

leona-ya: 👻

leona-ya (approved):

PR #1385, opened 2025-04-01
[DO NOT MERGE][HOTFIX] Ensure that IPv4 forwarding is enabled in the Docker role

sysvinit:
See commit message.

PL-133589

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Enabling indiscriminate forwarding is potentially a security risk, as this may result in hosts inadvertently forwarding packets for other hosts when this is not part of their function. However, the Docker daemon manages the iptables firewall itself, so we rely on Docker doing the right thing to prevent any unexpected surprises.
  - Hydra tests
  - Tested on a staging VM

### DIFF
--- a/nixos/roles/docker.nix
+++ b/nixos/roles/docker.nix
@@ -23,6 +23,14 @@ in
       flyingcircus.users.serviceUsers.extraGroups = [ "docker" ];
       virtualisation.docker.enable = true;
 
+      boot.kernel.sysctl = {
+        # IPv4 forwarding is enabled by default in the NixOS docker
+        # module, but as the platform overrides the upstream settings
+        # we need to in turn override the platform defaults here.
+        "net.ipv4.conf.all.forwarding" = fclib.mkOverridePlatformModule true;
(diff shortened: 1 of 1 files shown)


nixos/roles/docker.nix
```diff
@@ -23,6 +23,14 @@ in
       flyingcircus.users.serviceUsers.extraGroups = [ "docker" ];
       virtualisation.docker.enable = true;
 
+      boot.kernel.sysctl = {
+        # IPv4 forwarding is enabled by default in the NixOS docker
+        # module, but as the platform overrides the upstream settings
+        # we need to in turn override the platform defaults here.
+        "net.ipv4.conf.all.forwarding" = fclib.mkOverridePlatformModule true;
```
osnyx, line 30: This stacking of overrides design is a bit unfortunate, but definitely good enough for a hotfix. I also do not immediately see how we can do better, so we might be stuck with that design even after reconsiderations later on.

osnyx (approved):

PR #1387, opened 2025-04-01
Ensure forwarding is enabled in the Docker role

sysvinit:
This change is a regression fix for #1351, where we disabled IP forwarding by default in the platform. Forwarding is however required by the docker role, as containers may need to be able to access external services.

This change adds the missing forwarding sysctl configuration by overriding the platform defaults, and also updates the forwarding overrides in the external_net role to ensure that these roles can coexist on the same machine without causing evaluation errors.

PL-133589

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - We've previously identified enabling IP forwarding on all interfaces as a potential security problem, as this may result in hosts inadvertently forwarding packets for other hosts when this isn't intended behaviour. However, the Docker daemon actively manages the forwarding firewall itself, so we rely on Docker doing the correct thing and not generating insecure states.
  - Manually verified on a test VM, both with only the docker role and with the external_net role added.

### DIFF
--- a/nixos/roles/external_net/default.nix
+++ b/nixos/roles/external_net/default.nix
@@ -84,9 +84,13 @@ in
     flyingcircus.roles.vxlan.gateway = true;
 
     boot.kernel.sysctl = {
-      "net.ipv4.ip_forward" = fclib.mkOverridePlatformModule 1;
-      "net.ipv6.conf.all.forwarding" = fclib.mkOverridePlatformModule 1;
-      "net.ipv6.conf.default.forwarding" = fclib.mkOverridePlatformModule 1;
+      # Use manual override priorities here to avoid conflicts with
+      # the docker role.
+      "net.ipv4.ip_forward" = lib.mkOverride 70 true;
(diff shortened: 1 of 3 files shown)


nixos/roles/external_net/default.nix
```diff
@@ -84,9 +84,13 @@ in
     flyingcircus.roles.vxlan.gateway = true;
 
     boot.kernel.sysctl = {
-      "net.ipv4.ip_forward" = fclib.mkOverridePlatformModule 1;
-      "net.ipv6.conf.all.forwarding" = fclib.mkOverridePlatformModule 1;
-      "net.ipv6.conf.default.forwarding" = fclib.mkOverridePlatformModule 1;
+      # Use manual override priorities here to avoid conflicts with
```
osnyx, line 87: Verified that, somewhere down the type chain, lib.mergeOneOption is indeed at fault. So this is indeed necessary.

osnyx (approved):

PR #1404, opened 2025-04-09
Ensure that network configuration remains intact during critical path of switch-to-configuration

sysvinit:
Certain systemd units related to network configuration are configured so that they will be stopped by switch-to-configuration before systemd is reloaded, and then started again afterwards. This causes severe problems on machines with NFS mounts, as this can cause the systemd reload to fail when it attempts to stat the NFS mountpoint, which in the worst case can render the machine inoperable until a hard reboot is performed.

This change adds the `stopIfChanged = false` flag to the systemd units for ethernet link configuration and resolver and default gateway configuration to ensure that these units are restarted in-place after systemd has been reloaded during switch-to-configuration. Additionally, there is now an additional test case in the NFS test which reproduces this problem and should detect regressions in the future.

PL-133570

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Adjust the order of operations of service restarts during configuration switches to improve reliability and robustness.
  - Additional Hydra test case.
  - Manually verified on test VMs that upgrading to this change will use the restart settings of the new units.

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -538,6 +538,7 @@ in
                 pkgs.jq
                 pkgs.util-linux
               ];
+              stopIfChanged = false;
               script =
                 ''
                   set -e
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -511,6 +511,11 @@ in
           systemd-sysctl.restartTriggers = lib.mkIf (!isNull fclib.underlay) [
             config.environment.etc."sysctl.d/70-fcio-underlay.conf".source
           ];
+          # Stop+starting the network can break
+          # switch-to-configuration on NFS clients by reloading the
+          # systemd config when the NFS server is unreachable. Do an
+          # in-place restart after systemd has reloaded instead.
+          network-setup.stopIfChanged = fclib.mkOverrideUpstreamModule false;
         }
         //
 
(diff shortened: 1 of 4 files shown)


nixos/platform/network.nix
```diff
@@ -511,6 +511,11 @@ in
           systemd-sysctl.restartTriggers = lib.mkIf (!isNull fclib.underlay) [
             config.environment.etc."sysctl.d/70-fcio-underlay.conf".source
           ];
+          # Stop+starting the network can break
+          # switch-to-configuration on NFS clients by reloading the
+          # systemd config when the NFS server is unreachable. Do an
+          # in-place restart after systemd has reloaded instead.
+          network-setup.stopIfChanged = fclib.mkOverrideUpstreamModule false;
```
sysvinit, line 518: This should be merged into our nixpkgs fork rather than overridden here, however this works as an immediate quick fix.

ctheune (approved):
Ok, this does feel awkward but I think it's the right solution for now.

nixos/platform/network.nix
osnyx: Has this been done yet?

sysvinit: Not to my knowledge, I've been busy fixing other things in the mean time.

PR #1417, opened 2025-04-16
Don't set bridge options on routed networks in FRR config

sysvinit:
In the FRR config we set the `advertise-svi-ip` option for virtual networks using VXLAN, to ensure that the corresponding bridge interfaces are made visible in the EVPN control plane if they have IP addresses assigned. However, we were also setting this option for the routed pub network, which is an invalid configuration (layer 2 config applied to layer 3 network), which causes FRR to log an error message and then fail to process routes in the pub VRF properly. This can cause VMs with routed pub interfaces to lose connectivity.

This change ensures that the `advertise-svi-ip` option is only set for non-routed (i.e. bridged/switched) networks.

PL-133325

## Release process

- [ ] ~~Created changelog entry using `./changelog.sh`~~

- [x] PR has internal ticket

## Design notes

## Security implications

  - Configuration correctness affecting network availability.
  - Verified manually on hosts in dev.

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -454,7 +454,7 @@ in
                   " advertise-svi-ip"
                   "exit-vni"
                 ]
-              ) vxlanInterfaces
+              ) (filter (i: !i.routed) vxlanInterfaces)
             }
            exit-address-family
            !


leona-ya (approved):
I cannot judge the frr part, but the nix part looks fine to me

sysvinit:
Merging this manually to get it in to tomorrow's release.

PR #1429, opened 2025-04-24
fc.agent: use `boot` to switch between NixOS release upgrades

ctheune:
Fixes PL-133570

## Release process

- [x] PR has internal ticket

## Design notes

I strongly considered providing a feature toggle. However, due to the nature of the change, a feature toggle complicates this logic even further and IMHO increases the risk.

## Security implications

improve availability/reliability

manual testing and unit tests

### DIFF
--- a/pkgs/fc/agent/fc/util/channel.py
+++ b/pkgs/fc/agent/fc/util/channel.py
@@ -1,5 +1,8 @@
 import os
 import os.path as p
+import subprocess
+import time
+from pathlib import Path
 
 from fc.util import nixos
 from fc.util.nixos import RE_FC_CHANNEL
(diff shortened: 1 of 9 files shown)


pkgs/fc/agent/fc/util/channel.py
```diff
@@ -116,7 +119,38 @@ def switch(self, lazy=True, show_trace=False):
...
+                next_release=next_release,
+            )
+            while reboot_delay:
+                self.log.warn(
+                    "reboot-scheduled",
+                    _replace_msg=f"WILL REBOOT IN {reboot_delay} SECONDS. PRESS Ctrl-C TO ABORT.",
+                )
+                time.sleep(1)
+                reboot_delay -= 1
+            nixos.switch_to_system(self.system_path, lazy, "boot", self.log)
```
leona-ya: This should be
```
if nixos.switch_to_system(switch_path, lazy, "boot", self.log) is False:
  return False
```
to prevent reboot when a reboot when no config change happened. This is a case that should not happen tho.

PR #1448, opened 2025-04-30
s3users: error when unknown error occurs

leona-ya:
This is a safeguard against unexpected errors happen in the rgw user list / user info calls leading to a re-creation of the user with a new secret key.

PL-133656

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Availability: This decreases availability as it stops handling new users when an unknown error occurs. This is intended to ensure a clean state. We have a stamp file for this, so a failure gets noticed.
  - Consistency: Fail stop behavior
  - see above
  - tested user creation and key rotation in dev
  - unit tests

### DIFF
--- a/pkgs/fc/agent/fc/manage/s3users.py
+++ b/pkgs/fc/agent/fc/manage/s3users.py
@@ -9,6 +9,7 @@ import argparse
 import datetime
 import json
 import logging
+import subprocess
 import sys
 from dataclasses import dataclass
 from pathlib import Path
@@ -65,6 +66,12 @@ class RGWState:
         self.uid = uid
 
     def update(self, previously_deleted=False):
+        def check_user_doesnt_exist(code, stdout, stderr):
+            return (
+                code == 22
+                and b"could not fetch user info: no user info saved" in stderr
+            )
+
         try:
             state = run.json.radosgw_admin(
                 "user",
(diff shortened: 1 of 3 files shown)


ctheune (approved):

pkgs/fc/agent/fc/manage/s3users.py
```diff
@@ -65,27 +66,35 @@ def __init__(self, uid):
         self.uid = uid
 
     def update(self, previously_deleted=False):
+        def check_user_doesnt_exist(code, stdout, stderr):
+            return (
+                code == 22
+                and b"could not fetch user info: no user info saved" in stderr
```
ctheune, line 72: I don't think this has a strip() before getting the stderr passed and the real output has a `\n` in it:

```
>>> x = subprocess.Popen(["radosgw-admin", "user", "info", "--uid=asdf"], stderr=subprocess.PIPE, stdout=subprocess.PIPE)
>>>
>>> x.communicate()
(b'', b'could not fetch user info: no user info saved\n')
```

ctheune (changes requested):
```

pkgs/fc/agent/fc/manage/s3users.py
leona-ya: This is actually okay, since this is matched with `in`, but we could improve the check by matching on the exact string + `.strip()` if you want.

ctheune: you're right. i was misreading this as an equals match.

ctheune (approved):

PR #1451, opened 2025-04-30
fc.check-postfix: regenerate Cargo.lock for compatability with fetchCargoVendor

leona-ya:

- [x] PR has internal ticket

## Design notes

## Security implications

  - Availability: Regenerated Cargo lockfile within the boundaries defined by Cargo.toml
  Hydra tests, see above


leona-ya:
Note that `useFetchCargoVendor = true;` is not needed in 25.05 as it defaults to `true` there and is not removed only for backport compatability with 24.11.

osnyx:
> Note that useFetchCargoVendor = true; is not needed in 25.05 

I probably started the migration when this was not the default yet.

osnyx (approved):

PR #1476, opened 2025-05-13
[PL-133636] implement k3 monitoring

PhilTaken:

PL-133636

I picked risk:4 and urgency:3 labels due to the fact that the absence of k3s event monitoring can cause SLA relevant issues in production instances.
This PR should be considered a draft or proposal rather than a finished implementation due to some issues mentioned below.

The diff appears much larger than it is due to formatting changes.
This is due to introducing a `lib.mkMerge`  in line 371 (https://github.com/flyingcircusio/fc-nixos/compare/PL-133636/k3s-monitoring?expand=1#diff-4341733a592ffa27bc94c82a794a43aef9beb09ff47561caf6179449b23717a4R371).

Notable changes are an added check in https://github.com/flyingcircusio/fc-nixos/compare/PL-133636/k3s-monitoring?expand=1#diff-4341733a592ffa27bc94c82a794a43aef9beb09ff47561caf6179449b23717a4R436 and the alloy config in https://github.com/flyingcircusio/fc-nixos/compare/PL-133636/k3s-monitoring?expand=1#diff-4341733a592ffa27bc94c82a794a43aef9beb09ff47561caf6179449b23717a4R711.

Notable problems/ why this PR is not a good implementation:
1. ~~The sensu check will consume about 300M of /tmp space every time it runs.
This is due to the fact that k3s copies the entire k3s binary (296M) as well as a few other things into the specified $K3S_DATA_DIR.
This directory needs to exist however since k3s attempts to create a lockfile during operation and will abort if it can't - for example when the $K3S_DATA_DIR points at /var/empty.~~
The K3S_DATA_DIR was moved to /var/tmp so that is not an issue anymore

2. Events only persist in Kubernetes for about one hour which will affect the state of this check. Some events might be triggered again but if they aren't then the alert will show as resolved and will effectively only appear for Ops if it triggers a new ticket.
Querying the events via `logcli` from a Loki instance after exporting them via Alloy (using the kubernetes_events source) instead of calling the k3s binary as part of the Sensu check would be a robust solution but does require a Loki instance per resource group for the check to even work.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

- improved monitoring of critical infrastructure
- new sensu check

### DIFF
--- a/changelog.d/20250513_113952_PL-133636-k3s-monitoring_scriv.md
+++ b/changelog.d/20250513_113952_PL-133636-k3s-monitoring_scriv.md
@@ -1,0 +1,20 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+-
+
+### NixOS XX.XX platform
+
+- Kubernetes events are now shipped to both our custom monitoring platform as well as RG-specific loki instances to be used in grafana dashboards etc. (PL-133636)
--- a/nixos/roles/k3s/server.nix
+++ b/nixos/roles/k3s/server.nix
@@ -17,6 +17,8 @@ let
   srvFQDN = "${config.networking.hostName}.fcio.net";
   nodeAddress = head fclib.network.srv.v4.addresses;
 
+  lokiServer = fclib.findOneService "loki-collector";
+
   # We allow frontend access to the dashboard at the moment
   # via Nginx. The dashboard can be accessed by multiple names.
   # Unlike with the old kubernetes roles, the API is not public here.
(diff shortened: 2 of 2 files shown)


PhilTaken:
Since we cannot require a Loki setup for every customer, we need to make sure that the directory has alerting configured for these events such that we have an open ticket even if the k3s event is not triggered again.

changelog.d/20250513_113952_PL-133636-k3s-monitoring_scriv.md
```diff
@@ -0,0 +1,20 @@
...
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- improve auditing of kubernetes clusters
```
ctheune: This is not impact. 

Impact means "when rolling out this change, your production cluster will explode in THIS way.". Typically this means: rolling this out, you will experience a downtime of 5 minutes in service foobar.

nixos/roles/k3s/server.nix
```diff
@@ -17,6 +17,8 @@ let
   srvFQDN = "${config.networking.hostName}.fcio.net";
   nodeAddress = head fclib.network.srv.v4.addresses;
 
+  lokiServer = fclib.findOneService "loki-collector";
```
ctheune, line 20: According to @PhilTaken, @sysvinit might have some feedback here about the use of findOneService?

ctheune (changes requested):
You wanted feedback from @sysvinit - go get it! :)

nixos/roles/k3s/server.nix
PhilTaken: The mechanism of deriving if a Loki endpoint is present in the current RG is subject to change but due to it's noninvasive nature in this module, adjustments can be made as part of that increment without much hassle.
Additionally, since the specific implementation is still awaiting a conceptual phase with no clear timeline on when it might be implemented, I'd argue that this PR should not await it and instead be merged as-is.

ctheune (approved):

PR #1499, opened 2025-05-22
[PL-133581] fix roles.loki options and configuration templating

PhilTaken:

PL-133581

Few adjustments to the loki role.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

- not security relevant

### DIFF
--- a/nixos/roles/loki.nix
+++ b/nixos/roles/loki.nix
@@ -20,7 +20,7 @@ let
         backend = mkOption {
           type = enum [
             "filesystem"
-            "s3"
+            "aws"
           ];
         };
         schemaVersion = mkOption {
@@ -135,10 +135,10 @@ in
             filesystem.directory = "/var/lib/loki/chunk-store";
           }
           // lib.optionalAttrs (cfg.s3.enable) {
-            s3 = {
+            aws = {
               # authentication configured separately
               endpoint = cfg.s3.endpoint;
-              bucketNames = cfg.s3.bucketNames;
+              bucketNames = cfg.s3.bucketName;
               s3forcepathstyle = true;
             };
           };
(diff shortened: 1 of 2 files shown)


nixos/roles/loki.nix
```diff
@@ -135,10 +135,10 @@ in
             filesystem.directory = "/var/lib/loki/chunk-store";
           }
           // lib.optionalAttrs (cfg.s3.enable) {
-            s3 = {
+            aws = {
```
sysvinit, line 138: Is there any particular reason to use `aws_storage_config` here instead of `s3_storage_config`? As far as I understand from the docs these should be equivalent in our use case.

PhilTaken: that was mostly due to 

```
$ nix log /nix/store/dbws9wh7zsrqw27b4qgyj2p3mqbh2g09-validate-loki-conf.drv
failed parsing config: /nix/store/7ycfb1dxx6hmglipyqdmvxk8629dcdhd-loki-config.json: yaml: unmarshal errors:
  line 52: field s3 not found in type storage.Config. Use `-config.expand-env=true` flag if you want to expand environment variables in your config file

$
```

I had assumed the s3 storage_config type was deprecated but going through the docs in the loki repo that does not appear to be the case. I'm not quite sure why the config validation doesn't recognize the s3 storage config type then

sysvinit: I've read through what I think is the relevant code in Loki, and it looks like there's some post-processing of the config yaml beyond just deserialising it into a golang struct. Plausibly a Loki bug, but I won't object to changing our side to fix the config validation error.

sysvinit (approved):

PR #1508, opened 2025-05-27
fc.agent: improve test coverage for `release_path`

osnyx:
Provide test coverage for the code path responsible for PL-133713. Also corrects some type annotations.

The switch from `os.readlink` to `os.path.realpath` was done to simplify testing: `readlink` *requires* the target to be a link, while `realpath` only *allows* it to be. We do not have to enforce system paths to be links though, this is more an implementation decision of NixOS.

PL-133713

## Release process

  - no, unreleased release, not customer visible

- [x] PR has internal ticket

## Design notes

## Security implications

  - `release_path` needs test coverage
  - [ ] tests still pass
  - [x] htmlcov shows coverage for `release_path`
  - [x] investigated the involved return types of functions

### DIFF
--- a/pkgs/fc/agent/src/fc/maintenance/activity/update.py
+++ b/pkgs/fc/agent/src/fc/maintenance/activity/update.py
@@ -35,7 +35,10 @@ class UpdateActivity(Activity):
     """
 
     def __init__(
-        self, next_channel_url: str, next_environment: str = None, log=_log
+        self,
+        next_channel_url: str,
+        next_environment: str | None = None,
+        log=_log,
     ):
         super().__init__()
         self.next_environment = next_environment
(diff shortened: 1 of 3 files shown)


leona-ya:
I'm not the right reviewer for this change. I don't know much about the agent code.

PhilTaken:
likewise

PhilTaken:
I would however recommend type-annotating as much as possible here (or rather in general) so that static analysis can catch these kinds of errors

osnyx:
Regarding type annotations, I generally agree. But in these specific cases, the relevant types are defined by standard library functions like `os.readlink`/ `os.path.realpath`. As long as we do not restrict these any further, we better use the type signatures exposed by them.

My MyPy editor integration provides signatures for these standard library functions. Right now these are `typing.AnyString`, but as this type will be deprecated in Python 3.18 I decided not to explicitly add it to our code and make it just follow the changes with future python library versions.

ctheune (approved):

PR #1509, opened 2025-05-27
Remove timer indirection for keepalived service

sysvinit:
The upstream NixOS module for keepalived uses a timer to defer the start of the keepalived process at system boot to prevent the daemon from immediately entering the master state before the system is fully initialised. However, this means that if keepalived crashes, then it will not be restarted the next time switch-to-configuration is run, as the service is not wanted by multi-user.target. Restarting keepalived at a later time then requires manual intervention.

This change removes this indirection through the timer and starts keepalived directly through multi-user.target, so that a crashed keepalived will be restarted automatically.

PL-133434

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Improving robustness of the router role and the ability to recover from transient failures.
  - Manually verified on the dev routers running 24.11.

### DIFF
--- a/nixos/roles/router/default.nix
+++ b/nixos/roles/router/default.nix
@@ -265,14 +265,28 @@ in
 
     flyingcircus.firewall.enableSrvRgFirewall = false;
 
-    systemd.services = listToAttrs (
-      lib.forEach (filter (iface: iface.policy == "vxlan") gatewayInterfaces) (
-        iface:
-        lib.nameValuePair "network-bridge-suppress-flooding-${iface.link}" {
-          enable = fclib.mkPlatform false;
-        }
-      )
-    );
+    systemd.services =
+      (listToAttrs (
+        lib.forEach (filter (iface: iface.policy == "vxlan") gatewayInterfaces) (
+          iface:
+          lib.nameValuePair "network-bridge-suppress-flooding-${iface.link}" {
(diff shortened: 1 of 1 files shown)


osnyx:
> The upstream NixOS module for keepalived uses a timer to defer the start of the keepalived process at system boot to prevent the daemon from immediately entering the master state

So do you say this reasoning is an unnecessary one? Or is this approach just a compromise of not introducing too much complexity and hence being okay with keepalived immediately entering the master state?

Also, isn't that something better suited as an upstream contribution then?

sysvinit:
> So do you say this reasoning is an unnecessary one? Or is this approach just a compromise of not introducing too much complexity and hence being okay with keepalived immediately entering the master state?

keepalived's initial state is configurable, and we set it to the backup state in all our configs, so the extra systemd timer (which only waits five seconds anyway) is a redundant safety rail. The extra indirection with the timer is less robust under edge cases than just starting the daemon directly, as it can mean that keepalived can fail without being recovered and we may not notice that it's failed.

> Also, isn't that something better suited as an upstream contribution then?

The timer probably makes sense if you're using a keepalived configuration which initially starts in the master state. I don't really want to immediately take this upstream as other people almost certainly have use cases which we aren't aware of where this is useful behaviour.

osnyx (approved):

PR #1517, opened 2025-05-28
[PL-133563] services/redis: get rid of IFD for generating the Redis password

Ma27:

* Restrict Redis server names (i.e. `<name>` in `services.redis.servers.<name>`) to alphanumeric chars, underscores and dashes. Everything else gives an assertion error. This is because the name will be used for the envsubst in Telegraf. Dashes will be escaped to `__`. Using consecutive underscores in server names is also prohibited to avoid ambiguities.

  This would require people to use multiple redis instances (which isn't documented so far in our role), passwords and special chars in the name to run into problems. I'd suggest to backport it anyways @osnyx 

* Apply @leona-ya's patch to no longer use IFD to generate the Redis password: this means that `services.redis.servers."".requirePass` can no longer be used to obtain the Redis password at eval-time, with no alternatives (on purpose, kinda).

* use unittest-style assertions for the newly written testcases to get better diffs in case of errors (its own commit to keep the other parts backportable).

## Release process

- [x] PR has internal ticket

- [x] set urgency and risk labels: urgency=3 because of the restriction for naming of servers. We may want to split this, though.

## Design notes

## Security implications

  - Principle of Least privilege: the Redis password is no longer kept in the world-readable store, so it's not readable by each process anymore.
  - Availability: we don't risk broken configurations by the password derivation being garbage-collected (this happened in other cases with the same pattern already causing a rotated password and service-restart mid-day).
  - Implemented VM tests making sure redis gets configured as expected with in each case (i.e. no password; password from option; password from file).

### DIFF
--- a/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
+++ b/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
@@ -1,0 +1,25 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- If you access `services.redis."".password` in your NixOS config, please change this to
+  `flyingcircus.services.redis.password`.
+
+### NixOS XX.XX platform
+
+- redis: restructure internal password handling
+  The password file /etc/local/redis/password now gets written as systemd ExecStartPre. (PL-133653)
+
+- telegraf: Add new option `environmentVariablesFromFile` that allows users to store
+  passwords in files and using them as substituted environment variable in a telegraf input. (PL-133563)
--- a/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
+++ b/changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
@@ -13,13 +13,11 @@ branches.
 
 ### Impact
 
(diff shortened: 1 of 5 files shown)


changelog.d/20250324_095007_PL-133563-redis-remove-ifd_scriv.md
```diff
@@ -0,0 +1,22 @@
...
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- If you set `services.redis."".requirePassFile` in your NixOS config, please use
+  `flyingcircus.services.redis.password` instead.
```
Ma27: I guess this should also mention that the password can no longer be read from this.

ctheune: This isn't impact.

nixos/services/redis.nix
```diff
@@ -99,7 +88,7 @@ let
 
   telegrafInputs =
     let
-      mkPassword = name: getRedisPassword name (lib.const "\${REDIS_PASSWORD_${name}}");
+      mkPassword = name: getRedisPassword name (lib.const "\${REDIS_PASSWORD_${escape name}}");
```
ctheune: we'll need a refactoring here. I understand why you wanted to fold the if/else into the getRedisPassword function, however, this turns out to be too much inside/out wrapping things ...

ctheune (changes requested):
generally fine, but please refactor (and rename) the  getRedisPassword function

ctheune (approved):
Thanks!

PR #1520, opened 2025-05-30
xfs: enable default features

dhnasa:
This upgrades the root device to support bigtime, inobtcount and nrext64 (bigtime and inobtcount are already the default for new machines. nrext64 is supported since kernel 6.6).
Tmp will be formatted with the default options of mkfs.xfs.
Also copies enc/qemu data from tmp and cidata before formatting.

PL-133321, PL-130365, PL-133311

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

### DIFF
--- a/changelog.d/20250528_185829_PL-133321-xfs-bigtime_scriv.md
+++ b/changelog.d/20250528_185829_PL-133321-xfs-bigtime_scriv.md
@@ -1,0 +1,27 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+
+### NixOS XX.XX platform
+
+- Add a mechanism to upgrade XFS flags over time. Initially this will cause
+  `bigtime`, `inobtcount` and `nrext64` flags to be set. With `bigtime` set VMs
+  from this release on will consistently support file times beyond 2038, even if
+  they were bootstrapped on older releases. (PL-133321, PL-130365)
+
+- Prepare VMs to read ENC data (the config management metadata) seeded from the
+  host from the separate `cidata` volume in the future. This allows disabling or
+  reconfiguring /tmp to use tmpfs without breaking our configuration management.
+  (PL-133311)
--- a/nixos/infrastructure/flyingcircus-physical.nix
+++ b/nixos/infrastructure/flyingcircus-physical.nix
@@ -102,6 +102,16 @@ mkIf (cfg.infrastructureModule == "flyingcircus-physical") (
         };
       };
 
+      # Not formatting tmp for now because I dont want to refer to it by
+      # the filesystem label and some machines use lvm and others dont.
+      flyingcircus.initrd.upgradeXFS = {
+        "/dev/disk/by-label/root" = [
+          "bigtime=1"
+          "inobtcount=1"
+          "nrext64=1"
+        ];
+      };
+
       networking = {
         domain = "fcio.net";
         hostName = fclib.mkPlatform (attrByPath [ "name" ] "default" cfg.enc);
--- a/nixos/infrastructure/flyingcircus-virtual.nix
+++ b/nixos/infrastructure/flyingcircus-virtual.nix
@@ -83,10 +83,24 @@ mkIf (cfg.infrastructureModule == "flyingcircus") {
       fsType = "xfs";
     };
     "/tmp" = {
-      device = "/dev/disk/by-label/tmp";
+      device = "/dev/disk/by-partlabel/tmp";
       fsType = "xfs";
       noCheck = true;
     };
+  };
+
+  flyingcircus.initrd = {
+    upgradeXFS = {
+      "/dev/disk/by-label/root" = [
+        "bigtime=1"
+        "inobtcount=1"
+        "nrext64=1"
+      ];
+    };
+    formatXFS = {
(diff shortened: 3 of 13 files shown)


nixos/infrastructure/flyingcircus-virtual.nix
```diff
@@ -83,12 +83,24 @@ mkIf (cfg.infrastructureModule == "flyingcircus") {
...
 
+  flyingcircus.initrd = {
+    upgradeXFS = {
+      "/dev/disk/by-label/root" = [ "bigtime=1" ];
+    };
+    # XXX: remove nrext=0 once all VMs are on kernel 6.6+
+    formatXFS = {
+      "/dev/disk/by-partlabel/tmp" =
+        "-L tmp -q -K -m crc=1,finobt=1,bigtime=1 -i nrext64=0 -d su=4m,sw=1";
+    };
```
sysvinit: Is there a reason to do this here in fc-nixos instead of in fc.qemu? As far as I understand, the tmp partition is reformatted on every (cold) boot by fc.qemu, so doing it again here in the initrd as well would be redundant.

Additionally, `bigtime=1` has been set by default in mkfs.xfs for a few years now (compare the settings files for kernels 5.10 and 5.15 in xfsprogs), so this might have fixed itself already without us needing to do anything. I'm fairly sure this bit us once already with an old VM after upgrading the hardware from 21.05 to 24.11.

(Speaking of hardware, we might want to set `flyingcircus.initrd.upgradeXFS` on physical machines so that they get the filesystem upgrade as well.)

sysvinit: Another thought, regarding the tmp reformatting: hard-baked XFS feature flag settings spreading into more corners of the platform makes me somewhat uneasy, as it's another place we have to remember to update in future releases when new feature flags come along.

leona-ya: The idea is that `/tmp` isn't reformatted on every boot cycle in fc.qemu. We don't know the kernel version inside a VM in fc.qemu, so this is quite error-prone. Here we know the kernel version of the VM, so this allows us to use new XFS features more agile without waiting for the full fleet to be on a newer kernel.

leona-ya: So also this will be the primary place XFS flags are set

nixos/infrastructure/flyingcircus-virtual.nix
```diff
@@ -83,12 +83,24 @@ mkIf (cfg.infrastructureModule == "flyingcircus") {
...
       fsType = "xfs";
       noCheck = true;
     };
   };
 
+  flyingcircus.initrd = {
+    upgradeXFS = {
+      "/dev/disk/by-label/root" = [ "bigtime=1" ];
+    };
+    # XXX: remove nrext=0 once all VMs are on kernel 6.6+
```
leona-ya: This confuses me. As 25.05 uses Kernel 6.6+,  we should be able to use this here. Also the requirement _all VMs_ is not necessary anymore in this code path

dhnasa: fixed

nixos/infrastructure/flyingcircus-virtual.nix
dhnasa: switched to the default mkfs.xfs arguments and also upgraded the root partition of physical machines

changelog.d/20250528_185829_PL-133321-xfs-bigtime_scriv.md
```diff
@@ -0,0 +1,20 @@
...
+
+-->
+
+### Impact
+
+
+### NixOS XX.XX platform
+
+- Copy enc data from tmp and cidata and upgrade XFS on root and tmp to support
+  dates after 2038. (PL-133321, PL-130365, PL-133311)
```
ctheune: - Add a mechanism to upgrade XFS flags over time. Initially this will cause `bigtime`, `inobtcount` and `nrext64` flags to be set. With `bigtime` set VMs from this release on will consistently support file times beyon 2038, even if they were bootstrapped on older releases. (PL...)

- Prepare VMs to read ENC data (the config management metadata) seeded from the host from the separate `cidata` volume in the future. This allows disabling or reconfiguring /tmp to use tmpfs without breaking our configuration management. (PL...)

nixos/infrastructure/flyingcircus-physical.nix
```diff
@@ -102,6 +102,12 @@ mkIf (cfg.infrastructureModule == "flyingcircus-physical") (
         };
       };
 
+      # Not formatting tmp for now because I dont want to refer to it by
+      # the filesystem label and some machines use lvm and others dont.
+      flyingcircus.initrd.upgradeXFS = {
+        "/dev/disk/by-label/root" = [ "bigtime=1" ];
```
ctheune: inconsistent?

pkgs/fc/agent/src/fc/maintenance/maintenance.py
```diff
@@ -74,9 +73,6 @@ def request_reboot_for_kvm_environment(log) -> Optional[Request]:
 
     qemu_state_dir = Path(QEMU_STATE_DIR)
 
-    def cold_state_file(stem):
-        return Path(KVM_SEED_DIR) / f"qemu-{stem}-booted"
-
```
ctheune, line 79: Yeah, I agree with this - there is something weird where moving this off of tmp is not caused immediately after all reboot currently, so this is currently somewhat sloppy. We're going to be better with the new code and I don't need to make the time in between more perfect than it was previously ... no second order fixes here.

ctheune (changes requested):
changelog noch anpassen, danach ok

ctheune (approved):

PR #1528, opened 2025-06-03
[PL-133663] build fcsearch inputs on hydra

PhilTaken:

[PL-133663](https://yt.flyingcircus.io/issue/PL-133663)

The constructed test is very minimal, would appreciate specific feedback on that regarding missing module imports / incorrectly set options.
I would almost prefer generating the options from one of the release images instead.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

- adds two new hydra jobs, does not impact security
-

### DIFF
--- a/release/default.nix
+++ b/release/default.nix
@@ -200,6 +200,95 @@ let
 
   doc = {
     roles = platformRoleDoc;
+
+    options =
+      let
+        imgArgs = {
+          nixpkgs = nixpkgs_;
+          version = "${version}${versionSuffix}";
+          channelSources = combinedSources;
+          configFile = ../nixos/etc_nixos_local.nix;
+          contents = initialVMContents;
+        };
+
+        fc_eval = import "${nixpkgs_}/nixos/lib/eval-config.nix" {
+          inherit system;
+          modules = [
+            "${nixpkgs_}/nixos/modules/virtualisation/qemu-vm.nix"
+
+            (import ./vm-image.nix imgArgs)
+            ../nixos
+            ../nixos/roles
+
+            {
+              networking.hostName = "options";
+              mailserver.fqdn = "test.fcio.net";
+            }
+          ];
+        };
+      in
+      lib.hydraJob (
+        (pkgs.nixosOptionsDoc {
+          inherit (fc_eval) options;
+          warningsAreErrors = false;
+        }).optionsJSON
+      );
+
+    packages =
+      let
+        isValid =
+          d:
+          let
+            r = builtins.tryEval (
+              lib.isDerivation d
+              && !(lib.attrByPath [ "meta" "broken" ] false d)
+              && builtins.seq d.name true
+              && d ? outputs
+            );
+          in
+          r.success && r.value;
+        validPkgs = lib.filterAttrs (_: v: isValid v);
+
+        readPackages =
+          system: drvs:
+          lib.mapAttrs (
+            attribute_name: drv:
+            (
+              {
+                entry_type = "package";
+                attribute_name = attribute_name;
+                system = system;
+                name = drv.name;
+                # TODO consider using `builtins.parseDrvName`
+                version = drv.version or "";
+                outputs = drv.outputs;
+                # paths = builtins.listToAttrs ( map (output: {name = output; value = drv.${output};}) drv.outputs );
+                default_output = drv.outputName;
+              }
+              // lib.optionalAttrs (drv ? meta.homepage) {
+                inherit (drv.meta) homepage;
+              }
+              // lib.optionalAttrs (drv ? meta.description) {
+                inherit (drv.meta) description;
+              }
+              // lib.optionalAttrs (drv ? meta.longDescription) {
+                inherit (drv.meta) longDescription;
+              }
+              // lib.optionalAttrs (drv ? meta.license) {
+                inherit (drv.meta) license;
+              }
+            )
+          ) (validPkgs drvs);
+        packages_json = builtins.toFile "fc-search-packages.json" (
+          builtins.toJSON (readPackages system pkgs)
+        );
+      in
+      lib.hydraJob (
+        pkgs.runCommand "fc-search-packages" { buildInputs = [ pkgs.jq ]; } ''
+          mkdir -p $out
+          cat ${packages_json} | jq > $out/packages.json
+        ''
+      );
   };
 
   jobs = {
--- a/changelog.d/20250603_100901_FC-43799_build-fcsearch-input-on-hydra_scriv.md
+++ b/changelog.d/20250603_100901_FC-43799_build-fcsearch-input-on-hydra_scriv.md
@@ -1,0 +1,20 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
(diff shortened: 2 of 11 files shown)


ctheune:
@Ma27 @leona-ya would you mind giving @PhilTaken feedback here? That's out of my depth. I did support the original idea of baking this into the release instead of having the search server download and evaluate the channels every 10 minutes ...

Ma27:
Allocated a timeslot on friday.

release/default.nix
```diff
@@ -200,6 +200,96 @@ let
...
+          contents = initialVMContents;
+        };
+
+        fc_eval = import "${nixpkgs_}/nixos/lib/eval-config.nix" {
+          inherit system;
+          modules = [
+            "${nixpkgs_}/nixos/modules/virtualisation/qemu-vm.nix"
+
+            (import ./vm-image.nix imgArgs)
+            (import version_nix { })
```
Ma27: I'm not sure if it's truly necessary to import these?
We only care about the options and we should really not make the manual depend on the config IMHO.

release/default.nix
```diff
@@ -200,6 +200,96 @@ let
...
+              networking.hostName = "options";
+              mailserver.fqdn = "test.fcio.net";
+            }
+          ];
+        };
+      in
+      lib.hydraJob (
+        (pkgs.nixosOptionsDoc {
+          inherit (fc_eval) options;
+          warningsAreErrors = false;
```
Ma27, line 233: After checking the sources, it seems as if this allows both type and description to be empty.

I do wonder if we actually want to force having both considering that our ideal is to have user-discoverable documentation of our platform?

EDIT: OK, we have more options undocumnented than expected. Would be a nice follow-up then IMHO.

release/default.nix
```diff
@@ -200,6 +200,96 @@ let
...
+            (
+              {
+                entry_type = "package";
+                attribute_name = attribute_name;
+                system = system;
+                name = drv.name;
+                # TODO consider using `builtins.parseDrvName`
+                version = drv.version or "";
+                outputs = drv.outputs;
+                # paths = builtins.listToAttrs ( map (output: {name = output; value = drv.${output};}) drv.outputs );
```
Ma27, line 265: Random thought: it _might_ be useful in some cases to also list the outputs of a package in the search results?

(Obviously a follow-up and non-blocking)

release/default.nix
```diff
@@ -200,6 +200,96 @@ let
...
+              // lib.optionalAttrs (drv ? meta.longDescription) {
+                inherit (drv.meta) longDescription;
+              }
+              // lib.optionalAttrs (drv ? meta.license) {
+                inherit (drv.meta) license;
+              }
+            )
+          ) (validPkgs drvs);
+        packages_json = builtins.toFile "fc-search-packages.json" (
+          builtins.toJSON (readPackages system pkgs)
```
Ma27, line 283: Maybe use a package-set with `allowAliases = false;` here?

These aliases are cleaned up somewhat regularly, so it means more operational work for us on upgrades if people use them. Also, this would mean less noise in the eval-error tab on Hydra which is also used to spot any mistakes we do in platform engineering.

An excerpt from what I get when evaluating this by hand:

```
evaluation warning: lxd has been renamed to lxd-lts
evaluation warning: lxd-unwrapped has been renamed to lxd-unwrapped-lts
evaluation warning: marwaita-manjaro has been renamed to marwaita-teal
evaluation warning: marwaita-peppermint has been renamed to marwaita-red
evaluation warning: marwaita-pop_os has been renamed to marwaita-yellow
evaluation warning: marwaita-ubuntu has been renamed to marwaita-orange
evaluation warning: nixfmt was renamed to nixfmt-classic. The nixfmt attribute may be used for the new RFC 166-style formatter in the future, which is currently available as nixfmt-rfc-style
```

release/default.nix
```diff
@@ -200,6 +200,96 @@ let
...
+        );
```
Ma27, line 284: An additional note (new line for another thread). This introduces more IFD into our evaluation:

```
copying 3 paths...
copying path '/nix/store/p2hh8i8dx94l51iq5whw3qg40jvy453q-openjdk-headless-8u322-ga-jre' from 'https://s3.whq.fcio.net/hydra'...
copying path '/nix/store/5i1mf8784vx7dzh030ixggrx93j3fs9j-openjdk-headless-8u322-ga' from 'https://s3.whq.fcio.net/hydra'...
copying path '/nix/store/yj365yr01p6yp2axj943b4l8ngzzxvkw-graylog-3.3.16' from 'https://s3.whq.fcio.net/hydra'...
```

This __will be another slowdown__ of our Hydra evaluation because we now download an OpenJDK and a graylog during evaluation.

I'm not sure if there's much we can do, apparently IFD happens quite early when importing nixpkgs:
```
       … while calling the 'import' builtin
         at /home/ma27/fc-nixos/release/default.nix:20:6:
           19|
           20| with import "${nixpkgs_}/pkgs/top-level/release-lib.nix" {
             |      ^
           21|   inherit supportedSystems scrubJobs;

       … while realising the context of path '/nix/store/nvwpabrlvggangsmhi0qz29clr923max-nixpkgs-c1b76eb244f/pkgs/top-level/release-lib.nix'

       error: cannot build '/nix/store/si0n0kzm5ph89h8mfv0di25nqj5y5bhn-nixpkgs-c1b76eb244f.drv^out' during evaluation because the option 'allow-import-from-derivation' is disabled
```

Yes, this is probably an issue with our graylog packaging (cc @dpausp ), but AFAICS graylog isn't part of important packages and doesn't get evaluated on Hydra yet.

Personally, I'd suggest to merge this and fix the Graylog part in a follow-up. But I'd like to avoid accumulating new sources of slowing down evaluations after we made it fast last fall :)

Ma27 (commented):
Implementation looks good, just a few minor things I have.

Please keep in mind that I don't know the fc-search codebase too well, so if things don't make sense, please say so ;)

release/default.nix
PhilTaken: Good point, will remove!

release/default.nix
PhilTaken: Unfortunately this isn't really feasible. Setting `allowAliases = false;` introduces a few easily fixable errors in fc-nixos HEAD but since we import ceph from the 21.05 branch that one would require a bunch of adjustments as well. There might be more, not sure how many locations would break exactly. 

I'm not sure if the amount of whack-a-mole this would introduce is worth it

release/default.nix
PhilTaken: Yes, as well as nested package sets. I'll *eventually* get around to it I promise ;)

release/default.nix
PhilTaken: > AFAICS graylog isn't part of important packages and doesn't get evaluated on Hydra yet.

It is not and I doubt it will be since very few customers acutally use graylog. 

The IFD is very unfortunate here but imho removing it is far beyond the scope of this PR

release/default.nix
ctheune: Yeah, we'll have to start with this setting. Once we get a clean baseline we could reverse this.

ctheune (approved):

PR #1565, opened 2025-06-18
[24.11] Remove deprecated networks from configuration from DEV

sysvinit:
We are retiring old subnets which are currently in use in DEV. This change removes those networks from the static platform configuration and from the relevant router configurations.

Additionally, as DEV addresses were in use in some tests and documentation, I've updated those to use IPv6 ULA space or documentation space.

In spite of the changes to the static nameserver configuration this shouldn't need to be backported. Only the IPv6 nameserver has changed, which is only consumed in the KVM role and all of the affected KVM servers are on a current platform version.

PL-133308

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Addresses previously baked into static platform configurations are being removed.
  - Manually tested dev VMs
  - Manually tested dev hardware (KVM, storage, router)
  - bird and keepalived configs for WHQ pass config check on the relevant router

### DIFF
--- a/nixos/roles/router/keepalived/dev.conf
+++ b/nixos/roles/router/keepalived/dev.conf
@@ -68,7 +68,6 @@ vrrp_instance mgm {
     }
 
     virtual_ipaddress_excluded {
-        2a02:238:f030:1c1::1/64 preferred_lft 0
         172.20.1.1/24
     }
 }
@@ -86,7 +85,6 @@ vrrp_instance fe {
     }
 
     virtual_ipaddress_excluded {
-        2a02:238:f030:1c2::1/64
         172.20.2.1/24
     }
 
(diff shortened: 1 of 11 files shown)


nixos/roles/router/keepalived/dev.conf
```diff
@@ -68,7 +68,6 @@ vrrp_instance mgm {
     }
 
     virtual_ipaddress_excluded {
-        2a02:238:f030:1c1::1/64 preferred_lft 0
         172.20.1.1/24
```
ctheune, line 71: why is this going away instead of being replaced?

ctheune (changes requested):

nixos/roles/router/keepalived/dev.conf
sysvinit: It's already been replaced previously, see the address in the `virtual_ipaddress` block (not `_excluded`) for the mgm interface.

ctheune (approved):

PR #1666, opened 2025-07-25
[PL-133672] roles/jitsi: don't enroll coturn secret at eval-time

Ma27:
PL-133672

Generating the secret for prosody/coturn at eval-time via IFD has the problem that the derivation is not in the system's closure and gets GCed eventually. In rare cases this lead to restarts of Jitsi mid-day because of a GC and a subsequent rebuild of the password.

It's a little hard to test this correctly because Jitsi seems pretty resilient to a broken coturn. However, with the password-file not existing for coturn, the service failed to start up. With the password mismatching I got the following errors in coturn's log on a test video-conference with two devices:

    Jul 25 18:37:17 test42 turnserver[617432]: 17: (617438): ERROR: session 002000000000000002: check_stun_auth: Cannot find credentials of user <1753547825>

This only happens on a secret mismatch.

Prosody's Lua code would also fail hard if it couldn't read the secret:

    out>    Jul 25 19:08:22 test42 prosodyctl[634356]: A problem occurred while reading the config file /etc/prosody/prosody.cfg.lua
    out>    Jul 25 19:08:22 test42 prosodyctl[634356]: Error: /etc/prosody/prosody.cfg.lua:97: /run/credentials/prosody.service/turncredentials_secret2: No such file or directory

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Secrets shouldn't be stored world-readable.
  - Availability: avoid restarts duriung agent runs mid-day.
  - Activated on a test-VM
  - Checked that conferences work as expected
  - Tested error-cases (i.e. non-existing files or mismatching secrets).
  - Tested migration from a system with an active Jitsi role.

### DIFF
--- a/changelog.d/20250725_184711_fc-25.05-dev_scriv.md
+++ b/changelog.d/20250725_184711_fc-25.05-dev_scriv.md
@@ -1,0 +1,24 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+- Restart of `coturn.service` & `prosody.service`.
+
+
+### NixOS XX.XX platform
+
+- The secret for the communication between `coturn` & `prosody` is no longer generated at eval-time (PL-133672).
+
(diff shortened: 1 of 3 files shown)


osnyx (approved):
We (still) always generate a TURN secret no matter whether coturn is actually used, but as this simplifies config flow on Jitsi's side I am okay with that.
LGTM.

PR #1672, opened 2025-07-29
fc-manage check: display all Nix eval warnings

osnyx:
Nix evaluation warnings via `builtins.warn` or `lib.warn` do not show up in `config.warnings` and are just printed to stderr during evaluation. fc-agent now grabs and persists the warnings from there at system-build time.

As `config.warnings` warnings from the NixOS module system are also rendered with calls to `builtins.warn` (or a fallback to tracing), these are also captured using that mechanism.

PL-133786

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - provide better visibility for Nix warnings: package (name) deprecations announced via warnings did not show up in the fc-agent check before and consequently were ignored by customers
  - extend test coverage with warnings parsing
  - automated tests still pass
  - manually tested this on a VM with and without warnings

### DIFF
--- a/pkgs/fc/agent/src/fc/manage/manage.py
+++ b/pkgs/fc/agent/src/fc/manage/manage.py
@@ -126,7 +126,7 @@ def check(log, enc, config: ConfigParser) -> CheckResult:
     else:
         errors.append("`nixos` channel not set.")
 
-    nixos_warnings_file = Path("/etc/fcio_nixos_warnings")
+    nixos_warnings_file = Path("/etc/fcio_nix_eval_warnings")
 
     if nixos_warnings_file.exists():
         nixos_warnings_content = nixos_warnings_file.read_text()
--- a/pkgs/fc/agent/src/fc/manage/manage.py
+++ b/pkgs/fc/agent/src/fc/manage/manage.py
@@ -12,7 +12,7 @@ from fc.util.channel import Channel
 from fc.util.checks import CheckResult
 from fc.util.enc import STATE_VERSION_FILE
 from fc.util.lock import locked
-from fc.util.nixos import Specialisation
+from fc.util.nixos import NIX_EVAL_WARNINGS_FILE, Specialisation
 
 # Other platform code can also check the presence of this marker file to
 # change behaviour before/during the first agent run.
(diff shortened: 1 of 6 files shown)


pkgs/fc/agent/src/fc/manage/manage.py
```diff
@@ -126,7 +126,7 @@ def check(log, enc, config: ConfigParser) -> CheckResult:
     else:
         errors.append("`nixos` channel not set.")
 
-    nixos_warnings_file = Path("/etc/fcio_nixos_warnings")
+    nixos_warnings_file = Path("/etc/fcio_nix_eval_warnings")
```
sysvinit: If this file is now generated by the agent and not by the fc-nixos config, is there any reason to have this in `/etc` and not somewhere like `/var/lib` instead?

pkgs/fc/agent/src/fc/util/nixos.py
```diff
@@ -324,6 +324,32 @@ def update_system_channel(channel_url, log=_log):
...
+    """Returns the evaluation warnings raised by either `builtins.warn` or the
+    NixOS module warning system `config.warnings`.
+    Evaluation warnings raised in e.g. package definitions are not accessible via
+    the `config.warnings` module endpoint, hence need to be grabbed from stderr.
+    """
+
+    WARNINGS_MARKER = "evaluation warning: "
+    # As long as lix lacks a `builtins.warn`, the `lib.warn` NixOS code falls
+    # back to traces
+    WARNINGS_MARKER_LEGACY = "trace: evaluation warning: "
```
sysvinit, line 339: Do we actually have Lix deployed anywhere? (I think the compatibility is worth keeping, as `builtins.warn` is a relatively new addition to Nix in any case.)

osnyx: No, we don't. 

**edit**: we're already using lib.warn as a compatibility layer, not builtins.warn directly.

But I wanted to keep this in here in case we ever switch to lix and wonder what is going on here, as I had to investigate the implementation anyways.

pkgs/fc/agent/src/fc/manage/manage.py
osnyx: There is no good reason, I just wanted to keep the change minimal. If you have a strong opinion on this, I can of course move the file to /var/lib/fc-agent (that directoy does not exist yet though -> yet another tmpfiles rule)

osnyx: I'll stick to creating `/var/lib/fc-agent/` via tmpfiles.d rule instead of systemd unit `StateDirectory`, because the imperative `fc-manage` command invocation is independent from the service unit.

sysvinit (approved):

osnyx:
rebased to include formatting changes

sysvinit (approved):

PR #1695, opened 2025-08-06
nixos-mailserver: update and introduce automatic state version update

leona-ya:
NixOS mailserver introduced a stateVersion construct. We need to update that, and ideally automatic. This introduces a script doing the automatic migration and cleans up options that are no longer necessary.

PL-133851

## Release process

  - created update notes entry

- [x] PR has internal ticket

## Design notes

## Security implications

  - Don't run dovecot2 with old directory structure: assured by `requiredBy` condition on the systemd unit
  - When migration script fails, the migration systemd unit fails and dovecot won't start. Also the stateVersion won't be updated as `set -e` is set.
  - Backups ensure ability to restore
  - Migration step 2 is assured by nixos-mailserver assertion
  - Tested OS upgrade on test VM

### DIFF
--- a/nixos/platform/default.nix
+++ b/nixos/platform/default.nix
@@ -20,7 +20,7 @@ let
       )
     else
       [ ];
-  currentPlatform = "24.11";
+  currentPlatform = "25.11";
 
 in
 {
--- a/nixos/services/mail/default.nix
+++ b/nixos/services/mail/default.nix
@@ -56,6 +56,10 @@ in
     flyingcircus.services.mail.enable = lib.mkEnableOption ''
       Mail server (SNM) with Postfix, Dovecot, Rspamd, DKIM & SPF
     '';
+    flyingcircus.services.mail.stateVersionFile = mkOption {
+      type = types.path;
+      default = "/etc/local/nixos/mailserver_state_version";
+    };
   };
 
   config =
@@ -236,6 +240,11 @@ in
           # https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/blob/master/default.nix
           mailserver = {
             enable = true;
+            stateVersion =
+              if pathExists svc.stateVersionFile then
+                lib.toInt (lib.fileContents svc.stateVersionFile)
+              else
+                (if lib.versionOlder config.system.stateVersion "25.11" then 1 else 3);
             inherit domains;
             fqdn = role.mailHost;
             loginAccounts = fclib.jsonFromFile "/etc/local/mail/users.json" "{}";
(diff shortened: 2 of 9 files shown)


leona-ya:
Note that this update blocks Nixpkgs updates for 25.11

nixos/platform/default.nix
```diff
@@ -20,7 +20,7 @@ let
       )
     else
       [ ];
-  currentPlatform = "24.11";
+  currentPlatform = "25.11";
```
osnyx, line 23: Given that I apparently forgot to update this version indentifier in the previous release iteration: Is it really necessary that we hardcode this here, or can this also be inherited from nixpks/NixOS?

nixos/services/mail/default.nix
```diff
@@ -364,12 +373,10 @@ in
                 "permit_sasl_authenticated"
               ];
               # --- postsrsd integration ---
-              recipient_canonical_maps = "socketmap:unix:${config.services.postsrsd.socketPath}:reverse";
```
osnyx, line 367: Why is this removed? How is postsrsd integrated into postfix now?

osnyx (changes requested):
See code comments.

For some other migrations, we have integration tests in NixOS tests using specialisations. Do these make sense here as well or do you consider one-off manual testing sufficient, as this migration is a one-off for now as well?

In the security considerations you write
> Backups ensure ability to restore

Do you refer to the platform backups via backy here (for which the snapshot time can deviate from the time the migration script is started, potentially loosing changes in the interval between), or is there an additional backup mechanism in the migration script I have not spotted yet?

nixos/services/mail/default.nix
leona-ya: This has been removed as it's now the default in the Nixpkgs `postsrsd` module: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/mail/postsrsd.nix#L242

nixos/platform/default.nix
leona-ya: I don't know, probably this could be inherited

leona-ya:
There is no backup in the script, but the script primarily moves the maildirs around, so this shouldn't be much of a problem. I also don't think it's a good idea to integrate a copy of the maildir in the script as this would require more disk space and a (possibly manual) cleanup action on each host.

A NixOS test unfortunately doesn't work here I think, as as far as I know we don't have architecture to test a OS upgrade. The 25.11 NixOS-Mailserver module doesn't support `stateVersion < 3` (and therefore requires the migration before the first start) and the 25.05 one doesn't support the migration.

osnyx (approved):

PR #1701, opened 2025-08-08
[PL-128135] Add s3 gateway on rgw servers

Ma27:

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

### DIFF
--- a/nixos/roles/ceph/rgw.nix
+++ b/nixos/roles/ceph/rgw.nix
@@ -72,13 +72,12 @@ in
             bool
           ]);
         default = { }; # defaults are provided in the config section with a lower priority
-        description =
-          ''
-            config section of the Ceph config file of the radosgw client user.
-            Can override existing default setting values. Configuration keys like `mon osd full ratio`''
-          + ''
-            can alternatively be written in camelCase as `monOsdFullRatio`.
-          '';
+        description = ''
+          config section of the Ceph config file of the radosgw client user.
+          Can override existing default setting values. Configuration keys like `mon osd full ratio`''
+        + ''
+          can alternatively be written in camelCase as `monOsdFullRatio`.
+        '';
       };
 
       cephRelease = fclib.ceph.releaseOption // {
@@ -116,6 +115,14 @@ in
           PDNS_API_URL=https://dns.flyingcircus.io/
           PDNS_PROPAGATION_TIMEOUT=3610
           ```
+        '';
+      };
+
+      enableAccounting = lib.mkOption {
+        type = lib.types.bool;
+        default = true;
+        description = ''
+          Whether or not to enable traffic accounting.
         '';
       };
     };
--- a/nixos/roles/ceph/rgw.nix
+++ b/nixos/roles/ceph/rgw.nix
@@ -290,9 +290,13 @@ in
       };
 
       flyingcircus.services.sensu-client.checks = {
-        radosgw_probe_object = {
+        radosgw_probe_object_direct = {
           notification = "Probe object (/rgw-monitoring/probe) not OK.";
-          command = "check_http -u /rgw-monitoring/probe -H localhost -m 1000000:1500000 -w 5 -c 10";
+          command = "check_http -u /rgw-monitoring/probe -H localhost:7480 -m 1000000:1500000 -w 5 -c 10";
+        };
+        radosgw_probe_object_via_nginx = {
+          notification = "Probe object (/rgw-monitoring/probe) not OK.";
+          command = "check_http -u /rgw-monitoring/probe -H ${role.domain} -m 1000000:1500000 -w 5 -c 10 -f follow";
         };
       };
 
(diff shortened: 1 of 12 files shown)


nixos/roles/ceph/rgw.nix
```diff
@@ -21,18 +21,33 @@ let
 
   cephPkgs = fclib.ceph.mkPkgs role.cephRelease;
 
+  radosListenPort = toString 7480;
```
ctheune, line 24: I *think* we need to leave 80 in here for pre-signed URLs?

nixos/roles/ceph/rgw.nix
```diff
@@ -21,18 +21,33 @@ let
...
+  haproxyPort = toString 7475;
+  vhostName = lib.concatStringsSep "." (
```
ctheune: I'd rather have this being a proper option where the computed value is the default and we can completely override this. Otherwise this appears to be a potential "let trap".

nixos/roles/ceph/rgw.nix
```diff
@@ -21,18 +21,33 @@ let
...
     pidFile = "/run/ceph/radosgw.pid";
     adminSocket = "/run/ceph/radosgw.asok";
     rgwData = "/srv/ceph/radosgw/ceph-$id";
-    rgwEnableOpsLog = false;
+    rgwEnableOpsLog = true;
+    rgwOpsLogRados = true;
     rgwMimeTypesFile = "${pkgs.mime-types}/etc/mime.types";
     debugRados = "1 5";
-    rgwFrontends = "beast port=80";
+    rgwFrontends = "beast port=${radosListenPort}";
```
ctheune, line 38: please bind explicity to srv (but also keep the firewall!)

>  Sets the listening address in the form address[:port], where the address is an IPv4 address string in dotted decimal form, or an IPv6 address in hexadecimal notation surrounded by square brackets. Specifying a IPv6 endpoint would listen to v6 only. The optional port defaults to 80 for endpoint and 443 for ssl_endpoint. Can be specified multiple times as in endpoint=[::1] endpoint=192.168.0.100:8000.

nixos/roles/ceph/rgw.nix
```diff
@@ -144,37 +235,22 @@ in
       networking.firewall.extraCommands =
         let
           srv = fclib.network.srv;
-          sto = fclib.network.sto;
         in
-        lib.mkMerge [
-          (lib.mkOrder 700 ''
```
ctheune, line 150: those might be needed if we have to enable the port redirections again

nixos/roles/ceph/rgw.nix
```diff
@@ -210,7 +286,11 @@ in
       flyingcircus.services.sensu-client.checks = {
         radosgw_probe_object = {
```
ctheune: lets name this rados_gw_probe_object_nginx

nixos/roles/ceph/rgw.nix
```diff
@@ -210,7 +286,11 @@ in
...
           notification = "Probe object (/rgw-monitoring/probe) not OK.";
-          command = "check_http -u /rgw-monitoring/probe -H localhost -m 1000000:1500000 -w 5 -c 10";
+          command = "check_http -u /rgw-monitoring/probe -H localhost:7480 -m 1000000:1500000 -w 5 -c 10";
+        };
+        radosgw_probe_object_gw = {
```
ctheune: lets name this radosgw_probe_object_direct

nixos/roles/ceph/rgw.nix
```diff
@@ -271,6 +351,32 @@ in
...
+          interval = 60;
+          command = "${lib.getExe pkgs.python3} ${pkgs.writeText "check-s3-accounting" ''
+            import json
+            import sys
+            from datetime import datetime, timedelta
+
+            with open("/var/lib/fc-ceph-s3-accounting/s3-accounting-state", "r") as f:
+                data = json.load(f)
+
+            THRESH_WARN = timedelta(days=1)
```
ctheune: 6 hours warning
12 hours critical.

pkgs/fc/ceph/src/fc/ceph/logs.py
```diff
@@ -50,3 +55,40 @@ def slowreq_histogram(self, include, exclude, first_n, filenames):
             if n >= first_n:
                 break
             n += 1
+
+    def account_s3_traffic(self, state_file: Path, enc_file: Path):
```
ctheune, line 59: lets gate the accounting timer behind a feature flag so in case this explodes and we have to adjust things

ctheune: This is also a potential RAM bottleneck for processing if we have to parse the json completely. At least this is split by bucket so that helps, but we might have to switch to an event-based json parser (bleh) in the future ... leave as is for now.

pkgs/fc/ceph/src/fc/ceph/util/opslog.py
```diff
@@ -0,0 +1,130 @@
...
+            hours=1
+        )
+
+        yield stats_by_day
+
+        self.opslog_state.write_to(self.state_file)
+
+    def gc_log_objects(self):
+        day = self.opslog_state.last_gced_day + timedelta(days=1)
+        end = self.opslog_state.last_processed_datetime.date() - timedelta(
```
ctheune: lets have a small guard that ensures that end is always at most yesterday and never today or in the future.

ctheune (changes requested):
Definitely on the right track, see the comments.

nixos/roles/ceph/rgw.nix
Ma27: Do you happen to have a reproducer for the old issue? Because I cannot reproduce on dev (and the tests are fine either):

```
$ cat .aws/credentials
[dev2]
endpoint_url = https://objects.dev.fcio.net
<redacted>
$ aws --profile dev2 s3 ls s3://test/
[...]
2025-07-21 13:21:18         12 testfile
$ aws --profile dev2 s3 presign s3://test/testfile
https://objects.dev.fcio.net/test/testfile?AWSAccessKeyId=03KAF2L1WU7PHIH0H66M&Signature=Oh1zNWA0TnHbdNCPerhlEEG%2FqHU%3D&Expires=1755084588
$ curl "https://objects.dev.fcio.net/test/testfile?AWSAccessKeyId=03KAF2L1WU7PHIH0H66M&Signature=Oh1zNWA0TnHbdNCPerhlEEG%2FqHU%3D&Expires=1755084588"
Hello World
```

leona-ya:
continued in #2142

PR #1738, opened 2025-08-26
open-webui: init role

osnyx:

PL-133951

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - integrate authentication system with Flying Circus directory
  - ensure compatibility of role and NixOS upstream module
  - permission model needs to be known by the users
  - NixOS smoke test for role, ensuring that the NixOS code evaluates and the ExecStart remains compatible
  - manually tested OIDC auth integration with a dev deployment of the directory
  - permission model is documented

### DIFF
--- a/doc/src/open-webui.md
+++ b/doc/src/open-webui.md
@@ -10,9 +10,9 @@ Managed instance of the [Open WebUI](https://openwebui.com/) LLM interface. It a
 
 % TODO: link to subscription, verify subscription naming
 
-A *Flying Circus AI Services* subscription is required for this role.
+A *Flying Circus AI Services* subscription is required for this role. An access token must be created in the customer portal to allow the Open WebUI instance to access our central AI gateway.
 
-Enabling the `open-webui` role automatically sets up Open WebUI available under `https://ai-chat.<resource group name>.fcio.net`. As the first account accessing the application after installation automatically gains administrator privileges, we recommend that a designated admininstrator logs in immediately after setting up the role.
+Enabling the `open-webui` role automatically sets up Open WebUI available under `https://ai-chat.<resource group>.fcio.net`. The access token from the customer portal must then be copied into the file `/etc/local/open-webui/secret` on the VM with the role assigned. As the first account accessing the application after installation automatically gains administrator privileges, we recommend that a designated administrator logs in immediately after the role is assigned and configured.
 
 ## User Management
 
--- a/doc/src/open-webui.md
+++ b/doc/src/open-webui.md
@@ -1,0 +1,26 @@
+(nixos-role-open-webui)=
+
+# Open WebUI
+
+Managed instance of the [Open WebUI](https://openwebui.com/) LLM interface. It automatically integrates with the *Flying Circus AI services*.
+% TODO: link to more general FC AI docs/ overview
+
+## Initial Setup
+
+
+% TODO: link to subscription, verify subscription naming
+
+A *Flying Circus AI Services* subscription is required for this role.
+
+Enabling the `open-webui` role automatically sets up Open WebUI available under `https://ai-chat.<resource group name>.fcio.net`. As the first account accessing the application after installation automatically gains administrator privileges, we recommend that a designated admininstrator logs in immediately after setting up the role.
+
+## User Management
+
+User access is centrally managed via [Flying Circus users](/platform/reference/users/index.html). By default, all members of the resource group are able to log in to the service.  \
+To require the approval of an Open WebUI administrator, set `flyingcircus.roles.open-webui.usersNeedApproval = true;` via NixOS config. New users may then be approved by clicking their *pending* role badge in the administrator interface.
+
+User roles and groups have to be manually managed in the Open WebUI administrator settings. The first user successfully logging in after the initial setup of the role receives administrator privileges. Additional administrators can be authorised by that user in the Open WebUI administrator interface.
+
+## Available AI Models
+
+All models available through our *Flying Circus AI services* are automatically discovered by Open WebUI, but are only available to administrators by default. Under *Administration -> Settings -> Models*, each model can be either set public or be made accessible to certain groups.
--- a/nixos/roles/open-webui.nix
+++ b/nixos/roles/open-webui.nix
@@ -10,7 +10,14 @@ let
   scfg = config.services.open-webui;
   fclib = config.fclib;
   inherit (builtins) head;
-  inherit (lib) mkOption mkEnableOption types;
+  inherit (lib)
+    mkOption
+    mkEnableOption
+    mkOverride
+    types
+    optional
+    optionalString
+    ;
 in
 {
   options.flyingcircus.roles.open-webui = {
@@ -35,6 +42,13 @@ in
     };
     # expose through role config for better visibility
     environmentFile = options.services.open-webui.environmentFile;
+    llmApiSecretFile = mkOption {
+      example = "/etc/local/open-webui/secret";
+      default = null;
+      type = types.nullOr types.str;
+      description = "A file containing the authentication secret for the Flying
+      Circus AI Provider.";
+    };
   };
 
   # TODO: role shall also configure an nginx with TLS (or webgateway?)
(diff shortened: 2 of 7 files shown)


nixos/roles/open-webui.nix
```diff
@@ -0,0 +1,145 @@
...
+              proxy_buffering off;
+            '';
+          };
+          "/ws/" = {
+            proxyPass = "http://${scfg.host}:${toString scfg.port}";
+            extraConfig = ''
+              proxy_http_version 1.1;
+              proxy_set_header Upgrade $http_upgrade;
+              proxy_set_header Connection "upgrade";
+            '';
```
leona-ya: We should be able to simplify this with just having a `/` location and `proxyWebsockets = true`, I think. I think we don't need this special casing

nixos/roles/open-webui.nix
```diff
@@ -0,0 +1,145 @@
...
+      dir = "/etc/local/open-webui";
+    };
+
+    networking.firewall.allowedTCPPorts = [
+      80
+      443
+    ];
+
+    services.nginx = {
+      enable = true;
```
leona-ya, line 111: This feels very wrong. You're duplicating our default nginx config here, which might cause diversions and conflicts in the future. Instead we should probably use `flyingcircus.services.nginx.enable = true` (or maybe even `flyingcircus.roles.webgateway.enable = true` - I don't know that which one is better).

nixos/roles/open-webui.nix
```diff
@@ -0,0 +1,145 @@
...
+    };
+
+    flyingcircus.localConfigDirs.open-webui = {
+      dir = "/etc/local/open-webui";
+    };
+
+    networking.firewall.allowedTCPPorts = [
+      80
+      443
+    ];
```
leona-ya: This can be removed with the change suggested in the comment below

nixos/roles/open-webui.nix
```diff
@@ -0,0 +1,145 @@
...
+      default = "https://api.ai.flyingcircus.io";
+      type = types.str;
+      description = ''
+        Endpoint of the (OpenAI-style) LLM API used as main upstream AI provider.
+        In most cases, this should NOT contain a trailing /.
+      '';
+    };
+    usersNeedApproval = mkOption {
+      default = false;
+      type = types.boolean;
```
leona-ya: that's not a valid type, it's named `types.bool`

tests/default.nix
```diff
@@ -95,6 +95,7 @@ in
   opensearch = callTest ./opensearch.nix { };
   opensearch_dashboards = callTest ./opensearch_dashboards.nix { };
   openvpn = callTest ./openvpn.nix { };
+  open-webui = callTest ./open-webui { };
```
leona-ya:
```suggestion
  open-webui = callTest ./open-webui.nix { };
```

nixos/roles/open-webui.nix
```diff
@@ -0,0 +1,145 @@
...
+        admin settings.'';
+    };
+    # expose through role config for better visibility
+    environmentFile = options.services.open-webui.environmentFile;
+    llmApiSecretFile = mkOption {
+      example = "/etc/local/open-webui/secret";
+      default = null;
+      type = types.nullOr types.str;
+      description = "A file containing the authentication secret for the Flying
+      Circus AI Provider.";
```
leona-ya: You probably want to use an actual multiline string here

nixos/roles/open-webui.nix
```diff
@@ -0,0 +1,145 @@
...
+      ExecStart =
+        let
+          execStart = pkgs.writeShellScriptBin "open-webui-start" ''
+            ${optionalString (
+              cfg.llmApiSecretFile != null
+            ) "export OPENAI_API_KEYS=$(<$CREDENTIALS_DIRECTORY/OPENAI_API_KEYS)"}
+            exec ${lib.getExe scfg.package} serve --host "${scfg.host}" --port ${toString scfg.port}
+          '';
+        in
+        mkOverride 90 "${execStart}/bin/open-webui-start";
```
leona-ya:
```suggestion
        mkOverride 90 (lib.getExe execStart);
```

leona-ya (changes requested):

nixos/roles/open-webui.nix
sysvinit: We adapted this from the code for the statshost role, which uses the same pattern of setting `services.nginx.enable = true` from the depths of the role code. (I believe that webgateway also turns on an haproxy for load balancing which doesn't really seem necessary here.)

I was under the impression that `flyingcircus.services.nginx` is considered deprecated, at least it's documented as such in https://doc.flyingcircus.io/roles/fc-24.11-production/upgrade.html#webgateway-nginx.

leona-ya: That's partly true. We want to deprecate that we carry our own module mostly and use upstream. We will still need a point where we set our defaults - that might be `flyingcircus.roles.nginx` or `flyingcircus.services.nginx`, not sure which one

nixos/roles/open-webui.nix
sysvinit: Fixed in ce09f6669e1f3415ae2a7e0d499c1cc698147b66.

tests/default.nix
sysvinit: Fixed in ce09f6669e1f3415ae2a7e0d499c1cc698147b66.

nixos/roles/open-webui.nix
sysvinit: Fixed in ce09f6669e1f3415ae2a7e0d499c1cc698147b66.

nixos/roles/open-webui.nix
sysvinit: Fixed in ce09f6669e1f3415ae2a7e0d499c1cc698147b66.

nixos/roles/open-webui.nix
sysvinit: I've taken a look at `flyingcircus.services.nginx` and we do set quite a few defaults there – the principle of least surprise would speak in favour of using it here as well to ensure consistent behaviour with other nginx instances. I'll discuss this with @osnyx.

nixos/roles/open-webui.nix
sysvinit: Fixed in [2afee60](https://github.com/flyingcircusio/fc-nixos/pull/1738/commits/2afee60d1d2455e87bf8389a2a70cd41707049c5)

nixos/roles/open-webui.nix
sysvinit: Fixed in [2afee60](https://github.com/flyingcircusio/fc-nixos/pull/1738/commits/2afee60d1d2455e87bf8389a2a70cd41707049c5). We've switched to `flyingcircus.services.nginx` to ensure that we have uniform defaults.

nixos/roles/open-webui.nix
sysvinit: Fixed, see other comments.

doc/src/open-webui.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Initial Setup
+
+
+% TODO: link to subscription, verify subscription naming
+
+A *Flying Circus AI Services* subscription is required for this role.
+
+Enabling the `open-webui` role automatically sets up Open WebUI available under `https://ai-chat.<resource group name>.fcio.net`. As the first account accessing the application after installation automatically gains administrator privileges, we recommend that a designated admininstrator logs in immediately after setting up the role.
+
```
leona-ya, line 16: It would be good to mention the required secret setup process here.

nixos/roles/open-webui.nix
```diff
@@ -0,0 +1,128 @@
...
+    services.open-webui = {
+      enable = true;
+      host = fclib.mkPlatform config.networking.hostName;
+      environment = {
+        ENABLE_SIGNUP = "false";
+        ENABLE_LOGIN_FORM = "false";
+        ENABLE_OAUTH_SIGNUP = "true";
+        OAUTH_CLIENT_ID = "${config.networking.hostName}_open-webui";
+        OPENID_PROVIDER_URL = "https://auth.flyingcircus.io/realms/fcio/.well-known/openid-configuration";
+        OAUTH_PROVIDER_NAME = "FCIO";
```
leona-ya: We probably want something more accessible for our customers like "Flying Circus"

tests/open-webui.nix
```diff
@@ -0,0 +1,31 @@
...
+        # cannot download sentence transformers model without network
+        # connection.
+        services.open-webui.environment.RAG_EMBEDDING_MODEL = "";
+      };
+    testScript = ''
+      machine.wait_for_unit("multi-user.target")
+
+      # open webui is slow to boot, but should eventually respond
+      # with http status 200
+      machine.wait_until_succeeds("curl http://192.168.3.1:8080", timeout=30)
```
leona-ya, line 33: The timeout doesn't seem to be long enough, it fails sometimes in Hydra, e.g. https://hydra.flyingcircus.io/build/24410413/nixlog/41/tail

sysvinit: The timeout is due to a change in the configuration since I wrote that test, I've fixed it in [976d3ba](https://github.com/flyingcircusio/fc-nixos/pull/1738/commits/976d3ba125eef6b512a6e85ea78e708d079b16e4).

doc/src/open-webui.md
sysvinit: I've added a few sentences in [e42b7b6](https://github.com/flyingcircusio/fc-nixos/pull/1738/commits/e42b7b663aaa33ec33bf20d911463c4fcf6416e6), the details can be adjusted later once we have the other parts all put together.

nixos/roles/open-webui.nix
sysvinit: Fixed in [e42b7b6](https://github.com/flyingcircusio/fc-nixos/pull/1738/commits/e42b7b663aaa33ec33bf20d911463c4fcf6416e6).

leona-ya (dismissed):

leona-ya (approved):

PR #1759, opened 2025-09-02
Router: add options for configuration of the query source addresses used by Bind

sysvinit:
This change refactors some of the bind configuration, inlining it into the adjacent nix file, and adds options to the router role to allow setting the preferred source address used by bind for sending DNS queries.

PL-133852

## Release process

- [x] PR has internal ticket

## Design notes

  - The new options are nullable and default to null, so by default they have no effect. These options must be explicitly configured by the operator in order to have an effect.

## Security implications

  - Configuring specific source addresses may be required depending on the site-local network configuration and the usability of transfer addresses with the upstream network provider.
  - Validation tested with Hydra.
  - Functionality manually verified in dev against comparison with tcpdump packet traces.


osnyx (commented):
I do not know too much about the named.conf format, but does it make sense to provide a templating place for a generic `extraOptions` free-text escape hatch snippet towards the end of the file, or via inclusion of another file?
Or do we just want to `configFile = lib.mkForce <some new file>` in these cases?

sysvinit:
Our approach to overriding up until now has been `configFile = lib.mkForce ...`, which wouldn't change here.

Ideally we would have a uniform configuration in all locations and wouldn't need a flexible `extraOptions` escape hatch. It's also not a great fit for bind's nested configuration structure and locally overriding options which are set in the stock version – those are the cases where we've hitherto reached for `mkForce` as a bigger hammer.

osnyx (approved):

PR #1768, opened 2025-09-04
AI model server role

elikoga:
This PR introduces a new `ai-model-server` role for using Ollama for AI inference with acceleration on AMD GPUs.

PL-133949

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - This role provides an inference API which will handle potentially sensitive data. However, the ollama backend is stateless up to the KV cache.
  - ollama listens on the SRV interface which has the resource group firewall active to prevent unauthorised access to the inference backend. Access to ollama is handled by the AI gateway proxy.
  - Role manually verified on development inference server.

### DIFF
--- a/nixos/roles/ai-model-server.nix
+++ b/nixos/roles/ai-model-server.nix
@@ -19,6 +19,14 @@ in
       enable = lib.mkEnableOption "Enable GPU server role with AI inference capabilities";
       supportsContainers = fclib.mkDisableDevhostSupport;
 
+      # Enabled by default, can be disabled for running in VMs or
+      # tests.
+      enableRocm = lib.mkOption {
+        type = lib.types.bool;
+        default = true;
+        description = "Enable ROCM for AMD GPU acceleration";
+      };
+
       # Model configuration
       models = lib.mkOption {
         type = lib.types.attrsOf (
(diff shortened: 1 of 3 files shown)


nixos/roles/gpuserver.nix
```diff
@@ -0,0 +1,176 @@
...
+}:
+
+with builtins;
+
+let
+  fclib = config.fclib;
+  cfg = config.flyingcircus.roles.gpuserver;
+
+  # Pin specific package versions for model serving
+  pkgs_2505 = import (fetchTarball "https://releases.nixos.org/nixos/25.05/nixos-25.05.802491.7c43f080a7f2/nixexprs.tar.xz") { };
```
osnyx: `builtins.fetchTarball` happens during eval phase and significantly slows down our Hydra evaluation loop. Please replace with one the fixed-output-derivation fetchers from `pkgs`, like `fetchzip`, which happen during build phase of their own derivation.

sysvinit: Yes, this is a known issue, this PR is still a work in progress.

leona-ya:
We have decided on the name `ai-model-server`, that should be used here too.

nixos/roles/gpuserver.nix
```diff
@@ -0,0 +1,170 @@
...
+      models = cfg.modelsPath;
+      host = cfg.hostAddress;
+    };
+
+    # Firewall configuration for model access
+    networking.firewall = {
+      allowedTCPPorts = [
+        scfg.port
+      ];
+    };
```
leona-ya: This is most likely wrong. The port should be only open on SRV to members of the same resource group. This should already be achieved by `flyingcircus.firewall.enableSrvRgFirewall`.

sysvinit: Good point.

nixos/roles/ai-model-server.nix
```diff
@@ -0,0 +1,157 @@
...
+          [
+            "L+  /opt/rocm  - - - -  ${rocmEnv}"
+          ];
+
+        # AMD specific ollama setup
+        services.ollama = {
+          package = pkgs.ollama-rocm;
+          acceleration = "rocm";
+          # Pin to gfx1101 LLVM target for Radeon PRO W7900 GPUs
+          rocmOverrideGfx = "11.0.1";
```
leona-ya: Is this correct? https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html says gfx1100

sysvinit: gfx1101 is what I blindly copied out of the hand-constructed ollama configuration, and it all seems to still work properly, but specifics of GPU configuration is a bit beyond me.

leona-ya (approved):

PR #1778, opened 2025-09-08
Introduce skvaider pkg and ai-api-gateway role

leona-ya:

PL-133948

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Aramaki security requirements are handled seperately
  - Full enc file including directory password is not visible to the application
  - secret_salt is only readable by aramaki and root: ensured by systemd options
  - Secret for tokens is only stored hashed
  - Application is automatically restarted on failures

### DIFF
--- a/nixos/roles/ai-api-gateway.nix
+++ b/nixos/roles/ai-api-gateway.nix
@@ -1,0 +1,139 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.flyingcircus.roles.ai-api-gateway;
+  fclib = config.fclib;
+  enc = config.flyingcircus.enc;
+  settingsFormat = pkgs.formats.toml { };
+  baseConfigFile = settingsFormat.generate "skvaider-config.toml" {
+    aramaki = {
+      url = cfg.aramakiUrl;
+      state_directory = "/var/lib/skvaider/aramaki";
+      secret_salt = "@secret_salt@";
+      principal = enc.name;
+    };
+    backend = builtins.map (val: {
+      type = "openai";
+      url = "http://${val.address}:11434";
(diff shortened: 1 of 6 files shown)


pkgs/default.nix
```diff
@@ -1,16 +0,0 @@
...
-
-    fc = import ./fc {
-      inherit (self) callPackage;
-      inherit pkgs pyPackages;
-    };
-
-  };
-
-in
-self.fc
```
sysvinit: Why the refactoring here?

leona-ya: This is not strictly required for this change. I had this lying around and in a former version of this PR (using `buildPythonApplication`) this was helpful. In the past I had seen multiple unnecessary rebuilds of all tests just because of a change in the `fc` package set. At some point, I'd like to change this, but I could create a seperate PR.

sysvinit: Just a question for understanding – it's a separate commit anyway and now documented here in the GH comments.

leona-ya: I removed that change as it caused other problems because of awful hacks. Will try again at another time.

nixos/roles/ai-api-gateway.nix
```diff
@@ -0,0 +1,137 @@
...
+      delaycompress = true;
+      copytruncate = true;
+    };
+
+    flyingcircus.services.nginx.enable = true;
+    services.nginx.virtualHosts.${cfg.hostname} = {
+      enableACME = true;
+      forceSSL = true;
+
+      locations."/" = {
```
leona-ya, line 129: Do we need something like this here?

```
proxy_read_timeout 1200s;
proxy_buffering off;

ctheune (dismissed):

leona-ya:
(fixed merge conflict)

ctheune (approved):

PR #1820, opened 2025-09-23
Move VFS cache sysctl to backyserver role behind option

sysvinit:
Previously we've been setting `vm.vfs_cache_pressure=10` in the sysctl settings in the general hardware profile configuration as a holdover from Gentoo, to improve performance (and presumably latency) on filesystems with lots of small files backed by hard disks. However, we've recently encountered problems where workloads with high memory requirements (e.g. rclone backups of object storage) have been OOM-killed, because the VFS cache has been configured with such a high weight that the kernel does not evict entries in order to make memory available to the user space processes.

Empirically, setting the VFS cache pressure back to the default value of 100 fixes this problem. Such aggressive caching is no longer appropriate to have deployed everywhere given that the newest generation of backup servers have NVME flash storage (which do not have the performance bottlenecks of hard disks which require cache tuning), and the Ceph clusters have long since been migrated to Bluestore and access disks directly without a filesystem sitting in between.

This change removes this sysctl setting from the hardware infrastructure module. It also introduces a NixOS option in the backy role for backwards compatibility on older servers which still use hard disk backed storage which may still benefit from more eager caching.

PL-133712

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - The kernel cache settings are superfluous on contemporary hardware, and violate the principle of least astonishment when debugging OOMs.
  - Tested on DEV backup server. As the explicit configuration is removed without replacement, the implicit default will take effect at the next reboot.

### DIFF
--- a/nixos/roles/backyserver.nix
+++ b/nixos/roles/backyserver.nix
@@ -73,6 +73,19 @@ in
         description = "Codename of the Ceph release series used as external backy tooling.";
       };
 
+      increaseVfsCacheWeight = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Increase the relative weight of the kernel VFS cache to prevent/defer eviction of filesystem
+          objects from working memory.
+
+          Setting this option can improve performance on systems with HDD-backed filesystems where
+          the default cache eviction behaviour is too eager for regular scans of filesystems with
+          a large number of small files.
+        '';
+      };
+
       # this indirection is required by the tests as a config input
       fsOptions = lib.mkOption {
         type = lib.types.attrs;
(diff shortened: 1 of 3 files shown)


nixos/roles/backyserver.nix
```diff
@@ -131,6 +144,10 @@ in
       # Extracted to flyingcircus-physical.nix
       # kernel.sysctl."vm.vfs_cache_pressure" = 10;
```
osnyx: I think this is a leftover and can be removed.

osnyx (changes requested):

nixos/roles/backyserver.nix
sysvinit: Not sure how I managed to miss that given how it was there staring me in the face...

osnyx (approved):

PR #1826, opened 2025-09-25
statshost: OIDC as default, disable local login

dpausp:

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - reduce local login accounts and use OIDC instead
  - when OIDC enabled, local login must not be enabled. Grafana redirects to Keycloak automatically
  - only applies to statshost-master, global statshost has locally managed SSO config
  - automated tests check that statshost-master and statshost still work
  - checked on a test VM that switching from LDAP to OIDC works, also the other way round and back to OIDC. User accounts are updated properly and show the currently enabled login method after re-login (logged-in users are not logged out automatically, that's ok)

### DIFF
--- a/nixos/roles/statshost/default.nix
+++ b/nixos/roles/statshost/default.nix
@@ -166,10 +166,7 @@ in
       };
 
       ldap = {
-        enable = mkOption {
-          type = types.bool;
-          default = true;
-        };
+        enable = mkEnableOption "Enable LDAP for Grafana";
 
         memberOf = mkOption {
           default = config.flyingcircus.enc.parameters.resource_group;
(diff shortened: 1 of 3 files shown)


leona-ya:
I'd say that this change could lead to confusion or other behavior changes for our customers. I suggest moving this to 25.11 and introduce with this release. Now, there is no security reason to fast-track this change.

nixos/roles/statshost/default.nix
```diff
@@ -692,6 +635,74 @@ in
...
+            name = "FCIO oidc";
+            allow_sign_up = true;
+            client_id = "${config.networking.hostName}_statshost-master";
+            client_secret = fclib.derivePasswordForHost "oidc_statshost-master";
+            scopes = "openid email profile";
+            email_attribute_path = "email";
+            login_attribute_path = "preferred_username";
+            name_attribute_path = "full_name";
+            # Everybody becomes admin, ignoring user roles from OIDC.
+            role_attribute_path = "\"'Admin'\"";
```
PhilTaken, line 700: I don't think we should assign admin to everybody considering that access to the grafana webinterface is not limited to fcio staff.

dpausp: It's the same behaviour as LDAP always had, we can improve that later :)

leona-ya (approved):

PR #1846, opened 2025-10-01
Replace bind with knot-resolver as the recursive resolver on the routers

sysvinit:
PL-134065

## Release process

- [ ] PR has internal ticket

## Design notes

## Security implications


nixos/platform/network.nix
```diff
@@ -321,8 +321,8 @@ in
 
         resolvconf.extraOptions = [
           "ndots:1"
-          "timeout:1"
-          "attempts:6"
+          "timeout:5" # in correspondence with rfc8767
```
ctheune: ah, i remember: this isn't quite in correspondence. rfc8767 suggests 2 as a common option, but we can't use 2 because kresd isn't able to stale timeout in 1.8 seconds due to granularity.

nixos/roles/router/kresd/default.nix
```diff
@@ -0,0 +1,167 @@
...
+      -- can immediately respond to requests. however, this also means backends
+      -- may take a minute to recover
+      cache.ns_tout(60000)
+
+      -- access control: only answer queries from ourselves and own address
+      -- space
+      view:addr('127.0.0.0/8', policy.all(policy.PASS))
+      view:addr('::1/128', policy.all(policy.PASS))
+
+      ${lib.concatMapStringsSep "\n" (
```
ctheune, line 99: I was wondering whether this will cause unnecessary hard reloads. I was considering to put this into firewalling code instead if it does.

nixos/roles/router/kresd/default.nix
```diff
@@ -0,0 +1,167 @@
...
+
+      -- load a hosts file for our assignments out of rfc1918 address space in
+      -- order to synthesise PTR records for reverse dns. don't synthesise
+      -- nodata responses for queries with mismatching address type (i.e.
+      -- resolve AAAA records using public dns for names which only have an
+      -- ipv4 address in the hosts file). this is all wrapped in some logic to
+      -- ensure that the hosts file can be reloaded by sighup at runtime.
+
+      local function load_private_hosts()
+          hints.add_hosts('/etc/nixos/rfc1918-hosts')
```
ctheune: does this remove outdated hosts?

nixos/roles/router/kresd/default.nix
```diff
@@ -0,0 +1,167 @@
...
+        else
+          ''
+            -- the policy module is compliant with rfc6303 by default, override
+            -- these defaults to accept queries for rfc1918 reverse zones by the
+            -- hints module
+            policy.add(
+              policy.suffix(
+                policy.PASS,
+                policy.todnames({
+                  '10.in-addr.arpa.',
```
ctheune, line 159: I guess this is work in progress, right?

nixos/roles/router/kresd/default.nix
sysvinit: Well, the short answer is yes, this will cause a restart (not a reload) if the list changes. It's possible that kresd might restart fast enough for this to not be a problem, though this is indeed something which has bitten us with bind in the past.

As for the firewalling... if we put the list of allowed IP ranges in the firewall, then changing the allowed IPs will cause a *firewall* reload, which on the routers usually causes the BGP sessions to go down and cause a failover due to the BFD getting interrupted. I presume by "put this into firewalling" you mean restrict access to DNS to only "downstream" interfaces, and e.g. dropping incoming DNS from the transfer/uplink interfaces?

nixos/roles/router/kresd/default.nix
sysvinit: No, this is required. If we do not have forwarders configured in kresd, so it's performing iterative resolution itself, it will refuse to answer queries for the reverse zones for RFC1918 addresses due to the defaults of the policy module (see [here](https://gitlab.nic.cz/knot/knot-resolver/-/blob/master-5/modules/policy/policy.lua?ref_type=heads#L922) and [here](https://gitlab.nic.cz/knot/knot-resolver/-/blob/master-5/modules/policy/policy.lua?ref_type=heads#L1034-1040)).

These defaults are overridden by having any policy at all, which is why we don't need the list of reverse zones when we have forwarders configured.

nixos/roles/router/kresd/default.nix
sysvinit: Nope; fixed in [8b08068](https://github.com/flyingcircusio/fc-nixos/pull/1846/commits/8b08068522d5704322128c6575e2f9cedfd623e6).

nixos/platform/network.nix
sysvinit: Fixed in [b4a1705](https://github.com/flyingcircusio/fc-nixos/pull/1846/commits/b4a17059152a58d6c64a87828b16761fd05a18c9).

nixos/roles/router/kresd/default.nix
```diff
@@ -0,0 +1,167 @@
...
+
+      -- resilience: prefer serving stale data over not serving at all.
+      -- rfc8767 recommends a stale serving timeout of 1800s with a client
+      -- query timeout of 2s. however, kresd has a minimum granularity of
+      -- 2s (KR_CONN_RTT_MAX), so we need to configure a higher client query
+      -- timeout than recommended in the rfc to ensure that clients wait for
+      -- long enough to trigger the stale serving behaviour. we match the
+      -- rfc for the stale serving timeout though so we get the benefits
+      -- if the granularity changes in the future.
+      serve_stale.timeout = 1800
```
ctheune: reminder to me: set this to 3 s

nixos/roles/router/kresd/default.nix
ctheune: multi instance (4?) + execstop with jittered pause?

PR #1863, opened 2025-10-08
varnish: provide an option to define a fallback configuration and mix nix/non-nix configuration

PhilTaken:
[FC-41407](https://yt.flyingcircus.io/issue/FC-41407)

Add a new "fallback configuration" option to the Varnish service that serves as a fallback when none of the conditions for the defined virtualHosts are matched. With the default value (set via the webproxy role) being the contents of the file `/etc/local/varnish/default.vcl`, this serves as a means of defining the entire configuration there without the additional functionality of the NixOS module system. This was was possible before but limited to cases where no virtualHosts was defined in the NixOS configuration. If any virtualHost was defined and the above file present with the webproxy role enabled, the entire configuration would fail to evaluate.

With this PR, the above file - or rather it's contents - now also serve as a fallback VCL without having to assign the "default" virtualHost a condition which always evaluates to true and a name that happens to be always end up after the other virtualHosts alphabetically sorted and thereby relying on the fact that Nix sorts dictionary key alphabetically which just happens to be an implementation detail since according to the reference manual dictionaries have no defined ordering (https://nix.dev/manual/nix/2.18/language/values.html#attribute-set).

For cases where the above file is not present, the previous behavior of returning a 503 is retained if none of the defined virtualHost's conditions match or none are defined.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

- changed behavior regarding the inclusion of the file `/etc/local/varnish/default.vcl` as part of the varnish configuration 

- The file was also included previously and in cases where it was left behind after a migration to a nix based configuration the system would fail to evaluate.

### DIFF
--- a/nixos/services/varnish/default.nix
+++ b/nixos/services/varnish/default.nix
@@ -5,6 +5,7 @@
   ...
 }:
 let
+  inherit (config) fclib;
   cfg = config.flyingcircus.services.varnish;
   vcfg = config.services.varnish;
   vadm = "${vcfg.package}/bin/varnishadm";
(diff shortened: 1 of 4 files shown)


nixos/services/varnish/default.nix
```diff
@@ -129,6 +139,25 @@ in
       type = types.separatedString " ";
       default = "";
     };
+    # cfg.fallbackConfig is defined seperately to ensure proper ordering when assembling the final configuration.
+    # Unlike lists dictionaries do not have a defined ordering, which is why the fallback config is defined seperately.
+    # This ensures that it comes last in the finished varnish configuration - just as the name implies
+    fallbackConfig = mkOption {
+      type = types.str;
```
leona-ya: here, `types.lines` would be slightly better because of merging behavior

PhilTaken: Fair, added

ctheune (approved):

PR #1877, opened 2025-10-14
Add `--disk-size` argument in the fc-devhost for disk sizing

elikoga:

FC-48241

## Release process

- [x] PR has internal ticket

## Design notes

This *is* a feature toggle.

## Security implications

Defensive coding and such. However, there is one specific potential thing we decided to not do (now): we are not giving an upper limit for the size of the disk. This is a development host and even though it has impact on productivity, there is no clear hard upper limit that we could introduce that wouldn't potentially hamper the productivity, too.

We do not expect disastruous failure if a user creates an overly eager devhost VM disk size, just minor cleanups.

manual testing on largo01 and relying on the existing integration tests

### DIFF
--- a/doc/src/devhost.md
+++ b/doc/src/devhost.md
@@ -92,6 +92,7 @@ host = dev.example.com
 release = https://my.flyingcircus.io/releases/metadata/fc-25.05-staging
 # Older batou versions (<=2.5.0) only support the hydra_eval attribute instead of `release`
 # hydra-eval = 309628
+disk_size = 50G
 
 [host:myvm]
 provision-dynamic-hostname = True
@@ -150,6 +151,10 @@ through access to the UI of your applications and act similar to port forwards
 for port 443 -> 443. You should use self-signed certificates within the
 vms. (`batou_ext.ssl.Certificate` already allows switching between
 custom certificates).
+
+Using `provision-disk-size` allows you to specify the disk size for the VM
+(e.g., `25G`, `50G`, `100G`). If not specified, VMs default to 25G. The VM's
+filesystem will be automatically expanded to use the full disk space on boot.
 
 As the development VMs are not managed by our inventory you need to place relevant
 information about roles in a Nix expression file. You can then use a
--- a/doc/src/devhost.md
+++ b/doc/src/devhost.md
@@ -152,10 +152,6 @@ for port 443 -> 443. You should use self-signed certificates within the
 vms. (`batou_ext.ssl.Certificate` already allows switching between
 custom certificates).
 
-Using `provision-disk-size` allows you to specify the disk size for the VM
-(e.g., `25G`, `50G`, `100G`). If not specified, VMs default to 25G. The VM's
-filesystem will be automatically expanded to use the full disk space on boot.
-
 As the development VMs are not managed by our inventory you need to place relevant
 information about roles in a Nix expression file. You can then use a
 provisioning script {file}`provision.sh` to customize the VMs during
--- a/doc/src/devhost.md
+++ b/doc/src/devhost.md
@@ -148,8 +148,8 @@ become available as `<alias>.<vm>.dev.example.com` and are protected
 with Let's Encrypt certificates automatically. They are intended to pass
 through access to the UI of your applications and act similar to port forwards
 for port 443 -> 443. You should use self-signed certificates within the
-vms. (`batou_ext.ssl.Certifiate` already allows switching between
-custom )
+vms. (`batou_ext.ssl.Certificate` already allows switching between
+custom certificates).
 
 As the development VMs are not managed by our inventory you need to place relevant
 information about roles in a Nix expression file. You can then use a
--- a/nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py
+++ b/nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py
@@ -457,7 +457,7 @@ class Manager:
                     )
                     break
 
-            if current_size is not None:
+            if current_size is not None and self.cfg.get("disk_size"):
                 new_size_bytes = parse_disk_size(self.cfg["disk_size"])
                 if new_size_bytes < current_size:
                     raise ValueError(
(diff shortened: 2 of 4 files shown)


elikoga:
I'm not sure wether down-sizing VMs would be in-scope

leona-ya:
down-sizing VMs must be blocked, as XFS as filesystem does not support this

nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py
```diff
@@ -281,6 +320,35 @@ def ensure(
             self.data_dir.mkdir(parents=True, exist_ok=True)
             VM_BASE_IMAGE_DIR.mkdir(parents=True, exist_ok=True)
             vm_has_image = os.path.isfile(self.image_file)
+
```
ctheune, line 322: Are you sure this applies properly if the VM is already running?

elikoga: No, while we could in theory live-migrate this (using block_resize in qemu monitor), I'll add a shutdown and restart for the resize.

elikoga:
It works, however I see some service failing to start while testing which I suspect are from changes from 25.05 to 25.11

doc/src/devhost.md
```diff
@@ -148,8 +149,12 @@ become available as `<alias>.<vm>.dev.example.com` and are protected
 with Let's Encrypt certificates automatically. They are intended to pass
 through access to the UI of your applications and act similar to port forwards
 for port 443 -> 443. You should use self-signed certificates within the
-vms. (`batou_ext.ssl.Certifiate` already allows switching between
-custom )
+vms. (`batou_ext.ssl.Certificate` already allows switching between
+custom certificates).
+
+Using `provision-disk-size` allows you to specify the disk size for the VM
```
ctheune: this seems inconsistent

nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py
```diff
@@ -174,6 +209,7 @@ def ensure(
         memory,
         aliases,
         location,
+        disk_size=None,
```
ctheune: the correct way to evolve an api here would be:

* either append the new kw argument at the end,
* or force (via, `,**,`) that remaining arguments must be keywords

Otherwise this is an invalid API evolution that can break in unobvious ways.

nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py
```diff
@@ -228,6 +264,9 @@ def ensure(
         self.cfg["memory"] = memory
         self.cfg["aliases"] = aliases
         self.cfg["location"] = location
+        self.cfg["disk_size"] = (
```
ctheune: this default should move to the function definition

nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py
```diff
@@ -525,7 +628,8 @@ def proc(self):
                 # pid file may contain trailing lines with garbage
                 for line in p:
                     proc = psutil.Process(int(line))
-                    marker = "{name},process=kvm.{name}".format(name=self.name)
+                    # marker = "{name},process=kvm.{name}".format(name=self.name)
+                    marker = self.name  # i was unable to find the marker above in the cmdline of the qemu process
```
ctheune: yeah, this was copied from fc.qemu but our command line is different.

here's the better option:

```
marker = f" -name {self.name} "

File: nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py (hunk 1 above)
Comment (ctheune) [reply nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py:322]: Right, ok, so this is a *lot* easier if you go down the qmp route and use the `query-block` and `block_resize` commands. It then doesn't matter whether we're bootstrapping the vm or whether it's already running.

Something to double check, though: batou should trigger an fc-resize on every provisioning (every deployment basically). You *could* add that call to the provision bootstrap script in batou, probably.

**Review by ctheune — CHANGES_REQUESTED** [review]

File: nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py (hunk 3 above)
Comment (elikoga) [reply nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py]: Moved to last argument

File: nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py (hunk 4 above)
Comment (elikoga) [reply nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py]: default is now None, since the 25G default was picked up from the image size. It is not needed.

File: nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py (hunk 5 above)
Comment (elikoga) [reply nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py]: Done

**Comment by elikoga** [comment]
I added the call to fc-resize-disk in the fc-devhost command because it already accesses the VM via ssh and it returns the VM to the user in an already provisioned state. Batou could use /var/lib/devhost/ssh_bootstrap_key but since it's already used here we'll continue these operations here. This also makes `fc-devhost ensure` fit the verb `ensure` better

File: nixos/roles/devhost/fc.devhost/src/fc/devhost/manager.py [hunk 6]
```diff
@@ -394,6 +433,61 @@ def ensure(
                 else:
                     time.sleep(0.5)
 
+            # wait for port 22 to accept connections
+            print("Waiting for SSH to become available ...")
+            while True:
+                sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
ctheune, line 443: this needs a timeout

ctheune: also see the block above. 1 min seems like reasonable guess?

ctheune (approved):

ctheune:
This is much more urgent as the customer has been waiting for a while already and is blocked.

PR #1906, opened 2025-10-22
Add monitoring for underlay BGP sessions

sysvinit:
We have occasional recurring problems where after a reboot some hardware machines aren't able to bring the BGP sessions on their underlay interfaces up after a reboot, due to a suspected problem on the switch side. This change adds an additional sensu check to ensure that the configured BGP sessions are actually up, which can then be fed into the statuspage for better visibility over this problem.

PL-133897

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Better visibility into network functions to allow problems to be recognised and treated promptly.
  - Check script manually tested and verified in dev.

### DIFF
--- a/pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
+++ b/pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+"""Check that FRR BGP sessions are up."""
+
+import argparse
+import json
+import subprocess
+import sys
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-i",
+        "--interface",
+        dest="interfaces",
+        action="append",
+        default=[],
+        help="Check BGP session for named interface",
+    )
+    parser.add_argument(
+        "-c",
+        "--critical-age",
+        default=60,
+        metavar="SECONDS",
+        help="Report critical status if session not in established state for longer than SECONDS seconds",
+    )
+
+    args = parser.parse_args()
+
+    data = subprocess.check_output(["vtysh", "-c", "show bgp summary json"])
+    data = json.loads(data.decode("utf-8"))
+    data = data["ipv4Unicast"]["peers"]
+
+    oks = set()
+    errors = {}
+    warnings = {}
+    for iface in args.interfaces:
+        if iface not in data:
+            errors.update({iface: "no bgp session configured for interface"})
+            continue
+
+        if data[iface]["state"] == "Established":
+            oks.add(iface)
+            continue
+
+        age = int(data[iface]["peerUptimeMsec"] / 1000)
+
+        if age <= args.critical_age:
+            warnings.update(
+                {iface: f"bgp session not established since {age} seconds"}
+            )
+        else:
+            errors.update(
+                {iface: f"bgp session not established since {age} seconds"}
+            )
+
+    if errors:
+        print("SESSIONS CRITICAL - {}".format(", ".join(errors.keys())))
+    elif warnings:
+        print("SESSIONS WARNING - {}".format(", ".join(warnings.keys())))
+    else:
+        print("SESSIONS OK - {}".format(", ".join(args.interfaces)))
+
+    for iface, error in errors.items():
+        print(f"CRITICAL {iface}: {error}")
+    for iface, warning in warnings.items():
+        print(f"WARN {iface}: {warning}")
+    for iface in oks:
+        print(f"OK {iface}")
+
+    if errors:
+        sys.exit(2)
+    elif warnings:
+        sys.exit(1)
+    else:
+        sys.exit(0)
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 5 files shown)


pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
```
osnyx, line 1: This impure shebang is probably a leftover from development? It's not needed and bad design in addition, please remove.

pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
```diff
@@ -0,0 +1,80 @@
...
+        "-c",
+        "--critical-age",
+        default=60,
+        metavar="SECONDS",
+        help="Report critical status if session not in established state for longer than SECONDS seconds",
+    )
+
+    args = parser.parse_args()
+
+    data = subprocess.check_output(["vtysh", "-c", "show bgp summary json"])
```
osnyx, line 30: I do not see anything in the packaging that ensures the availability of vtysh as part of the closure, please add as a dependency to default.nix.

pkgs/fc/sensuplugins/fc/sensuplugins/bgp_sessions.py
```diff
@@ -0,0 +1,80 @@
...
+    for iface in args.interfaces:
+        if iface not in data:
+            errors.update({iface: "no bgp session configured for interface"})
+            continue
+
+        if data[iface]["state"] == "Established":
+            oks.add(iface)
+            continue
+
+        uptime = int(data[iface]["peerUptimeMsec"] / 1000)
```
osnyx: What kind of `uptime` is this, uptime of the interface or of the bgp session? Depending on that, this can read to me like we're still issuing warnings or errors despite there being a session with a certain uptime. This is counterintuitive to me.

leona-ya (dismissed):

pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
osnyx: Actually, I see a counter reason against adding this to the pile in fc.sensuplugins: Adding frr to the closure of sensuplugins also pulls it into VMs, which do not yet need it as of now.
Maybe you want to package this as a standalone script? I suggest looking at `fclib.python3BinFromFile`.

osnyx (changes requested):

pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
sysvinit: Also present in other bits of sensuplugins, but yeah I can strip it out.

pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
sysvinit: Good point, I'll shuffle that around into its own package then. I didn't know we had `python3BinFromFile`, that would be useful for a few of the other network monitoring scripts.

pkgs/fc/sensuplugins/fc/sensuplugins/bgp_sessions.py
sysvinit: `peerUptimeMsec` is a counter for how long the BGP session has been in its current state. This will be reset if the session goes down and then count up from zero again, so really "uptime" is a bit of a misnomer – the JSON API of FRR is like that in a lot of places.

osnyx: Ah, so singling out all interfaces with a properly functioning session is already done above with `oks.add; continue`, this here is just applying a grace period before alerting as critical. Understood.

pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
sysvinit: I've ultimately left the shebang in, as I've repackaged it as a standalone python script. The shebang will be rewritten by the standard nixpkgs fixup setup.

pkgs/fc/check-bgp-sessions/check_bgp_sessions.py
sysvinit: Fixed, this is now a separate package, which has FRR in the PATH via wrapper.

(`python3BinFromFile` being part of `fclib` means that it's not usable in our package overlay. Having the check scripts being standalone packages instead of part of the role is occasionally useful for writing tests.)

pkgs/fc/sensuplugins/fc/sensuplugins/bgp_sessions.py
sysvinit: Fixed, renamed to `age`.

pkgs/fc/check-bgp-sessions/default.nix
```diff
@@ -0,0 +1,35 @@
+{
+  lib,
+  stdenv,
```
osnyx, line 3: informational: This *could* be stdenvNoCC instead, but the impact is basically nonexistent here.

osnyx (approved):

PR #1934, opened 2025-11-06
Update WHQ router configs for new uplink topology

sysvinit:
This change updates the WHQ router configuration for the new uplink topology, where the shared layer 2 `tr-whq-sl` network has been replaced with two `tr-up-a` and `tr-up-b` point-to-point links between the routers and the upstream. Routing is now managed using BGP instead of keepalived as well.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - Physical redundancy between WHQ and the upstream router.
  - These configs are already running live from a dev checkout.

### DIFF
--- a/nixos/roles/router/bird2/whq.conf
+++ b/nixos/roles/router/bird2/whq.conf
@@ -1,38 +1,58 @@
-protocol static uplink_whq_4 {
+# host config must include: PEER_IPV4, PEER_IPV6
+
+protocol static local_whq_4 {
   ipv4 {
     table master4;
   };
 
-  route 0.0.0.0/0 via 185.105.255.1%brfe onlink yes;
+  route 185.105.255.0/25 unreachable;
+  route 185.105.255.160/27 unreachable;
+  route 185.105.255.192/27 unreachable;
 }
 
-protocol static uplink_whq_6 {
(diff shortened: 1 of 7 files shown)


nixos/roles/router/bird2/whq.conf
```diff
@@ -1,38 +1,58 @@
...
   scan time 60;
 }
 
-protocol bfd {
-  strict bind on;
-  interface "ethtr";
+# bfd config in host-specific config in this location because the
+# parser does not allow using defines for interface names.
+#
+# protocol bfd {
```
osnyx, line 29: This is meant to be an explanation why something you'd expect here is only defined somehow in line 92? Or what is the reason for keeping this commented out part still in?

osnyx (commented):
Not my main field of expertise, so mostly LGTM, but that one leftover comment makes me a bit suspicious.

nixos/roles/router/bird2/whq.conf
sysvinit: This is here for future us so we know that there's additional configuration needed which has to be defined elsewhere, with an example of how that should look.

osnyx (approved):

PR #1958, opened 2025-11-17
nixos/postgresql: remove postgresql13 role, add postgresql18 role

leona-ya:

PL-134199 PL-133850
## Release process

  will be handled in upgrade notes

- [x] PR has internal ticket

## Design notes

## Security implications

  - Upgrades tested in automated tests
  - checksums are supported since PG14, so we have no problem rolling back


pkgs/fc/agent/src/fc/util/postgresql.py
```diff
@@ -520,6 +528,15 @@ def retrieve_shared_preload_libraries_setting(
                     return line.split("=")[1].strip()
 
 
+def run_pg_checksums_enable(log, old_bin_dir: Path, old_data_dir: Path):
+    log.info("enable-checksums")
+    try:
+        run_as_postgres([old_bin_dir / "pg_checksums", "-D", old_data_dir, "--enable"], check=True)
```
osnyx: I assume this is idempotent when run on clusters with already enabled checksums?

osnyx (dismissed):

pkgs/fc/agent/src/fc/util/postgresql.py
leona-ya: Good point, I thought so, but haven't checked that. I updated this part of code.

ctheune:
Only skimmed, but looks generally good. 

One question: what's the "unattended upgrade" story for k3s servers? Will this just break installations and have them start with a fresh database or is there something that ensures we upgrade?

leona-ya:
We already have auto-updates just by changing the role version on the next rebuild. This also takes effect here.

osnyx (approved):

leona-ya (commented):
Seems good to me.

ctheune (approved):

PR #1969, opened 2025-11-21
nginx: Add certificate validation for non-acme vhosts

dpausp:
- replace shell-based cert checker from ACME module with reusable Python implementation
- check-tls-cert command is also globally installed for manual cert checks

PL-134018

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - monitoring: detect invalid certs used in nginx vhosts 
  -  verified on a test VM that file-based and HTTPS-based Sensu checks stil work for ACME cert
  -  verified on a test VM that file-based and HTTPS-based Sensu checks also work for non-ACME certs

### DIFF
--- a/pkgs/fc/check-tls-cert/default.nix
+++ b/pkgs/fc/check-tls-cert/default.nix
@@ -1,0 +1,39 @@
+{
+  lib,
+  openssl,
+  buildPythonApplication,
+  pytestCheckHook,
+  setuptools,
+}:
+buildPythonApplication {
+  pname = "check-tls-cert";
+  version = "1.0";
+  pyproject = true;
+  src = ./.;
+
+  __structuredAttrs = true;
+
+  build-system = [ setuptools ];
+
+  nativeCheckInputs = [
+    pytestCheckHook
+  ];
+
+  makeWrapperArgs = [
+    "--prefix"
+    "PATH"
+    ":"
+    "${lib.makeBinPath [ openssl ]}"
+  ];
+
+  preCheck = ''
+    export PATH="${openssl}/bin:$PATH"
+  '';
+
+  meta = with lib; {
+    description = "TLS certificate checker for monitoring";
+    homepage = "https://github.com/flyingcircusio/fc-nixos";
+    license = licenses.mit;
+    platforms = platforms.unix;
+  };
+}
--- a/changelog.d/20251121_180304_PL-134018-check-non-acme-certs_scriv.md
+++ b/changelog.d/20251121_180304_PL-134018-check-non-acme-certs_scriv.md
@@ -1,0 +1,16 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### NixOS XX.XX platform
+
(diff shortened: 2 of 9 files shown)


leona-ya:
What does the python implementation does **technically** better? It isn't such a huge mess as before, but still quite a lot of code.

dpausp:
> What does the python implementation does **technically** better? It isn't such a huge mess as before, but still quite a lot of code.

The shell script is broken: there's some extra parsing needed now because the exit code of openssl -checkend is unreliable as of OpenSSL 3.6.0. 

The rest is mostly better output, for all cases: how long the cert is still valid for, hostname, file does not exist, expired or will expire soon, like: `OK: Certificate for tstest01.fe.rzob.fcio.net is valid for more than 25 days. (notAfter=Jan 18 18:39:41 2026 GMT)`. Makes it easier to see what's actually wrong which is more important for arbitrary user-specified certs. Before, we only checked ACME certs which are less error-prone. 

There's some boilerplate which is very similar to other Python sensu checks both not worth the refactoring effort right now, I think.

pkgs/fc/check-tls-cert/default.nix
```diff
@@ -0,0 +1,35 @@
...
+}:
+buildPythonApplication {
+  pname = "check-tls-cert";
+  version = "1.0";
+  pyproject = true;
+  src = ./.;
+
+  build-system = [ setuptools ];
+
+  buildInputs = [
```
osnyx: This does not propagate `openssl` to the runtime PATH and only works coincidentally because `openssl` is also globally installed on most systems.
Needs to be either `propagatedBuildInputs`, or even better, solved with wrapping as nixpkgs python contribution guidelines recommend:

> Non-python runtime dependencies should be added via explicit wrapping or patching (using e.g. substituteInPlace), rather than through propagation via dependencies/propagatedBuildInputs, to reduce clutter in $PATH.

pkgs/fc/check-tls-cert/default.nix
```diff
@@ -0,0 +1,35 @@
...
+  preCheck = ''
+    export PATH="${openssl}/bin:$PATH"
+  '';
+
+  meta = with lib; {
+    description = "TLS certificate checker for monitoring";
+    homepage = "https://github.com/flyingcircusio/fc-nixos";
+    license = licenses.bsd3;
+    platforms = platforms.unix;
+    maintainers = with maintainers; [ ];
```
osnyx: no-op boilerplate

pkgs/fc/check-tls-cert/default.nix
```diff
@@ -0,0 +1,35 @@
...
+  ];
+
+  preCheck = ''
+    export PATH="${openssl}/bin:$PATH"
+  '';
+
+  meta = with lib; {
+    description = "TLS certificate checker for monitoring";
+    homepage = "https://github.com/flyingcircusio/fc-nixos";
+    license = licenses.bsd3;
```
osnyx: How come you chose that particular license?

osnyx (changes requested):
Overall good, but I have a few details that I need answers for or changes.

pkgs/fc/check-tls-cert/default.nix
osnyx: I've decided to already address this myself and changed this.

pkgs/fc/check-tls-cert/default.nix
dpausp: some of our own tools use BSD3 but I've changed it to MIT here to fit the general license for the project

osnyx (approved):

PR #2118, opened 2026-01-19
jitsi: fix and cleanup role

leona-ya:
Remove vendored upstream modules while still using our watchdog mechanism. This also fixes the problems we had with prosody.

PL-135075

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - not more attack surface than before
  - tested on test machine
  - currently no automated test (see PL-134009)


ctheune (commented):
I like. I have no idea whether the xmpp stuff is relevant, though.

nixos/roles/jitsi/default.nix
```diff
@@ -267,29 +255,23 @@ in
           inherit (cfg) resolution;
           startVideoMuted = 8;
           stunServers = [ ];
-          fileRecordingsEnabled = cfg.enableRecording;
-          liveStreamingEnabled = cfg.enableLivestreaming;
+          recordingService.enabled = cfg.enableRecording;
+          liveStreaming.enabled = cfg.enableLivestreaming;
```
osnyx, line 263: If we know this was broken the last time we checked, we might want to add a code comment or eval warning to this.
Maybe upstream will have magically solved this (I see no nixpkgs issue for this though) once we need it next, but if not it might save us a surprise.

nixos/roles/jitsi/default.nix
```diff
@@ -363,35 +314,22 @@ in
...
-          if credentials_directory then
-            -- FIXME when prosody is updated to v13, use
-            --   turncredentials_secret = Credential("turncredentials_secret")
-            local secret = assert(io.open(credentials_directory .. "/turncredentials_secret", "r"))
-            turncredentials_secret = secret:read("*a"):gsub("\n*$", "")
-            secret:close()
-          else
-            print("Warning: Running outside of systemd service context, `turncredentials_secret` will not be set")
-          end
+          turncredentials_secret = Credential("turncredentials_secret")
```
osnyx, line 336: Did you manually check that `prosodyctl` can be invoked on a test machine?
I noticed that we do not have any NixOS tests for *our* role, this could be a candidate to get started.

osnyx (commented):
Quoting the [nixpkgs docs](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/web-apps/jitsi-meet.md) on jitsi-meet:

> Jitsi Meet depends on the Prosody XMPP server only for message passing from the web browser while the default Prosody configuration is intended for use with standalone XMPP clients and XMPP federation. If you only use Prosody as a backend for Jitsi Meet it is therefore recommended to also enable {option}services.jitsi-meet.prosody.lockdown option to disable unnecessary Prosody features such as federation or the file proxy.

Have you considered doing that? At a first glance on what that option does, this looks sensible.

nixos/roles/jitsi/default.nix
leona-ya: yes, works fine (test46 btw)

leona-ya:
enabled and tested prosody lockdown mode.

osnyx (approved):

PR #2123, opened 2026-01-20
Add support for dual-stack k3s clusters

sysvinit:
This change introduces support for k3s Kubernetes clusters with dual-stack networking. In particular, this adds a NixOS option which, when enabled, configures k3s to set up the cluster with dual stack networking. This flag is disabled by default in 25.11, but will become enabled by default starting in 26.05. This flag can only be set before the cluster is first created, and changing its value afterwards is unsupported and will break the cluster. Migration requires wiping and rebuilding the cluster.

The flag ensures that each node in the cluster advertises both IPv4 and IPv6 node IPs, and configures the control plane with dual-stack cluster DNS and both IPv4 and IPv6 subnets for pods and services.

In addition, the tests have been refactored to remove hard-coded address family specific logic. Container image side-loading has been refined to use a mechanism similar to the upstream `services.k3s.images` NixOS option, and there is a new subtest to check the webgateway integration in haproxy functions correctly.

PL-133774

## Release process

- [x] PR has internal ticket

## Design notes

  - Master `flyingcircus.kubernetes.networking.enableIPv6` flag. Off by default for now to avoid impact on existing clusters, will on by default in the next major release.
  - Flag usage documented in role docs.

## Security implications

  - Dual-stack networking is desirable for interoperability, however we should not break existing clusters.
  - We introduce some new attack surface, but it's limited to the v6 masqueraded traffic which we consider irrelevant (it's needed for things to function and we don't see additional new risk here). The IPv6 network is a similar "iptables overlay" so that the network isn't easily reachable from neighbouring RGs. 
  - Manually verified that provisioning of both single and dual-stack clusters works correctly in a test environment.
  - Manually verified that changing the `enableIPv6` flag after cluster creation will irrecoverably break the cluster(!).
  - Hydra tests refactored to be independent of address family, with parameterisation to check both the single-stack and dual-stack cases.

### DIFF
--- a/changelog.d/20260121_170410_PL-133774-k3s-ipv6-support_scriv.md
+++ b/changelog.d/20260121_170410_PL-133774-k3s-ipv6-support_scriv.md
@@ -1,0 +1,32 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+<!-- Impact means "when this change is rolled out, there
+     might be interruptions/downtimes/required actions/... that
+     IMPACT THE RUNNING APPLICATION NEGATIVELY.
+
+     Having new features or changed is not an "impact". That's what
+     the main changelog (see below) is for.
+     -->
+
+
+### NixOS XX.XX platform
+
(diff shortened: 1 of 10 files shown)


sysvinit:
@PhilTaken will the option definitions in `k3s/default.nix` work properly with our NixOS option search as-is, or do I need to go back and add `defaultText` entries?

sysvinit:
Phil has said in chat that in the current state this looks okay.

ctheune (approved):
LGTM

PR #2124, opened 2026-01-20
fc-ceph maintenance leave: stagger osd recovery

osnyx:
Stagger the removal of the `noup` flag on individual OSDs when leaving maintenance. This reduces the peering storm stress on the cluster. We hope that this can resolve the stuck, inactive PGs encountered in PL-133952.

To allow migration from the previous mechanism, we keep unsetting the noup flags on the host-level crush nodes. This will become a no-op after all hosts have migrated to this new code, then that part can be removed.

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - attempt at enabling ceph host automatic maintenance again
  - as the staggered unset causes a longer runtime of the maintenance leave hook, only consider own OSDs if work is required
  - consider migration from old maintenance hook 
  - keep blanket instructions for leaving the previous maintenance mechanism
  - tested on host in dev cluster:
    - maintenance migration from previous mechanism
    - maintenance cacle in the new mechanism
    - standalone command as well as fc-agent hook integration
  - extended and updated test coverage

### DIFF
--- a/pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
+++ b/pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
@@ -1,6 +1,7 @@
 """Configure pools on Ceph storage servers according to the directory."""
 
 import json
+import re
 import socket
 import subprocess
 import sys
@@ -13,7 +14,7 @@ from fc.ceph.api.cluster import CephCmdError
 from fc.ceph.maintenance.images_nautilus import (
     load_vm_images as load_vm_images_task,
 )
-from fc.ceph.util import kill, mount_status, run
+from fc.ceph.util import run
 
 
 class ResourcegroupPoolEquivalence(object):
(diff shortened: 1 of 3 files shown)


pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
```diff
@@ -257,8 +274,15 @@ def enter(self):
     def leave(self):
         host_buckets, osd_ids = get_host_crush_buckets()
         if host_buckets:
+            # PL-133952: still needed for getting hosts upgrading from the
```
ctheune, line 277: .. and likely refactor/ream get_host_crush_buckets() ... or just make the host_bucket assignments_

pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
```diff
@@ -257,8 +274,15 @@ def enter(self):
...
+            # previous mechanism out of maintenance. Remove later.
             run.ceph("osd", "unset-group", "noup", *sorted(host_buckets))
 
+        for osd_id in sorted(filter_noup_osds(osd_ids)):
+            # remove noup flags in a staggered fashion, to reduce peering storm
+            run.ceph("osd", "unset-group", "noup", str(osd_id))
+            time.sleep(15)
```
ctheune, line 284: We'll need to keep an eye on this. On a host with 24 OSDs this means we'll have 360s until all of them are back.

pkgs/fc/ceph/src/fc/ceph/tests/fc-ceph/test_maintenance.py
ctheune, line 38: if you want a bit more assurance, you could use a mock here and check that we called sleep the right number of times (with the expected parameters)

ctheune (approved):

ctheune (approved):

pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
osnyx: Yes, this is why I introduced the filter instead of continuing to indiscriminately unset all osds at each run.
I do not see how we can get around this as the delays – the staggering – **are** what we want here. Only thing we can do is investigate whether the delays can be shortened a bit.

pkgs/fc/ceph/src/fc/ceph/maintenance/nautilus.py
osnyx: This comment looks a bit incomplete, but I *guess* I get what you want and plan to do that anyways 😄

osnyx:
Merging as-is¹ to give this more time in QA, but I'll reconsider the suggestions when cleaning up the transitory remains of the old maintenance mechanism.

¹once I managed to tame the flaky Hydra test

PR #2173, opened 2026-02-12
[PL-135168] improve lgtm roles

PhilTaken:
PL-135168

add role configuration to the loki and tempo roles and add a new faro frontend role

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

all sensitive http and grpc endpoints are only exposed on the srv network (besides faro)
all relevant services bind to srv addresses

### DIFF
--- a/nixos/roles/faro_frontend.nix
+++ b/nixos/roles/faro_frontend.nix
@@ -1,0 +1,84 @@
+{
+  config,
+  lib,
+  ...
+}:
+let
+  cfg = config.flyingcircus.roles.faro;
+  fclib = config.fclib;
+in
+{
+  options = with lib; {
+    flyingcircus.roles.faro = {
+      enable = mkEnableOption "Flying Circus Grafana Faro Frontend configuration";
+      supportsContainers = fclib.mkEnableDevhostSupport;
+
+      host = fclib.mkRoleOption "faro" {
+        type = lib.types.str;
+        description = "Hostname for the frontend";
+        default = c: c.hostname;
+      };
+
+      api_key = fclib.mkRoleOption "faro" {
+        type = types.str;
+        description = "api key to authenticate metrics sent by the frontend";
+        default = c: c.api_key;
+      };
+
+      allowed_origins = fclib.mkRoleOption "faro" {
+        type = types.listOf types.str;
+        description = "allowed origins for CORS";
+        # NOTE this needs to be passed as a list in the role_configuration json
+        default = c: c.allowed_origins;
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    # See https://github.com/grafana/faro-web-sdk/blob/main/docs/sources/tutorials/quick-start-browser.md
+    environment.etc."alloy/faro-integration.alloy".text = ''
+      loki.process "logs_process_client" {
+          forward_to = [loki.write.fcio_rg_loki.receiver]
+
+          // TODO consider making this processing stage configurable? otherwise scrap it and route output from faro.receiver to loki directly
+          stage.logfmt {
+              mapping = { "kind" = "", "service_name" = "", "app" = "" }
+          }
+
+          stage.labels {
+              values = { "kind" = "kind", "service_name" = "service_name", "app" = "app" }
+          }
+      }
+
+
+      faro.receiver "integrations_app_agent_receiver" {
+          server {
+              listen_address           = "127.0.0.1"
+              listen_port              = 8060
+              cors_allowed_origins     = ${builtins.toJSON cfg.allowed_origins}
+              api_key                  = "${cfg.api_key}"
+              max_allowed_payload_size = "10MiB"
+
+              rate_limiting {
+                  rate = 100
+              }
+          }
+
+          sourcemaps { }
+
+          output {
+              logs   = [loki.process.logs_process_client.receiver]
+              traces = [otelcol.exporter.otlp.fcio_rg_tempo.input]
+          }
+      }
+
+    '';
+
+    services.nginx.virtualHosts.${cfg.vhost} = {
+      http2 = true;
+      enableACME = true;
+      forceSSL = true;
+      locations."/".proxyPass = "http://127.0.0.1:8060";
+    };
+  };
+}
--- a/nixos/roles/loki.nix
+++ b/nixos/roles/loki.nix
@@ -46,7 +46,7 @@ in
 {
   options = with lib; {
     flyingcircus.roles.loki = {
-      enable = mkEnableOption "Flying Circus Grafana Loki server";
+      enable = mkEnableOption "the Flying Circus Grafana Loki server";
       supportsContainers = fclib.mkEnableDevhostSupport;
 
       logRetentionPeriod = mkOption {
@@ -60,20 +60,19 @@ in
         default = { };
         type = types.submodule {
           options = {
-            enable = mkEnableOption "store log data in S3";
-            endpoint = mkOption {
-              description = "HTTP(S) endpoint of S3 storage server";
-              type = types.str;
-              default = "http://rgw.local:7840";
+            enable = mkEnableOption "storing the logs in object storage" // {
+              default = config.flyingcircus.enc ? role_configuration.loki;
             };
-            bucketName = mkOption {
+            endpoint = mkOption {
+              description = "HTTP(S) endpoint of the object store";
+              type = types.str;
+              default = "http://rgw.local:7480";
+              defaultText = "<local rgw address>";
+            };
+            bucketName = fclib.mkRoleOption "loki" {
               description = "S3 bucket name";
               type = types.str;
-            };
-            credentialFile = mkOption {
-              description = "Path to file containing credentials used for authenticating to the S3 server (must be readable by the loki user)";
-              type = types.path;
-              default = "/etc/local/loki/s3.cfg";
+              default = c: c.object_store_bucket;
             };
           };
         };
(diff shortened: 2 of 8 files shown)


nixos/roles/loki.nix
```diff
@@ -64,11 +67,12 @@ in
             endpoint = mkOption {
               description = "HTTP(S) endpoint of S3 storage server";
               type = types.str;
-              default = "http://rgw.local:7840";
+              default = "http://${hostRgwAddress}";
             };
             bucketName = mkOption {
               description = "S3 bucket name";
               type = types.str;
+              default = config.flyingcircus.enc.role_configuration.tempo.s3_bucket;
```
zagy: I suppose this should be loki.s3_bucket?

nixos/roles/loki.nix
```diff
@@ -108,32 +112,98 @@ in
...
+                  "/loki/api/v1/query"
+                  "/loki/api/v1/query_range"
+                  "/loki/api/v1/labels"
+                  "/loki/api/v1/label"
+                  "/loki/api/v1/series"
+                  "/loki/api/v1/index/stats"
+                  "/loki/api/v1/index/volume"
+                  "/loki/api/v1/index/volume_range"
+                  "/loki/api/v1/patterns"
+                  "/loki/api/v1/tail"
```
zagy: Hum. What's the reason for not passing all of /loki/ to loki?

nixos/roles/loki.nix
```diff
@@ -142,70 +212,26 @@ in
...
-        # configuration stubs for multi-process management plane
-        common = {
-          replication_factor = 1;
-          ring.kvstore.store = "inmemory";
-        };
-      };
-    };
+          script = ''
+            export AWS_ACCESS_KEY_ID=$(cat ${config.flyingcircus.encPath} | jq '.role_configuration.loki.s3_access_key')
+            export AWS_SECRET_ACCESS_KEY=$(cat ${config.flyingcircus.encPath} | jq '.role_configuration.loki.s3_secret_key')
```
zagy: I think we should call that `object_store_access_key` and `object_store_secret_key`

nixos/roles/loki.nix
```diff
@@ -64,11 +67,12 @@ in
...
-              default = "http://rgw.local:7840";
+              default = "http://${hostRgwAddress}";
             };
             bucketName = mkOption {
               description = "S3 bucket name";
               type = types.str;
+              default = config.flyingcircus.enc.role_configuration.tempo.s3_bucket;
             };
             credentialFile = mkOption {
               description = "Path to file containing credentials used for authenticating to the S3 server (must be readable by the loki user)";
```
zagy: As far as I understand it this is either not used or not useful this way. I think if one defines it here it should be used and *not* read from role data. But if not defined we use role data. In this case we probably should write the cred file to `/run/…`?

nixos/roles/tempo.nix
```diff
@@ -0,0 +1,142 @@
...
+            };
+            bucketName = mkOption {
+              description = "S3 bucket name";
+              type = types.str;
+              default = config.flyingcircus.enc.role_configuration.tempo.s3_bucket;
+            };
+            credentialFile = mkOption {
+              description = "Path to file containing credentials used for authenticating to the S3 server (must be readable by the tempo user)";
+              type = types.path;
+              default = credentialFile;
```
zagy: See comment for loki

nixos/roles/tempo.nix
```diff
@@ -0,0 +1,142 @@
...
+                local_blocks = {
+                  filter_server_spans = false;
+                  flush_to_storage = true;
+                };
+              };
+            };
+            distributor.receivers.otlp = {
+              protocols = {
+                grpc.endpoint = "0.0.0.0:4317";
+                http.endpoint = "0.0.0.0:4318";
```
zagy: We *probably* want to use SRV here?

nixos/roles/loki.nix
```diff
@@ -9,6 +9,9 @@ let
   cfg = config.flyingcircus.roles.loki;
   fclib = config.fclib;
 
+  inherit (config.flyingcircus) hostRgwAddress;
+  credentialFile = "/etc/local/loki/s3.cfg";
```
zagy: If the user should be able to write here the /etc/local/loki dir must be made writable to group service.

nixos/roles/tempo.nix
```diff
@@ -0,0 +1,142 @@
...
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.flyingcircus.roles.tempo;
+  fclib = config.fclib;
+
+  inherit (config.flyingcircus) hostRgwAddress;
+  credentialFile = "/etc/local/tempo/s3.cfg";
```
zagy: group service, see loki

nixos/roles/faro_frontend.nix
```diff
@@ -0,0 +1,95 @@
...
+              values = { "kind" = "kind", "service_name" = "service_name", "app" = "app" }
+          }
+      }
+
+
+      faro.receiver "integrations_app_agent_receiver" {
+          server {
+              listen_address           = "127.0.0.1"
+              listen_port              = 8060
+              cors_allowed_origins     = ["*"]
```
zagy: This needs to be configurable.

zagy (changes requested):

nixos/roles/loki.nix
PhilTaken: they are not really supposed to write to this file since the source of truth for the access and secret keys is the directory

nixos/roles/faro_frontend.nix
```diff
@@ -0,0 +1,104 @@
...
+{
+  options = with lib; {
+    flyingcircus.roles.faro = {
+      enable = mkEnableOption "Flying Circus Grafana Faro Frontend configuration";
+      supportsContainers = fclib.mkEnableDevhostSupport;
+
+      host = mkOption {
+        type = lib.types.str;
+        description = "Hostname for the frontend";
+        default = config.flyingcircus.enc.role_configuration.faro.hostname;
```
zagy: I think this is too optimistic. We cannot *expect* this data to be in ENC. It should be used if there but I suppose this way the build will fail with some "attribute error" instead of "give me a value for this"?

This applies to all the defaults or accesses to role_configuration.

zagy (changes requested):

nixos/roles/faro_frontend.nix
PhilTaken: Can we not expect that it is included in ENC if the role is set up via the directory and it is marked as a required field with a sane default being provided when adding the role? We can add some extra code here to inform the user that the attribute is missing, that is just a question of what is configured where. I wouldn't expect users enable the role via nix config and then wonder about the error here.

Maybe an evaluation error pointing users towards the directory could be useful.

leona-ya: Currently, e.g. devhost doesn't have this field in it's ENC config. We should at least provide some documented way to deploy roles requiring role config on devhosts.

PhilTaken: This is just the default value, on devhosts this could be configured using the option `flyingcircus.roles.faro.host`.

zagy: The point is that it fails horribly. Having defaults which cannot be computed doesn't seem to be a smart way. So I'd say no default and something like `flyingcircus.roles.faro.host = mkDirectory config.flyingcircus.enc.role_configuration.faro.hostname;` (only if that value exists of course).

nixos/roles/loki.nix
```diff
@@ -108,32 +107,69 @@ in
...
+                active_index_directory = "/var/lib/loki/tsdb-shipper-index";
+                cache_location = "/var/lib/loki/tsdb-shipper-cache";
+              };
+              # log data configuration
+              filesystem.directory = "/var/lib/loki/chunk-store";
+            };
+            compactor = {
+              working_directory = "/var/lib/loki/compactor-workdir";
+              retention_enabled = true;
+              delete_request_store = "filesystem";
```
zagy, line 140: I think this needs to be `aws` to match the storage config (if enabled).

nixos/roles/loki.nix
```diff
@@ -108,32 +107,69 @@ in
...
+            locations."/".proxyPass = "http://127.0.0.1:3100";
           };
-          # log data configuration
-          filesystem.directory = "/var/lib/loki/chunk-store";
-        }
-        // lib.optionalAttrs (cfg.s3.enable) {
+        };
+      }
+
+      (lib.mkIf cfg.s3.enable {
```
zagy: Do we really need this? I mean … we expect S3 to be available automatically.

nixos/roles/loki.nix
PhilTaken: > delete_request_store also needs to be configured when retention is enabled to process delete requests, this determines the storage bucket that stores the delete requests.

This configures a separate store that is specifically used to store delete requests, see https://grafana.com/docs/loki/latest/operations/storage/logs-deletion/. 
Storing those on the file system should not be an issue.

nixos/roles/loki.nix
PhilTaken: I supposed we can remove this part if there is no conceivable scenario where the data should be stored on disk.

nixos/roles/loki.nix
```diff
@@ -57,23 +60,20 @@ in
 
       s3 = mkOption {
         description = "Configure log storage in S3-compatible object store";
-        default = { };
         type = types.submodule {
           options = {
-            enable = mkEnableOption "store log data in S3";
+            enable = mkEnableOption "store log data in S3" // {
```
leona-ya: This will render strangely as `Whether to enable store log data in s3.` in the docs. Either make it fit in this format or use a normal `mkOption`.

nixos/roles/loki.nix
```diff
@@ -108,32 +108,69 @@ in
...
+          virtualHosts."${config.networking.hostName}" = {
+            serverAliases = [
+              (fclib.fqdn { vlan = "srv"; })
+              "${config.networking.hostName}.${config.networking.domain}"
+            ];
+            listen = builtins.map (addr: {
+              inherit addr;
+              port = 3100;
+            }) fclib.network.srv.dualstack.addressesQuoted;
+            locations."/".proxyPass = "http://127.0.0.1:3100";
```
leona-ya: Loki also supports WebSocket on one endpoint, we should proxy that too: https://grafana.com/docs/loki/latest/reference/loki-http-api/#stream-logs

nixos/roles/loki.nix
```diff
@@ -142,70 +179,27 @@ in
...
 
-        # configuration stubs for multi-process management plane
-        common = {
-          replication_factor = 1;
-          ring.kvstore.store = "inmemory";
-        };
-      };
-    };
+          script = ''
+            export AWS_ACCESS_KEY_ID=$(cat ${config.flyingcircus.encPath} | jq '.role_configuration.loki.object_store_access_key')
```
leona-ya: I'd like that we either use `${lib.getExe pkgs.jq}` here or add it `jq` to the path of the systemd unit. The indirection via the global PATH could cause problems in the future.

nixos/roles/tempo.nix
```diff
@@ -0,0 +1,144 @@
...
+  options = with lib; {
+    flyingcircus.roles.tempo = {
+      enable = mkEnableOption "Flying Circus Grafana Tempo server";
+      supportsContainers = fclib.mkEnableDevhostSupport;
+
+      s3 = mkOption {
+        description = "Configure trace storage in S3-compatible object store";
+        type = types.submodule {
+          options = {
+            enable = mkEnableOption "store trace data in S3" // {
```
leona-ya: Here also the rendered description in the docs: `Whether to enable store trace data in S3.` (see above).

nixos/roles/tempo.nix
```diff
@@ -0,0 +1,144 @@
...
+            s3 = {
+              bucket = cfg.s3.bucketName;
+              endpoint = cfg.s3.endpoint;
+              forcepathstyle = true;
+              insecure = true;
+            };
+          };
+        };
+
+        systemd.services.tempo.serviceConfig.Environment = "AWS_SHARED_CREDENTIALS_FILE=${credentialFile}";
```
leona-ya: This should be a list, as then we have the option to override on a VM if needed (or alternatively `systemd.services.tempo.environment`)

nixos/roles/loki.nix
```diff
@@ -142,70 +179,27 @@ in
           };
         };
 
-        compactor = {
-          working_directory = "/var/lib/loki/compactor-workdir";
-          retention_enabled = true;
-          delete_request_store = "filesystem";
-        };
+        systemd.services.loki.serviceConfig.Environment = "AWS_SHARED_CREDENTIALS_FILE=${credentialFile}";
```
leona-ya: This should be a list, as then we have the option to override on a VM if needed (or alternatively `systemd.services.loki.environment`)

nixos/roles/tempo.nix
```diff
@@ -0,0 +1,144 @@
...
+
+        systemd.services.tempo-s3-setup = {
+          wantedBy = [ "default.target" ];
+          before = [ "tempo.service" ];
+
+          script = ''
+            export AWS_ACCESS_KEY_ID=$(cat ${config.flyingcircus.encPath} | jq '.role_configuration.tempo.object_store_access_key')
```
leona-ya: Here also the same comment about jq and PATH.

leona-ya (changes requested):

nixos/roles/loki.nix
```diff
@@ -142,70 +179,27 @@ in
...
 
-        limits_config = {
-          retention_period = (toString cfg.logRetentionPeriod) + "d";
-        };
+        systemd.services.loki-s3-setup = {
+          wantedBy = [ "default.target" ];
```
leona-ya: Please don't use `default.target` here

> For typical unit files please set "WantedBy=" to a regular target (like multi-user.target or graphical.target), instead of default.target, since such a service will also be run on special boots like on system update, emergency boot...
> \- https://man7.org/linux/man-pages/man7/systemd.special.7.html

you could use `wantedBy = [ "loki.service" ]` e.g.

nixos/roles/tempo.nix
```diff
@@ -0,0 +1,144 @@
...
+              forcepathstyle = true;
+              insecure = true;
+            };
+          };
+        };
+
+        systemd.services.tempo.serviceConfig.Environment = "AWS_SHARED_CREDENTIALS_FILE=${credentialFile}";
+
+        systemd.services.tempo-s3-setup = {
+          wantedBy = [ "default.target" ];
```
leona-ya: same comment about `default.target`.

leona-ya (changes requested):

nixos/roles/loki.nix
```diff
@@ -108,32 +108,69 @@ in
...
+        };
 
-        storage_config = {
-          # index file management
-          tsdb_shipper = {
-            active_index_directory = "/var/lib/loki/tsdb-shipper-index";
-            cache_location = "/var/lib/loki/tsdb-shipper-cache";
+        flyingcircus.services.nginx = {
+          enable = true;
+          virtualHosts."${config.networking.hostName}" = {
```
leona-ya: I'd rather use something like `virtualHosts.loki` tbh and explicitly set `serverName = `config.networking.hostName`. Then we don't have the potential for conflicts with other vhosts on VMs.

leona-ya (dismissed):

leona-ya (commented):
Looks good from my side now :)

zagy (approved):

PR #2196, opened 2026-02-24
[PL-133773] Add initial ModSecurity implementation

nichmoe:

- [X] PR has internal ticket
- [X] internal issue ID (PL-…) part of branch name
- [X] internal issue ID mentioned in PR description text
- [X] ticket is on Platform agile board

## Design notes

- [X] Provide a feature toggle if the change might need to be adjusted/reverted quickly depending on context. Consider whether the default should be `on` or `off`. Example: rate limiting.
- [X] Provide warnings in previous platform version and upgrade notes when introducing a breaking change --> no breaking changes

## Security implications

No specific requirements
Manual tests on dev server

This PR adds a minimal initial integration of ModSecurity. It is intended to be properly documented and supported with tests, after a bit of practical testing and further development.

See PL-133773

### DIFF
--- a/nixos/platform/modsecurity/default.nix
+++ b/nixos/platform/modsecurity/default.nix
@@ -1,0 +1,110 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.flyingcircus.modsecurity;
+
+  mkModsecurityConfig =
+    rules:
+    (import ./mk-modsecurity-config.nix) {
+      inherit lib pkgs rules;
+    };
+
+  rulesetOptionsModule = {
+    options = import ./modsecurity-options.nix {
+      inherit config lib pkgs;
+    };
+  };
+
+  rulesetImplModule =
+    { config, name, ... }:
+    {
+      options = with lib; {
+        nginxMode = mkOption {
+          type = (
+            types.enum [
+              "On"
+              "Off"
+              "Transparent"
+            ]
+          );
+          default = "Transparent";
+        };
+
+        transactionID = mkOption {
+          type = types.str;
+          default = "$request_id";
+        };
+
+        configContent = mkOption {
+          type = types.str;
+          internal = true;
+        };
+
+        rulesFile = mkOption {
+          type = types.path;
+          internal = true;
+        };
+
+        nginxConfig = mkOption {
+          type = types.str;
+        };
+
+        mkNginxConfig = mkOption {
+          internal = true;
+        };
+      };
+
+      config = {
+        configContent = mkModsecurityConfig config;
+        rulesFile = pkgs.writeText name config.configContent;
+        nginxConfig = config.mkNginxConfig config.nginxMode "";
+        mkNginxConfig =
+          nginxMode: extraConfig:
+          lib.concatStringsSep ";\n" [
+            (lib.optionalString (nginxMode != "Transparent") "modsecurity ${(lib.toSentenceCase nginxMode)}")
+            "modsecurity_rules_file ${config.rulesFile}"
+            "modsecurity_transaction_id ${config.transactionID}"
+            extraConfig
+          ];
+      };
+    };
+
+  nginxIntegrationSubmodule =
+    { config, name, ... }:
+    {
+      config = lib.mkIf cfg.enable {
+        extraConfig = lib.mkBefore (
+          lib.optionalString (lib.hasAttr name cfg.rulesets) cfg.rulesets.${name}.nginxConfig
+        );
+      };
+    };
+
+in
+{
+  options = with lib; {
+    flyingcircus.modsecurity = {
+
+      enable = mkEnableOption { };
+
+      rulesets = mkOption {
+        type =
+          with lib.types;
+          lazyAttrsOf (submodule [
+            rulesetOptionsModule
+            rulesetImplModule
+          ]);
+        default = { };
+      };
+    };
+
+    services.nginx.virtualHosts = mkOption {
+      type = types.attrsOf (types.submodule nginxIntegrationSubmodule);
+    };
+
+  };
+
+}
--- a/nixos/platform/modsecurity/modsecurity-options.nix
+++ b/nixos/platform/modsecurity/modsecurity-options.nix
@@ -1,0 +1,92 @@
+{ lib, ... }:
+
+with lib;
+
+{
+  debugLog = {
+    path = mkOption {
+      type = types.str;
+      default = "/tmp/debug.log";
+    };
+    level = mkOption {
+      type = types.ints.positive;
+      default = 3;
+    };
+  };
+  audit = {
+    engine = mkOption {
+      type = (
+        types.enum [
+          "On"
+          "Off"
+          "RelevantOnly"
+        ]
+      );
+      default = "RelevantOnly";
+    };
+    # Logrotate for "/var/log/nginx/modsec_*.log" is already configured
+    log = {
+      format = mkOption {
+        type = (
+          types.enum [
+            "JSON"
+            "Native"
+          ]
+        );
+        default = "Native";
+      };
+      parts = mkOption {
+        type = types.str;
+        default = "ABIJDFHZ";
+      };
+      path = mkOption {
+        type = types.str;
+        default = "/var/log/nginx/modsec_audit.log";
+      };
+      type = mkOption {
+        type = (
+          types.enum [
+            "Serial"
+            "Concurrent"
+            "HTTPS"
+          ]
+        );
+        default = "Serial";
+      };
+    };
+  };
+  requestBody = {
+    access = mkOption {
+      type = types.bool;
+      default = true;
+    };
+  };
+  collectionTimeout = mkOption {
+    type = types.ints.positive;
+    default = 600;
+  };
+  extraConfig = mkOption {
+    type = types.lines;
+    default = "";
+    description = ''
+      Rules to append to the config module file verbatim.
+    '';
+  };
+  coreRuleSet = {
+    enable = mkEnableOption "OWASP Core Rule Set";
+    blocking = mkOption {
+      type = types.bool;
+      default = false;
+    };
+    anomalyThreshold = {
+      inbound = mkOption {
+        type = types.ints.positive;
+        default = 5;
+      };
+      outbound = mkOption {
+        type = types.ints.positive;
+        default = 4;
+      };
+    };
+  };
+}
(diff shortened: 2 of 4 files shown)


nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+          type = types.str;
+        };
+
+        mkNginxConfig = mkOption {
+          internal = true;
+        };
+      };
+
+      config = {
+        configContent = mkModsecurityConfig config;
```
osnyx, line 62: Ambiguous naming, which `config` is this supposed to refer to? My intuition would be `cfg`, but you provide `config` which refers to the whole top-level NixOS system config?

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+          lazyAttrsOf (submodule [
+            rulesetOptionsModule
+            rulesetImplModule
+          ]);
+        default = { };
+      };
+    };
+
+    services.nginx.virtualHosts = mkOption {
+      type = types.attrsOf (types.submodule nginxIntegrationSubmodule);
```
osnyx, line 105: What is the reason of deferring this to an own submodule? At least now, the `nginxIntegrationSubmodule` does not even provide an option interface on its own, so it would be straightforward to just inline the code here.

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+        );
+      };
+    };
+
+in
+{
+  options = with lib; {
+    flyingcircus.modsecurity = {
+
+      enable = mkEnableOption { };
```
osnyx, line 91: AFAIK this is the [wrong type signature](https://noogle.dev/f/lib/mkEnableOption), usually an option description goes here. Did this ever evaluate successfully?

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+let
+  cfg = config.flyingcircus.modsecurity;
+
+  mkModsecurityConfig =
+    rules:
+    (import ./mk-modsecurity-config.nix) {
+      inherit lib pkgs rules;
+    };
+
+  rulesetOptionsModule = {
```
osnyx, line 16: Why does the submodule definition have its own file, do you plan to re-use it somewhere else?

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.flyingcircus.modsecurity;
+
+  mkModsecurityConfig =
+    rules:
+    (import ./mk-modsecurity-config.nix) {
```
osnyx, line 12: Why does this function definition have its own file, do you plan to re-use it somewhere else?

nixos/platform/modsecurity/modsecurity-options.nix
```diff
@@ -0,0 +1,92 @@
+{ lib, ... }:
+
+with lib;
```
osnyx, line 3: top-level `with` declarations are [strongly discouraged against](https://discourse.nixos.org/t/removing-most-uses-of-top-level-with/41233).
I know when have plenty of those in our code, but for new modules I prefer starting with a cleaner slate.

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+          with lib.types;
+          lazyAttrsOf (submodule [
+            rulesetOptionsModule
+            rulesetImplModule
+          ]);
+        default = { };
+      };
+    };
+
+    services.nginx.virtualHosts = mkOption {
```
osnyx, line 104: This conflicts with the approach taken in https://github.com/flyingcircusio/fc-nixos/blob/875c6034feb067bb54fcfc021e3390c058143662/nixos/services/nginx/default.nix#L325-L348.

nixos/platform/modsecurity/default.nix
osnyx: It is probably a better idea to simplify the option definitions into one set of options, instead of deferring to multiple submodules, and then puzzel this together in the single `config` block with `mkIf` conditionals.

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.flyingcircus.modsecurity;
```
osnyx, line 8: I suggest moving this downwards to `flyingcircus.services.modsecurity` or, as this is currently nginx-specific, even `flyingcircus.services.nginx.modsecurity`.

osnyx (commented):
I am not really convinced by several architecture decisions taken here, these are a bit unusual at least.
Additionally, please provide an actual example config using this module so I can make sense of what this is used for in practice. Then we can simplify the module together.

nixos/platform/modsecurity/default.nix
nichmoe: It did evaluate, but it will most likely blow up, when something tries to get the description. This needs fixing.

nixos/platform/modsecurity/default.nix
nichmoe: Moving to flyingcircus.services.modsecurity is the best solutions, as ModSecurity can also be used with Apache.

nixos/platform/modsecurity/default.nix
nichmoe: This isn't reused but could be in the future, see ModSecurity for Apache.

nixos/platform/modsecurity/default.nix
osnyx: This `config` attribute name refers to the top-level config of *this submodule scope*. Let's see if this could be renamed or has to be this way.
The general idea here is that the `rulesetImplModule` generates a modsecurity config that is standalone in a way that it could be used in several places like nginx, apache, or other supporting software. This is then expose dthrough the `configContent` option.

According to @dpausp, this is built this way to avoid infinite recursions encoutered when building otherwise. (together with lazyAttrs in line 96).

nixos/platform/modsecurity/default.nix
osnyx: Yes, thhis is the wrong type signature, but currently only causes the `description` to fail to evaluate.

nixos/platform/modsecurity/default.nix
nichmoe: Sensible point, will be set inline.

nixos/platform/modsecurity/default.nix
nichmoe: Can be moved into the rulesetImplModule, or a renamed equivalent.

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+        mkNginxConfig = mkOption {
+          internal = true;
+        };
+      };
+
+      config = {
+        configContent = mkModsecurityConfig config;
+        rulesFile = pkgs.writeText name config.configContent;
+        nginxConfig = config.mkNginxConfig config.nginxMode "";
+        mkNginxConfig =
```
osnyx, line 65: The idea here was to expose this function as a handle for deployments to re-use, enabling more composability.

nixos/platform/modsecurity/default.nix
```diff
@@ -0,0 +1,110 @@
...
+          nginxMode: extraConfig:
+          lib.concatStringsSep ";\n" [
+            (lib.optionalString (nginxMode != "Transparent") "modsecurity ${(lib.toSentenceCase nginxMode)}")
+            "modsecurity_rules_file ${config.rulesFile}"
+            "modsecurity_transaction_id ${config.transactionID}"
+            extraConfig
```
nichmoe, line 71: extraConfig needs a trailing newline.

nichmoe:
```
  flyingcircus.services.nginx.virtualHosts."modsectest.test.fcio.net" = {
    forceSSL = true;
    enableACME = true;
    extraConfig = config.flyingcircus.modsecurity.rulesets."modsectest.test.fcio.net.foo.bar".mkNginxConfig "On" "##foo";
  };
```
Creates
```
modsecurity On;
modsecurity_rules_file /nix/store/mxs83kyjnsv38k91cl6q3ynppjlvrph6-modsectest.test.fcio.net.foo.bar;
modsecurity_transaction_id $request_id;
##fooaccess_log /var/log/nginx/access-modsectest.test.fcio.net.log;
error_log /var/log/nginx/error-modsectest.test.fcio.net.log;
```

nixos/platform/modsecurity/default.nix
osnyx: Okay, I think it still makes sense to expose this for re-use in a deployment, but here is my idea:
- turn this into a pure function: add the ruleset to use as a 3rd argument
- then expose this as a function, similar to like [generators](https://github.com/NixOS/nixpkgs/blob/66d9241e3dc2296726dc522e62dbfe89c7b449f3/lib/generators.nix#L923:C3) are exposed in the NixOS lib code

Because this is not really an *option* in the *option* sense of "it makes use of the NixOS `evalModules` for achieving merging of option values, checking eligible option values, and keeping state. It's just a convenience function.

**Update:** At a closer look, it is apparently not that simple to expose plain non-options under the same naming scope as the surrounding options. Then at least mark the option as `lib.types.raw`.

nixos/platform/modsecurity/default.nix
osnyx: The `mkBefore` approach taken here is probably even preferrable to the `apply` approach in the nginx module, as it composes better.
Just make sure to unify both sides to take the same approach, I suggest adjusting the code in `ngix/default.nix`.

osnyx (changes requested):
After reviewing this together, here are the suggestions for restructuring this:

- Split this into 2 main modules, similar to how `security.acme` defines the certificates that are then used and referenced by several other modules, like in the nginx modules
- `flyingcircus.services.modsecurity` module: merge the rulesetOptionsModule and rulesetImplModule. Specifying a ruleset and then rendering it into an actual modsecurity config format are tightly coupled
  - we might even be able to get rid of the submodule-level `config` in line 62, as this is now a module on its own, not a submodule
- `flyingcircus.`(?)`services.nginx.virtualHosts` receives an option that can then reference these rule sets and configure their particular usage within that nginx host
  - this is then also the place for helpers generating the nginx-specific config embedding

Avoid convoluting submodules unless they are really necessary. The `nginxIntegrationSubmodule` stops being a submodule and instead configures nginx directly.

PR #2201, opened 2026-02-27
[FC-52201] add package lamp_php85

PhilTaken:

FC-52201

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

does not apply

### DIFF
--- a/pkgs/overlay.nix
+++ b/pkgs/overlay.nix
@@ -91,6 +91,18 @@ builtins.mapAttrs (_: patchPhps phpLogPermissionPatch) {
     {
       inherit (super) php83 php84;
     }
+//
+  builtins.mapAttrs
+    (
+      _:
+      patchPhps (fetchpatch {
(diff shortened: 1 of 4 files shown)


pkgs/overlay.nix
```diff
@@ -451,6 +451,17 @@ builtins.mapAttrs (_: patchPhps phpLogPermissionPatch) {
     ]
   );
 
+  lamp_php85 = self.php85.withExtensions (
```
osnyx, line 466: We need to apply patches to php, to allow udo-srv access to slowlog files. See the definition of other php packages above.
If you're not comfortable with doing that, feel free to hand this over to me.

osnyx (changes requested):

pkgs/overlay.nix
PhilTaken: ah I missed the patch up top

PhilTaken: and of course it failed to apply

osnyx (approved):

PR #2235, opened 2026-03-11
[PL-135231] alloy configuration for k3s logs

PhilTaken:

PL-135231

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

- provides configuration that exports k3s pod journal logs to a non-public loki instance
- very similar security requirements to the VM's journal - potentially exposes secrets to another system. the former is considered acceptable risk so this should be too

### DIFF
--- a/nixos/roles/k3s/default.nix
+++ b/nixos/roles/k3s/default.nix
@@ -11,6 +11,8 @@
 }:
 let
   inherit (config.flyingcircus.kubernetes.network) enableIPv6;
+  location = lib.attrByPath [ "parameters" "location" ] "standalone" config.flyingcircus.enc;
+  lokiServer = config.fclib.findOneService "loki-collector";
 in
 {
   imports = with lib; [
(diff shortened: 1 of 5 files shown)


nixos/roles/k3s/default.nix
```diff
@@ -135,5 +137,117 @@ in
           }
         ];
       })
+
+      (lib.mkIf (server && (!builtins.isNull lokiServer)) (
+        let
+          ca_file = "/var/lib/k3s/server/tls/serving-kube-apiserver.crt";
+          cert_file = "/var/lib/k3s/server/tls/client-kube-apiserver.crt";
+          key_file = "/var/lib/k3s/server/tls/client-kube-apiserver.key";
```
sysvinit: Does alloy really need keys and certificates in addition to the bearer token? Telegraf manages just fine with only the latter.

The API server's TLS key is a very big hammer that will grant highly privileged access to the cluster – I think we can probably find a way to grant alloy more limited and better scoped read-only access to the relevant bits of the API.

We could, similar to sensu and telegraf, provide vendor manifests which specify a ClusterRole with the necessary API permissions and then bind this to alloy's ServiceAccount. Upstream should have a kube manifest somewhere which lists the required permissions.

nixos/roles/k3s/default.nix
```diff
@@ -135,5 +137,117 @@ in
...
+                  }
+              }
+
+              forward_to = [loki.write.fcio_rg_loki.receiver]
+            }
+          '';
+
+          systemd.tmpfiles.rules = [
+            # alloy *expects* the token here, a symlink seems to suffice though
+            "L /run/secrets/kubernetes.io/serviceaccount/token - - - - /var/lib/k3s/tokens/alloy"
```
sysvinit: I presume alloy assumes the token to be available at this path because this is the path where the service token would be found if it were running as a pod inside the cluster. I don't *like* this, but if this path really isn't configurable at all (e.g. upstream doesn't expect people to run alloy as a sidecar outside a kube cluster) then this might be the next best solution.

nixos/roles/k3s/server.nix
```diff
@@ -596,8 +595,20 @@ in
...
+        };
+        systemd.services.fc-k3s-token-sensuclient = makeAuthTokenService {
+          user = "sensuclient";
+          secret = "io.flyingcircus.service-token.sensu-client";
+        };
+        systemd.services.fc-k3s-token-alloy = makeAuthTokenService {
+          user = "root";
+          group = "service";
+          tokenname = "alloy";
+          secret = "io.flyingcircus.service-token.alloy";
```
sysvinit, line 644: Does the bearer token for alloy really need to be owned by `root:service`? Alloy runs as its own `alloy` user and group, so we could spare ourselves the effort of making token file permissions configurable.

This probably needs a guard for `lokiServer != null` though, to handle RGs where there's a k3s cluster but no loki server, as alloy won't be deployed and the user won't exist.

sysvinit (changes requested):

nixos/roles/k3s/server.nix
PhilTaken: > Alloy runs as its own alloy user and group, so we could spare ourselves the effort of making token file permissions configurable.

The alloy service runs has `DynamicUser = true;`.
We could add a new group specifically for this token and add it to the service's supplementary groups I guess.

Will add the guard.

sysvinit: Ah, I'd forgotten that dynamic users are a thing – they do tend to make cross-service integration like this a bit of a headache. Would it make our lives easier to turn dynamic users off in the alloy service and then allocate a fixed system user in the role? Or will that cause other problems with existing alloy instances?

Alternatively, would an `ExecStartPre` on alloy which adjusts the permissions be an option?

nixos/roles/k3s/default.nix
PhilTaken: I switched the authentication to bearer tokens, only the ca file is required for TLS now. see the newest commit

sysvinit: Yeah, that looks better. I'm a bit uneasy about modifying the permissions for k3s's data directory to give external processes access to the CA certificate. In the telegraf config we set `insecure_skip_verify` to sidestep this issue, which we could also set here in alloy's configuration – we're connecting to the API server over localhost anyway.

leona-ya: Shouldn't it be possible to use systemd credentials here, then the file in k3s's data dir doesn't need a permission change

sysvinit: That's a good point, but the TLS keys and certificates are only created when k3s runs for the first time, so in terms of correctness we can't a priori rely on the CA certificate file existing when alloy (or telegraf) are started.

I would have otherwise suggested a sidecar which copies the CA certificate out of the k3s data directory into a world-readable location in `/var` or into `/etc/kubernetes`.

sysvinit: Although, now that I'm thinking out loud... given that alloy has a dependency on the unit which creates the bearer token, which also requires k3s to be up and initialised, that point is moot.

PhilTaken: insecure_skip_verify does do the trick with in combination with a bearer token that has the required permissions

nixos/roles/k3s/server.nix
PhilTaken: systemd's loadcredential seems to do the trick here. 

Should I amend the changes to the token generation then? Some of the features that I implemented are not required any more, the only thing that needs to stay is a functional split between name and user/group due to dynamicuser.
Permissions for the alloy token can be root:root, it just needs a more descriptive file name than "root".

sysvinit (approved):

PR #2239, opened 2026-03-12
PL-131408 ceph upgrade to pacific

osnyx:

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - test new cluster functionality (Ceph Pacific)
  -  prevent regressions in Nautilus: NixOS tests run for ceph-pacific and ceph-nautilus in parallel
  - Telemetry Security considerations: confidentiality
    We want different descriptions by cluster. Usually we identify our ceph
    clusters by location + resource group.
    These might reveal customer or project names though. We obscure them
    into an identifier by hashing (SHA512).
    I am aware that an unsalted hash is just a strong obfuscation, given the
    existence of rainbow tables. For projects with such strong
    confidentiality requirements that this is not acceptable, we want to
    disable sending of telemetry data altogether in the first place, which
    is possible via NixOS option. 
  - [x] all automated tests pass
  - [x] Ceph pacific has been running in our dev cluster for the last weeks, monitoring is green


nixos/roles/ceph/rgw.nix
```diff
@@ -11,7 +11,13 @@ let
   fclib = config.fclib;
   role = config.flyingcircus.roles.ceph_rgw;
   enc = config.flyingcircus.enc;
-  inherit (fclib.ceph) expandCamelCaseAttrs expandCamelCaseSection releaseAtLeast;
+  inherit (fclib.ceph) normaliseCephOptionAttrs normaliseCephOptionSection releaseAtLeast;
+  bucketNameValidationPy = fclib.python3BinFromFile ./rgw-check-bucket-names.py {
+    flakeIgnore = [ "E501" ];
+  };
+  rgw-validate-bucket-names = pkgs.writeShellScriptBin "rgw-validate-bucket-names" ''
+    sudo radosgw-admin bucket list | ${lib.getExe bucketNameValidationPy}
```
ctheune: the sudo should be invoked by whoever is calling this, lets not bake this into the command

nixos/roles/ceph/rgw.nix
```diff
@@ -89,32 +87,38 @@ in
...
+        fc-ceph.settings.RgwUserManager = {
+          release = role.cephRelease;
+          path = cephPkgs.fc-ceph-path;
+        };
+        server = {
+          enable = true;
+          cephRelease = role.cephRelease;
+          # no fc-ceph settings necessary so far
+        };
+        # PL-135287
```
ctheune: i'd prefer if we set up new clusters with this set to false, meaning this should be the default and we need to pay attention to set this to true in all our existing clusters.

nixos/services/mail/default.nix
```diff
@@ -16,7 +16,7 @@ with lib;
 # - domains: list of mail domains for which regular mail accounts exist
 
 let
-  inherit (import ../../../versions.nix { }) nixos-mailserver;
+  inherit (import ../../../release/versions.nix { }) nixos-mailserver;
```
ctheune, line 19: ?!?

pkgs/fc/ceph/src/fc/ceph/api/cluster.py
```diff
@@ -1,33 +1,17 @@
...
 
-    def __init__(
-        self,
-        ceph_conf=CEPH_CONF,
-        ceph_id=CEPH_ID,
-        dry_run=False,
-        default_encoding="utf-8",
-    ):
+    def __init__(self, ceph_conf=CEPH_CONF):
+        # XXX customising the ceph_conf is not used or exposde somewhere, consider removing
```
ctheune: typo

pkgs/overlay.nix
```diff
@@ -14,6 +14,19 @@ let
     rev = "a467063b4abb8bd7636d9bb2475edbd2f0e6c6b6";
   };
 
+  nixpkgs-ceph-pacific =
```
ctheune, line 17: das reasoning hier bitte nochmal dranschreiben. am hash kann ich hier nicht erkennen, dass wir das aus einem 23er branch ziehen müssen

pkgs/fc/ceph/batou/conftest-sample.py
```diff
@@ -0,0 +1,74 @@
+def ceph_live_setup():
```
ctheune: this looks like residue

ctheune (changes requested):
generally fine, there's a few things I'd like see fixed, the changing of the default can happen later

nixos/services/mail/default.nix
osnyx: That's necessary because in thhe batou dev deployment, the provisioning step does not copy symlinks properly https://github.com/flyingcircusio/batou/issues/528

This causes eval to break, because versions.nix is just a compatibility symlink to release/versions.nix nowadays.

nixos/roles/ceph/rgw.nix
osnyx: Added this to the upgrade WMS instead and gave the dev cluster RGWs the necessary local config.

ctheune (approved):

PR #2294, opened 2026-04-10
fc-luks: improve detection of encrypted volumes to find unopened ones

osnyx:
In most cases, we want *all* LUKS volumes of the system to be discovered, even if they are not currently opened. That helps
- with rekeying volumes when the keystore failed and they could not be openend in the first place
- avoiding stale keys for e.g. backup volumes

PL-133176

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - prevent stale volumes from getting out of sync after rekeying
  - improve operability of hosts with a failed keystore
  - extended unit tests to cover new scenarios
  - manually tested in dev cluster (on the Ceph pacific dev branch though)

### DIFF
--- a/changelog.d/20260411_014408_PL-133176-improve-luks-detection_scriv.md
+++ b/changelog.d/20260411_014408_PL-133176-improve-luks-detection_scriv.md
@@ -1,0 +1,27 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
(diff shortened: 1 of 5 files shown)


ctheune (approved):
There's a slight tummyache that the "only_active=False" default and logic is actually a double negative: i became confused three times reading this:

- when reading the low level functions using only_active
- when comparing the previous with the new behaviour
- when reading the argparser default

We'll go ahead with this, but I'm getting a beer in roughly a year when this bites us.

PR #2300, opened 2026-04-15
PL-135151 fix interface specific sysctls

dhnasa:
PL-135151

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - fix bug
  - automated test
  - manually tested on test vm

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -120,6 +120,36 @@ let
   concatMapLinesIndent = concatFuncWithIndent lib.concatMapStringsSep;
   concatMapLines = lib.concatMapStringsSep "\n";
 
+  extractSysctlInterface =
+    sysctl:
+    lib.pipe
+      [
+        ''net\.ipv[46]\.conf\.([/[:alnum:]]+)\..+''
+        "net/ipv[46]/conf/([.[:alnum:]]+)/.+"
+        ''net\.ipv[46]\.neigh\.([/[:alnum:]]+)\..+''
+        "net/ipv[46]/neigh/([.[:alnum:]]+)/.+"
+      ]
+      [
+        (lib.map (re: lib.match re sysctl))
+        (lib.findFirst lib.isList [ null ])
+        lib.head
+        (lib.mapNullable (replaceStrings [ "/" ] [ "." ]))
+      ];
+
+  interfaceSysctls = lib.pipe config.boot.kernel.sysctl [
+    lib.attrsToList
+    (map (e: e // { iface = extractSysctlInterface e.name; }))
+    (lib.filter (
+      { value, iface, ... }:
+      value != null
+      && !elem iface [
+        null
+        "default"
+        "all"
+        "*"
+      ]
+    ))
+  ];
 in
 {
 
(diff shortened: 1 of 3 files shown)


nixos/platform/network.nix
```diff
@@ -731,6 +781,36 @@ in
               (map (link: lib.nameValuePair (unitName link) (unitTemplate link false)) virtualLinks)
               ++ (map (link: lib.nameValuePair (unitName link.link) (unitTemplate link.link true)) ethernetLinks)
             )
+          ++ (
+            let
+              sysctlFile =
+                link:
+                pkgs.writeText "sysctl-${link}" (
+                  lib.concatMapStrings (
+                    { name, value, ... }: "${name}=${if value == false then "0" else toString value}\n"
```
sysvinit, line 792: This presumably needs a case for `value == true`, so that it's translated to 1 correctly, right?

nixos/platform/network.nix
```diff
@@ -148,6 +178,26 @@ in
...
+        [
+          {
+            assertion = lib.length invalidSysctls == 0;
+            message = ''
+              The following sysctls refer to nonexisting interfaces: ${
+                lib.concatMapStringsSep ", " ({ name, iface, ... }: "${name} (${iface})") invalidSysctls
+              }
+            '';
+          }
+        ];
```
sysvinit, line 198: This could cause problems with Wireguard interfaces if anyone ever tries to configure sysctls on them, as they don't appear in `networking.interfaces` (the Wireguard configuration is generated and managed by a different NixOS module than the rest of the system network configuration). We don't have Wireguard deployed in many places and I can't think of any scenarios where we would need any sysctls to be set, but I could see this unexpectedly causing problems for Appops if a customer comes along with odd requirements.

Having a warning might be a valuable sanity check which could help identify typos, though on the other hand printing warnings for every evaluation if there's a Wireguard interface with sysctls is also unhelpful.

Maybe there's a solution here in adding a NixOS option to list interface names which should be ignored in the assertion as an escape hatch.

sysvinit (changes requested):
The logic which extracts interface names out of the sysctls is a little mind-bending (not the only code like that in our network configuration!) but looks correct.

This change makes a few other systemd units in `network.nix` which handle sysctl configuration redundant, but that doesn't need to be cleaned up here and now – the entire file could probably do with a separate round of review and refactoring.

I have a few small nits otherwise.

nixos/platform/network.nix
ctheune: Clarification question: my impression here is that generally this is a move in the right direction. I think we could move forward having the assertion in place, when an escape hatch is given in the form of "this doesn't do what you think it does -> do this instead until we support this properly" ... ?

sysvinit: Yeah, it's a move in the right direction – the assertion will catch typos and cases where obvious-seeming configuration might not do the right thing. Having a "this doesn't work, go do that instead" message instead of a list of exceptions is another way to solve the issue.

dhnasa: expanded the error message to include what to do instead

nixos/platform/network.nix
dhnasa: `toString true` already evaluates to "1", while `toString false` evaluates to an empty string for some reason

sysvinit (approved):

PR #2318, opened 2026-04-21
nix: use nix 2.34 and cleanup tarball-cache-v1

leona-ya:
Nix 2.34 is the new default in NixOS 26.05.

According to the relase notes, it now uses tarball-cache-v2, but doesn't cleanup the old directory. This change creates a systemd unit that cleans it up.

PL-135274

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - Tested on dev machine

### DIFF
--- a/nixos/platform/nix.nix
+++ b/nixos/platform/nix.nix
@@ -8,7 +8,7 @@
 let
   production = lib.attrByPath [ "parameters" "production" ] "" config.flyingcircus.enc;
 
-  nixPackage = pkgs.nixVersions.nix_2_28;
+  nixPackage = pkgs.nixVersions.nix_2_34;
 in
 {
   options.flyingcircus = {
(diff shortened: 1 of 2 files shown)


osnyx (changes requested):
There's also no regular maintenance (cleanup of stale cache entries) done for tarball-cache-v2, according to nix release notes.

> Currently, Nix doesn’t perform any maintenance on the ~/.cache/nix/tarball-cache-v2 repository, which will be addressed in future versions. Users that wish to reclaim disk space used by the tarball cache may want to run:
>
> rm -rf ~/.cache/nix/tarball-cache # Historical tarball-cache, not used by Nix >= 2.33
> cd ~/.cache/nix/tarball-cache-v2 && git multi-pack-index write && git multi-pack-index repack && git multi-pack-index expire

Do we want to do something about this, to not fill up disks with old channel revisions?

leona-ya:
I don't really know if we should do something about that. Repack operations are quite IO-intensive and I would suggest that we see if this becomes a problem or not. Also maybe we just should see when Nix fixes this.
Currently, I cannot really see when this directory gets created (it doesn't with local checkouts apparently)

osnyx:
Okay, then let's keep observing this.

osnyx (approved):

PR #2319, opened 2026-04-21
fc-update-channel: stagger execution over 2h instead of 1h

osnyx:
When fc-update-channel encounters a new release, it immediately downloads the corresponding sources and pre-builds the package. We already staggered the individual service execution over a stretch of one hour on a cluster-wide level to reduce strain on the infrastructure.

When rolling out especially large releases, e.g. after a full rebuild due to low-level dependency changes, this strain became so high it impacted VM operations.
Stretching the rollout of such updates over twice the time, reduces load on all involved components. Judging by the metrics, the Ceph storage clusters (both WHQ, most strongly affected by the Hydra fetch) as well as RZOB (most affected due to the high number of production VMs doing updates) were able to cope. The clear bottle neck was the object storage gateway serving the build artifacts, and this gateway now also gets its load reduced.

FC-52891

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - improve availability: reduce infrastructure load during release rollout
  - checked whether the interval is mentioned in our documentation – did not find any mentions, no need to update the docs

### DIFF
--- a/nixos/platform/agent.nix
+++ b/nixos/platform/agent.nix
@@ -528,8 +528,8 @@ in
 
         timerConfig = {
           OnActiveSec = "1m";
-          OnCalendar = "hourly";
-          RandomizedDelaySec = "60m";
+          OnCalendar = "00/2:00";
+          RandomizedDelaySec = "120m";
           FixedRandomDelay = true;
           Persistent = true;
         };
--- a/changelog.d/20260421_100309_FC-52891-stretch-updates_scriv.md
+++ b/changelog.d/20260421_100309_FC-52891-stretch-updates_scriv.md
@@ -1,0 +1,28 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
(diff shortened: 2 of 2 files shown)


nixos/platform/agent.nix
```diff
@@ -529,7 +529,7 @@ in
         timerConfig = {
           OnActiveSec = "1m";
           OnCalendar = "hourly";
-          RandomizedDelaySec = "60m";
+          RandomizedDelaySec = "120m";
```
leona-ya, line 532: This is wrong as `RandomDelaySec > OnCalendar`. This could have two outcomes
- RandomDelaySec < 1h: The service would still run hourly
- RandomDelaySec > 1h: The service never runs, because the first run gets shadowed by the new run triggered by the new hour.

I tested this on a VM.

leona-ya (changes requested):

nixos/platform/agent.nix
osnyx: Thanks for catching this, I plainly missed the `hourly` altogether.

Now it works as expected (tested on a VM).

```
$ systemctl list-timers | grep update
Tue 2026-04-21 14:12:34 CEST 1h 30min Tue 2026-04-21 12:26:00 CEST     16min ago fc-update-channel.timer            fc-update-channel.service
```

leona-ya (approved):

PR #2342, opened 2026-05-05
[PL-135131] mailserver: add an evaluation warning hinting at the new defaults for imap and smtp ports

PhilTaken:

PL-135131

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

- [x] Security requirements tested? (EVIDENCE
- this pr adds a single evaluation warning in the 25.05 release

### DIFF
--- a/nixos/roles/mailserver.nix
+++ b/nixos/roles/mailserver.nix
@@ -106,7 +106,7 @@ let
     };
 
     dynamicMaps = mkOption {
-      description = '''';
+      description = "";
       type = with types; attrsOf (listOf path);
       default = { };
       example = {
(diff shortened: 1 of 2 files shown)


nixos/roles/mailserver.nix
```diff
@@ -304,6 +304,17 @@ in
   config = lib.mkMerge [
 
     (lib.mkIf roles.mailserver.enable {
+      warnings = lib.optionals (config.flyingcircus.platform.version == "25.05") [
```
ctheune, line 307: this looks weird. this only goes to the 25.05 branch, so why the version check here?

also: is there a way for users to silence this when they prepare their config early and have nothing left to do?

ctheune (changes requested):

nixos/roles/mailserver.nix
PhilTaken: Version check could be removed, sure.

Since this mostly affects mail-clients that wish to connect to the server, there is no direct way to turn this off. Adding an option just for this one check that would then have to be removed when switching the VM to the next release seems a bit redundant.

leona-ya:
I agree that this doesn't really make sense. Eval warnings only really make sense when they check if this specific VM is affected. We cannot do this here, so this would just cause alert fatigue

PR #2354, opened 2026-05-12
fc-luks: adjust memory lock test to MAP_DROPPABLE

osnyx:
Since glibc-2.41 [^1], glibc utilises the MAP_DROPPABLE mmap flag for certain getrandom specific state. An updated glibc in the NixOS 26.05 release thus enables this behaviour.
Droppable pages are also never written to swap, but do not officially count as memlocked. Instead, they may be dropped to retain memory.

The flag exists since linux-6.11 [^2] but is still not documented in the kernel man pages. But the glibc docs [^3] mention the `MAP_DROPPABLE` flag.

[^1]: https://inbox.sourceware.org/libc-alpha/6975370.R56niFO833@pinacolada/T/#u
[^2]: https://lwn.net/Articles/983186/
[^3]: https://sourceware.org/glibc/manual/latest/html_node/Memory_002dmapped-I_002fO.html

Necessary for PL-134240 (channel blocker)

## Release process

  - no, internal test change sufficiently documented in the surrounding code (comments)

- [ ] PR has internal ticket

## Design notes

## Security implications

  - unblock test
  - research in documentation that droppable pages have the same security characteristics (not swapping) as memlocked pages before
  - fc-ceph tests pass again

### DIFF
--- a/pkgs/fc/ceph/src/fc/ceph/tests/fc-ceph/test_main.py
+++ b/pkgs/fc/ceph/src/fc/ceph/tests/fc-ceph/test_main.py
@@ -1,4 +1,5 @@
 import os
+import re
 from pathlib import Path
 from sys import argv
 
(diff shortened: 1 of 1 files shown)


osnyx:
As discussed with @ctheune, merging this after integrating and cleaning up his proposal.

PR #2395, opened 2026-05-26
webproxy: fix vinyl-cache-http sensu check

leona-ya:
PL-135368

## Release process

or
  no: follow-up to broken change, not yet released

- [x] PR has internal ticket

## Design notes

## Security implications

  - checked vinyl case on VM
  - extended NixOS test for Vinyl and Varnish case

### DIFF
--- a/nixos/roles/check-varnish-cache-http.py
+++ b/nixos/roles/check-varnish-cache-http.py
@@ -4,7 +4,7 @@ import sys
 ports = []
 
 for line in (
-    subprocess.check_output(["vinyladm", "debug.listen_address"])
+    subprocess.check_output(["varnishadm", "debug.listen_address"])
     .decode("ascii")
     .splitlines()
 ):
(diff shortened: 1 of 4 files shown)


nixos/roles/check-varnish-cache-http.py
```diff
@@ -0,0 +1,40 @@
+import subprocess
+import sys
+
+ports = []
+
+for line in (
+    subprocess.check_output(["vinyladm", "debug.listen_address"])
```
osnyx: This is inconsistent. This file is used when still using `varnish` as role package, but then we also pass that package – containing `varnishadm` – to the check.

osnyx (changes requested):
Please also specify with which variants of the role you tested that on the VM (varnish/ vinyl)

nixos/roles/check-varnish-cache-http.py
leona-ya: fixed.

leona-ya:
I tested on a VM with vinyl. We currently have none with varnish and 26.05. I instead extended the NixOS test

osnyx (approved):

PR #2398, opened 2026-05-28
k3s release notes for 26.05

sysvinit:
This change adds release notes for the k3s roles in 26.05, including the change in defaults for dual stack networking which affect new clusters.

I've additionally added newer k3s versions available in 26.05 to the important packages.

PL-135129

## Release process

or
- [x] Add a upgrade node in `doc/upgrade.md`

- [x] PR has internal ticket

## Design notes

## Security implications

  - Prevent users from breaking their k3s clusters due to interactions between defaults derived from different NixOS stateVersions.
  - Release notes.

### DIFF
--- a/doc/src/kubernetes.md
+++ b/doc/src/kubernetes.md
@@ -429,27 +429,35 @@ on your requirements there are a few more options available:
 
 ### IPv6 support
 
-By default, Kubernetes clusters in our environment are configured with
-single-stack IPv4-only networking. This means that pods are only assigned a single
-IPv4 address, and cannot access external IPv6-only hosts.
+By default, new Kubernetes clusters in our environment are configured with
+dual-stack networking enabled. This means that pods are assigned both an IPv4
+and an IPv6 address, and can access external hosts over both address families.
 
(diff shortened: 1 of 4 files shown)


doc/src/kubernetes.md
```diff
@@ -31,7 +31,7 @@ run in a process called the `k3s agent`. We will prefer to use the words
 `server` and `agent` through the remainder of this document.
 :::
 
-This platform release defaults to using k3s version 1.33.x.
+This platform release defaults to using k3s version 1.35.x.
```
leona-ya: That's not correct I think, as https://github.com/flyingcircusio/fc-nixos/blob/fc-26.05-dev/nixos/roles/k3s/default.nix#L122. I know that this is not optimal, but we only really can bump one minor per large release or maybe long term have no default.

leona-ya (changes requested):

doc/src/kubernetes.md
sysvinit: Good catch, I only looked at the package versions JSON file and assumed we were using the upstream default.

leona-ya (approved):

PR #2411, opened 2026-06-03
ai-api-gateway: add tmpfiles rules for skvaider directories and log cleanup

elikoga:
- /var/lib/skvaider: 0755 skvaider service
- /var/log/skvaider: 0750 skvaider service (no world access)
- ACLs for sudo-srv and admins groups (read + execute)
- /var/lib/skvaider/debug: 0750, cleaned after 24h
- /var/log/skvaider: cleaned after 4 days

Mirrors the permissions fix already applied to ai-model-server (PL-135413) but for the gateway role.

## Release process

or

- [ ] PR has internal ticket

## Design notes

## Security implications

### DIFF
--- a/nixos/roles/ai-api-gateway.nix
+++ b/nixos/roles/ai-api-gateway.nix
@@ -150,7 +150,7 @@ in
       };
     };
     systemd.tmpfiles.rules = [
-      "d /var/lib/skvaider 0755 skvaider service -"
+      "d /var/lib/skvaider 0750 skvaider service -"
       "d /var/log/skvaider 0750 skvaider service -"
       "A /var/log/skvaider - - - - g:sudo-srv:r-x,g:admins:r-x"
       "a /var/log/skvaider - - - - d:g:sudo-srv:r,d:g:admins:r"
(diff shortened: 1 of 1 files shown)


nixos/roles/ai-api-gateway.nix
```diff
@@ -149,6 +149,15 @@ in
         isSystemUser = true;
       };
     };
+    systemd.tmpfiles.rules = [
+      "d /var/lib/skvaider 0755 skvaider service -"
```
ctheune: just noticed this, why access for world?

ctheune (changes requested):

nixos/roles/ai-api-gateway.nix
elikoga: I think I copied the rules from nixos/roles/ai-model-server.nix where it used to also be 0755, but I think that was originally chosen because the model files and state of the inference system are non-sensitive and world readable for ease of debugging (?)

With debug and logs that should be more restricted then

ctheune (approved):

PR #2436, opened 2026-06-14
Experimental Wazuh agent package/service module

dpausp:
Add the Wazuh agent (v4.14.5) as experimental role:

- pkgs/wazuh/agent: Nix package for wazuh-agent, including patches for the Makefile, libbpf bootstrap, cstdint includes, and snap onError signature.
- nixos/services/wazuh: services.wazuh.agent (modulesd, logcollector, syslogcollector, authd, agent, agent-auth)
- nixos/roles/wazuh-agent: thin wrapper (flyingcircus.roles.wazuh-agent) requiring only enable, managerAddress, and authPasswordFile (default /run/wazuh-pass, only needed once for enrollment). A setup-pre-wazuh oneshot seeds /var/ossec/ from the package on first boot; enrollment is skipped once `.agent-registered` exists.
- tests/wazuh-agent: NixOS VM test with a threaded TLS fake manager implementing the OSSEC PASS/K enrollment protocol, exercising the full enrollment flow end-to-end.

PL-135426

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications


nixos/roles/wazuh-agent.nix
```diff
@@ -0,0 +1,137 @@
...
+      settings.rootcheck.skip_nfs = lib.mkDefault "yes";
+
+      settings.syscheck.disabled = lib.mkDefault "no";
+
+      settings.sca.enabled = lib.mkDefault "yes";
+      settings.sca.scan_on_start = lib.mkDefault "yes";
+      settings.sca.interval = lib.mkDefault "12h";
+      settings.sca.skip_nfs = lib.mkDefault "yes";
+
+      settings.logging.log_format = lib.mkDefault "plain";
```
zagy: That look strange. We should use sub dicts  …

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+    concatMapStringsSep
+    filterAttrsRecursive
+    types
+    ;
+
+  # Filter null values from settings to avoid empty XML elements
+  # Wazuh's XML parser doesn't handle <tag></tag> well for optional fields
+  filterNulls = filterAttrsRecursive (_: v: v != null);
+
+  xmlValue = (pkgs.formats.xml { }).type;
```
leona-ya, line 25: In the normal style you use (with the suggestion below)

```
format = pkgs.formats.xml { withHeader = false; };
```

and then in later placed reference `format.type`, `format.generate ...` etc without instatiating again

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+            };
+            xmlContent = builtins.readFile xmlWithProlog;
+            # Remove XML prolog line
+            xmlWithoutProlog = lib.removePrefix ''
+              <?xml version="1.0" encoding="utf-8"?>
+            '' xmlContent;
+          in
+          # Add Wazuh-style comment header
+          "<!--  Wazuh - Agent - NixOS generated configuration  -->${xmlWithoutProlog}"
+        );
```
leona-ya: This can be simplified without IFD (with the suggestion above):

```nix
default = format.generate "osssec.conf" {
  ossec_config = filterNulls cfg.settings;
}
```
note that this removes the comment header, but I don't see that as important, as we don't do this anywhere else and the file is RO and in the Nix store.

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+        default = { };
+        description = ''
+          Wazuh-agent configuration written in Nix. This will be serialized to XML and passed as `ossec.conf`
+
+          Note that the root `ossec_config` tag is added automatically
+
+          Not all possible configuration options are listed here - see the [config reference](https://documentation.wazuh.com/${cfg.package.version}/user-manual/reference/ossec-conf/index.html) for possible values
+        '';
+        type = types.submodule {
+          freeformType = types.attrsOf xmlValue;
```
leona-ya, line 131: The idea of the RFC42 `pkgs.formats` types is that they already represent every single valid type that is serializable:

```
let
  baseType = oneOf [
    bool
    int
    float
    str
    path
    (attrsOf valueType)
    (listOf valueType)
  ];
  valueType = (nullOr baseType) // {
      description = "XML value";
  };
in
  valueType;
```

So, here 

```suggestion
          freeformType = format.type;
```
is enough.

nixos/services/wazuh/agent.nix
leona-ya, line 115: This can be simplified without IFD (with the suggestion above):

note that this removes the comment header, but I don't see that as important, as we don't do this anywhere else and the file is RO and in the Nix store.

```suggestion
        default = format.generate "osssec.conf" {
          ossec_config = filterNulls cfg.settings;
        };
```

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+
+          options = {
+            client = {
+              server = {
+                address = mkOption {
+                  type = types.nullOr types.nonEmptyStr;
```
leona-ya, line 137: I don't think it makes sense that this is nullable, does it?

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+          Password for the auth service.
+
+          Written to `/var/ossec/etc/authd.pass` by `setup-pre-wazuh` and used
+          for initial enrollment with the manager.
+
+          Note: this value is embedded in the Nix store and the generated
+          system configuration, both of which are world-readable. Set it only
+          if that exposure is acceptable.
+        '';
+      };
```
leona-ya, line 217: Why do we need to dig the hole deeper here and not just directly only offer to enter a path to a file with the password in it?

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+
+    ];
+
+    users.users.${cfg.user} = {
+      isSystemUser = true;
+      inherit (cfg) group;
+      #TODO Determine if the agent should use a separate user than the other microservices
+      description = "Wazuh agent user";
+      home = stateDir;
+      #TODO This should be set as `SupplementaryGroups` in the systemd service config if only specific daemons need the capabilities
```
leona-ya, line 246: I agree that this would be good, can we do this now? -- I can however live with the current state if this is too much work

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+    users.groups.${cfg.group} = { };
+
+    systemd = {
+      tmpfiles.rules = [
+        "d ${stateDir}/tmp 0750 ${cfg.user} ${cfg.group} 1d"
+      ];
+      targets = {
+        multi-user.wants = [ "wazuh.target" ];
+        wazuh.wants = forEach daemons (d: "${d}.service") ++ [ "wazuh-agent-auth.service" ];
+      };
```
leona-ya, line 262: I'd like if we do not directly modify `systemd.targets.multi-user` here. Could we instead


```suggestion
      targets.wazuh = {
        wantedBy = [ "multi-user.target" ];
        wants = forEach daemons (d: "${d}.service") ++ [ "wazuh-agent-auth.service" ];
      };
```

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+
+      services = listToAttrs (map (daemon: nameValuePair daemon (mkService daemon)) daemons) // {
```
leona-ya, line 264:
```suggestion
      services = (lib.genAttrs daemons (daemon: mkService daemon)) // {
```

nixos/services/wazuh/agent.nix
```diff
@@ -0,0 +1,364 @@
...
+        };
+
+        setup-pre-wazuh = {
+          description = "Sets up wazuh's directory structure";
+          wantedBy = [ "wazuh-agent-auth.service" ];
+          before = [ "wazuh-agent-auth.service" ];
+          serviceConfig = {
+            Type = "oneshot";
+            # Run as root to create /var/ossec directory structure
+            ExecStart = "${pkgs.writeShellScriptBin "wazuh-prestart" ''
```
leona-ya, line 316: There is `pkgs.writeShellApplication` for this, this directly accepts `bashOptions`, ideally combine with `lib.getExe`.

nixos/roles/wazuh-agent.nix
```diff
@@ -0,0 +1,142 @@
...
+        `setup-pre-wazuh` simply skips creating `authd.pass`.
+
+        Once the agent has enrolled (`.agent-registered` exists), this
+        option is no longer needed — it can be left unset on subsequent
+        deployments.
+
+        Note: this value ends up in the Nix store and the generated system
+        configuration, which are world-readable.
+      '';
+    };
```
leona-ya, line 54: Here again: could we just read from a file (and then use our service to copy the value to the right place)

nixos/roles/wazuh-agent.nix
```diff
@@ -0,0 +1,142 @@
...
+        # The upstream module's freeformType (attrsOf xmlValue from
+        # pkgs.formats.xml) CONCATENATES list definitions at equal priority
+        # rather than replacing them. This means a user who writes
+        # `services.wazuh.agent.settings.syscheck.directories = ["/srv/app"]`
+        # AUTOMATICALLY EXTENDS the default `["/etc/local"]` — the result is
+        # `["/etc/local" "/srv/app"]`, not a replacement. Same for localfile
+        # and wodle (lists of attrsets → multiple XML elements).
+        #
+        # To REPLACE a default list entirely, use `lib.mkForce` on the
+        # upstream setting.
```
leona-ya, line 123: We don't need this LLM boiler plate hallucination stuff in our codebase. This is normal Nix knowledge, while also being partly wrong about how it should work (i.e. `freeformType`)

pkgs/wazuh/agent/dependencies/prefetch-external-dependencies.sh
```diff
@@ -0,0 +1,59 @@
...
+#TODO Ensure this stays in sync with dependency version from the package, mismatches can cause prefetched content to be substituted incorrectly
+DEPENDENCY_VERSION=51
+BASE_URL="https://packages.wazuh.com/deps/$DEPENDENCY_VERSION/libraries/sources"
+
+echo "{" >external-dependencies.nix
+
+for dep in "${EXTERNAL_DEPS[@]}"; do
+    HASH=$(
+        nix-prefetch-url "$BASE_URL/$dep.tar.gz" --type sha256 |
+            xargs nix hash convert --from nix32 --to sri --hash-algo sha256 --extra-experimental-features nix-command
```
leona-ya, line 46: Please use `nurl` here

pkgs/wazuh/agent/01-makefile-patch.patch
```diff
@@ -0,0 +1,96 @@
...
++	cd ${EXTERNAL_OPENSSL} && perl ./Configure $(OPENSSL_FLAGS) && ${MAKE} build_libs
+ endif
+ endif
+ endif
+@@ -1225,7 +1223,7 @@
+ 	 ${MAKE} -C $(EXTERNAL_LIBDB) libdb.a
+ 
+ $(EXTERNAL_LIBDB)Makefile:
+-	cd ${EXTERNAL_LIBDB} && CPPFLAGS=-fPIC ../dist/configure --with-cryptography=no --disable-queue --disable-heap --disable-partition --disable-mutexsupport --disable-replication --disable-verify --disable-statistics ac_cv_func_pthread_yield=no
++	cd ${EXTERNAL_LIBDB} && CPPFLAGS=-fPIC CFLAGS="-Wno-error=incompatible-pointer-types -Wno-error=old-style-definition" ../dist/configure --with-cryptography=no --disable-queue --disable-heap --disable-partition --disable-mutexsupport --disable-replication --disable-verify --disable-statistics ac_cv_func_pthread_yield=no
```
leona-ya, line 40: Urgh, could we write custom CFLAGS like this in the Nix derivation instead of some patch file?

pkgs/wazuh/agent/01-makefile-patch.patch
```diff
@@ -0,0 +1,96 @@
...
+ endif
+ 
+ #### libarchive ######
+--- a/src/shared/file_op.c	2026-05-13 00:19:17.597828305 +0200
++++ b/src/shared/file_op.c	2026-05-13 00:19:59.664897979 +0200
```
leona-ya, line 45: This file is named `makefile-patch`, but this is certainly not part of the Makefile, please move into another patch

pkgs/wazuh/agent/02-libbpf-bootstrap.patch
```diff
@@ -0,0 +1,76 @@
...
+ set(LIBBPF_INCLUDE_DIRS ${CMAKE_CURRENT_BINARY_DIR}/libbpf)
+ set(LIBBPF_LIBRARIES ${CMAKE_CURRENT_BINARY_DIR}/libbpf/libbpf.a)
++set(LIBBPF_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/libbpf/src)
+ find_package(BpfObject REQUIRED)
+ 
+ include_directories(
+   ${CMAKE_CURRENT_SOURCE_DIR}/vmlinux.h/include/${ARCH}
+ )
+ 
++set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu17 -Wno-error=implicit-function-declaration -Wno-error=int-conversion")
```
leona-ya, line 73: here again, this should likely be part of the Nix derivation instead

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+    curl
+    file # Required by configure scripts (e.g., popt)
+    perl
+    ps
+    pkg-config
+    policycoreutils
+    python312
+    python312.pkgs.setuptools
+    zlib
+    #breakpointHook
```
leona-ya, line 108: please remove

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+
+  src = fetchFromGitHub {
+    owner = "wazuh";
+    repo = "wazuh";
+    tag = "v${version}";
+    hash = "sha256-Vtld3DCp3OEFcevydZC6gZkL2ngbPsasBiyzBc5VRDY=";
+  };
+
+  dontConfigure = true;
+  #dontFixup = true;
```
leona-ya, line 88: remove

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+    };
+    modern_bpf_c = fetchurl {
+      url = "https://raw.githubusercontent.com/wazuh/wazuh/v${version}/src/syscheckd/src/ebpf/src/modern.bpf.c";
+      hash = "sha256-D7NPWwrBblP43U7DoBgZewo4wmn3HWGr14wU85+fOC8=";
+    };
+  };
+in
+stdenv.mkDerivation {
+  pname = "wazuh-agent";
+  inherit version;
```
leona-ya, line 76: We could make overriding work here better by using a `finalAttrs`-derivation and move version from the let-in block just into the derivation

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+    tar -xzf ${libbpf_bootstrap_deps.cmakeLists} -C src/external/libbpf-bootstrap --strip-components=2
+
+    echo 'grabbing modern_bpf_c...'
+    cp ${libbpf_bootstrap_deps.modern_bpf_c} src/external/libbpf-bootstrap/src/modern.bpf.c
+
+    echo 'grabbing wazuh-http-request...'
+    mkdir -p src/shared_modules/http-request
+    cp -r --preserve=timestamps --reflink=auto -- ${wazuh-http-request}/* src/shared_modules/http-request
+
+    #chmod +x src/analysisd/compiled_rules/register_rule.sh
```
leona-ya, line 161: remove?

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+
+  prePatch = ''
+    substituteInPlace src/init/wazuh-server.sh \
+      --replace-fail "cd ''${LOCAL}" ""
+
+    substituteInPlace src/external/audit-userspace/autogen.sh \
+      --replace-fail "cp INSTALL.tmp INSTALL" ""
+
+    #substituteInPlace src/external/openssl/config \
+    #  --replace-fail "/usr/bin/env" "env"
```
leona-ya, line 173: remove?

pkgs/overlay.nix
```diff
@@ -745,4 +745,8 @@ builtins.mapAttrs (_: patchPhps phpLogPermissionPatch) {
       in
       if dependencies == [ ] then python3Writer else pathWrapper;
   };
+
+  xtrabackup = lib.warn "The `xtrabackup` package has been renamed to `percona-xtrabackup`." self.percona-xtrabackup;
```
leona-ya, line 749: This seems like a wrong merge, remove

pkgs/wazuh/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The build tries to download dependencies at build time. In the Nix sandbox there's no network access, so `curl` fails — but the Makefile continues anyway due to `|| true`. We prefetch dependencies with `fetchurl` and place them where Wazuh expects them, so the failed download doesn't matter.
+
+The relevant Makefile logic lives at `src/Makefile:1503` as of Wazuh 4.14.5.
+
+This error indicates either a missing dependency or an incorrect name. With `cpython` specifically, the source tarball has a `_x86_64` suffix but the Makefile expects `cpython.tar.gz`. Fix this by adding a rename step in `default.nix` using a `let in` expression with `fetchurl`.
+
+For a missing dependency, add its name under `{url}` in `prefetch-external-dependencies.sh` and run the script.
+
+## Patches
```
leona-ya, line 45: I don't think we need this section, better move some of the comments in the patch files themself or make the patch files smaller and self-explanatory by name + comment in the derivation

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+    INSTALLDIR=$out USER_DIR=$out ./install.sh binary-install
+
+    substituteInPlace $out/bin/wazuh-control \
+      --replace-fail "cd ''${LOCAL}" "#"
+
+    chmod u+x $out/bin/* $out/active-response/bin/*
+  '';
+
+  fixupPhase = ''
+    ${getExe removeReferencesTo} \
```
leona-ya, line 219: here, `removeReferencesTo` should be part of `nativeBuildInputs` and then here just run the command.

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+    export CXX=${stdenv.cc}/bin/c++
+    make -C src TARGET=agent settings
+    make -C src TARGET=agent INSTALLDIR=$out deps
+  '';
+
+  installPhase = ''
+    mkdir -p $out/{bin,etc/shared,queue,var,wodles,logs,lib,tmp,agentless,active-response}
+
+    substituteInPlace install.sh \
+      --replace-warn "Xroot" "Xnixbld"
```
leona-ya, line 207: This could better be a `--replace-fail`

pkgs/wazuh/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```sh
+make TARGET=agent
+```
+
+## Common Build Errors
+
+If a build fails, check the errors below. These crop up when dependencies change or versions shift.
+
+### External-dependencies base URL
```
leona-ya, line 19: I'm much more interested how to choose a dependency version, as this will come up when updating.

pkgs/wazuh/README.md
```diff
@@ -0,0 +1,134 @@
...
+test -e external/cpython.tar || \
+(curl -so external/cpython.tar.gz https://packages.wazuh.com/deps/51/libraries/sources/cpython_x86_64.tar.gz && \
+cd external && gunzip cpython.tar.gz && tar -xf cpython.tar && rm cpython.tar)
+make: *** [Makefile:1503: external/cpython.tar.gz] Error 6
+make: Leaving directory '/build/src'
+```
+
+The build tries to download dependencies at build time. In the Nix sandbox there's no network access, so `curl` fails — but the Makefile continues anyway due to `|| true`. We prefetch dependencies with `fetchurl` and place them where Wazuh expects them, so the failed download doesn't matter.
+
+The relevant Makefile logic lives at `src/Makefile:1503` as of Wazuh 4.14.5.
```
leona-ya, line 39: Specifics like this will just get outdated, I don't think we have a huge benefit here.

tests/wazuh-agent.nix
```diff
@@ -0,0 +1,322 @@
...
+            description = "Fake Wazuh Manager for Testing";
+            after = [
+              "network.target"
+              "wazuh-fake-manager-certs.service"
+            ];
+            before = [ "wazuh-agent-auth.service" ];
+            wantedBy = [ "wazuh-agent-auth.service" ];
+            serviceConfig = {
+              Type = "simple";
+              ExecStart = "${pkgs.python3}/bin/python3 ${pkgs.writeText "fake-manager.py" ''
```
leona-ya, line 57: there is `pkgs.writePython3` for this.

tests/wazuh-agent.nix
```diff
@@ -0,0 +1,322 @@
...
+      for daemon in [
+          "wazuh-modulesd",
+          "wazuh-logcollector",
+          "wazuh-syscheckd",
+          "wazuh-agentd",
+          "wazuh-execd",
+      ]:
+          configured.succeed(f"systemctl is-active {daemon}.service")
+
+      # 6. Password file exists with correct owner (permissions are 750 due to upstream find -exec chmod)
```
leona-ya, line 240: What does `upstream` here mean? We (in our code introduced here) set 750, so if this is not optimal, we can and should change that!

pkgs/wazuh/agent/default.nix
```diff
@@ -0,0 +1,233 @@
...
+    chmod u+x $out/bin/* $out/active-response/bin/*
+  '';
+
+  fixupPhase = ''
+    ${getExe removeReferencesTo} \
+      -t ${libgcc.out} \
+      $out/lib/*
+
+    ${getExe patchelf} --add-rpath ${systemd}/lib $out/bin/wazuh-logcollector
+    rm -rf $out/src
```
leona-ya, line 224: `src` doesn't exist in `$out`, so no need to remove it.

leona-ya (changes requested):
Next time, I'd appreciate a deeper (human) review before submitting the PR

nixos/services/wazuh/agent.nix
zagy: Yeah, sorry. That's only temporary until we get the KV store. For testing we need something to set the password customer-wide (via directory). While we could write nix-code to write the file and then read it we would end up with a file lying around.

While this is really not nice to store the password this way, this password is *only* used once to register the agent. All further communication does not even use this password. So the impact would be registration of superfluous agents.

PR #2463, opened 2026-07-01
ceph/rgw: fix maintenance hook absolute paths

osnyx:
The `fc-ceph` command was not found when the hooks were run in scope of `fc-agent.service`, due to the systemd unit having a reduced PATH. This did not show up during manual testing, because interactive invocations of `fc-maintenance` copy the system-wide PATH.

Note: `systemctl` is already in PATH of the unit. But as I could not trace where this PATH addition originates, I've decided to be better explicit than implicit there and declare a full absolute path to the executable nonetheless.

PL-135499

Fixup for #2441

## Release process

  - changelog was missing altogehter in original PR anyways

- [x] PR has internal ticket

## Design notes

## Security implications

  - fix regression in new rgw maintenance hooks
  - identify reason why this was not caught in testing
  - ensure cleanup of stuck maintenances in staging, where this surfaced
  - this was not caught due to different PATH handling in maunal `fc-maintenance` invocations vs. `fc-agent.service`
  - manually ensured that `fc-agent.service` correctly invokes *leave* hook and cleans up locks
  - taking care of stuck staging servers has been noted down as a follow-up action

### DIFF
--- a/nixos/roles/ceph/rgw.nix
+++ b/nixos/roles/ceph/rgw.nix
@@ -145,8 +145,8 @@ in
       };
 
       flyingcircus.agent.maintenance.rgw = {
-        enter = "fc-ceph maintenance lock .radosgw && systemctl stop fc-ceph-rgw";
-        leave = "systemctl start fc-ceph-rgw && fc-ceph maintenance unlock .radosgw";
+        enter = "${cephPkgs.fc-ceph}/bin/fc-ceph maintenance lock .radosgw && ${config.systemd.package}/bin/systemctl stop fc-ceph-rgw";
+        leave = "${config.systemd.package}/bin/systemctl start fc-ceph-rgw && ${cephPkgs.fc-ceph}/bin/fc-ceph maintenance unlock .radosgw";
       };
 
       networking.firewall.extraStopCommands = ''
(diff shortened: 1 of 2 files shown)


sysvinit:
systemctl is included in the default service path, see here: https://github.com/flyingcircusio/nixpkgs/blob/7deb5f0815807d8635f37b432083bbb098faad2d/nixos/lib/systemd-lib.nix#L708-L720

nixos/roles/ceph/rgw.nix
```diff
@@ -145,8 +145,8 @@ in
       };
 
       flyingcircus.agent.maintenance.rgw = {
-        enter = "fc-ceph maintenance lock .radosgw && systemctl stop fc-ceph-rgw";
-        leave = "systemctl start fc-ceph-rgw && fc-ceph maintenance unlock .radosgw";
+        enter = "${cephPkgs.fc-ceph}/bin/fc-ceph maintenance lock .radosgw && ${config.systemd.package}/bin/systemctl stop fc-ceph-rgw";
+        leave = "${config.systemd.package}/bin/systemctl start fc-ceph-rgw && ${cephPkgs.fc-ceph}/bin/fc-ceph maintenance unlock .radosgw";
```
sysvinit, line 149: Indirecting through `config.systemd.package` isn't 100% necessary here, as systemctl will always be in the default service unit path.

sysvinit (approved):

PR #2471, opened 2026-07-07
Dynamic handling of private layer 2 networks

sysvinit:
This change introduces platform support for VMs with private, customer-specific layer 2 networks. This includes support for new ENC data, and moves handling of guest interfaces in the KVM role into fc.qemu.

PL-135250

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - Introduction of private layer 2 networks where each private network is only attached to VMs in the same resource group.
  - Private layer 2 networks currently only supported in VMs and not on hardware.
  - Guest network interface handling has largely been removed from the KVM role and is now handled by fc.qemu itself directly
  - Tested in DEV with managed NixOS and unmanaged Ubuntu/cloud-init VMs.


pkgs/fc/agent/src/fc/manage/dynamic_interface.py
```diff
@@ -0,0 +1,249 @@
...
+from fc.util.logging import init_logging
+from fc.util.typer_utils import FCTyperApp
+
+QEMU_CONFIG_DIR = Path("/etc/qemu/vm")
+QEMU_ALTNAME_PATTERN = r"fcqemu-vm-(?P<id>[0-9]+)-net-(?P<net>[0-9]+)"
+
+VXLAN_PORT = 4789
+
+app = FCTyperApp("fc-dynamic-interface")
+log = structlog.get_logger()
```
sysvinit: This currently does not have any inline logging. Logging needs to be configured in order to use the lock-handling context manager, but I'm not sure what level of inline logging in the rest of the script would be appropriate here.

pkgs/fc/agent/src/fc/manage/dynamic_interface.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+
+    net, net_id = resolve_network(interface)
+    bridge = f"br{net}"
+    vxlan = f"vx{net}"
+
+    with locked(log, context.lock_dir, "dynamic_interfaces.lock"):
+        check_call(["ip", "link", "set", interface, "nomaster"])
+
+        # XXX: delete guest iface here too?
```
sysvinit: On the KVM side, guest tap interfaces for dynamic networks are created by fc.qemu, rather than relying on qemu itself, however it's yet quite clear how deletion and cleanup is managed and whether the ifdown script needs to handle this. This might become clearer once we've actually tried live-fire testing this.

sysvinit:
Hydra tests seem to have an issue with the routed pub tests which I'm not able to reproduce on the devhost.

For some reason the guest tap device is not in the up state when fc.qemu does the route reconciliation, which causes the `ip route add` command to fail.

adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,23 @@
+# 1. Record architecture decisions
```
leona-ya, line 1: I would like to have a future discussion where/for what we want to use ADRs.

leona-ya (approved):

leona-ya:
The urgency/risk suggests that this won't make the release cycle. We decided to shorten this to meet deadlines.

I will merge as soon as dev->staging was merged.

PR #2501, opened 2026-07-28
ceph/mon: adjustments after Pacific upgrade

osnyx:
Introduces 2 useful but non-urgent changes for Ceph Pacific and more recent:

- sensu check for `rbd trash` images: Right now we do not generally use the rbd trash feature, so finding images there is unexpected and requires investigation. Does not need to be version-gated as the command already works in Nautilus.
- re-enable ceph osd autorepair at scrub time in Pacific and newer

PL-131408

## Release process

  - no, internal

- [x] PR has internal ticket

## Design notes

## Security implications

  - ensure that data supposed to be deleted is not accidentally forgotten in trash
  - the spurious autorepair display issues from PL-131662 need to be gone
  - check ran successfully on dev machine
  - no `repair` state appeared over 3 days after temporarily enabling autorepair in 3 different clusters via injectargs

### DIFF
--- a/nixos/roles/ceph/mon.nix
+++ b/nixos/roles/ceph/mon.nix
@@ -186,6 +186,13 @@ in
             package = cephPkgs.fc-ceph;
             groups = [ "sensuclient" ];
           }
+          {
+            commands = [
+              "bin/rbd trash ls"
+            ];
+            package = cephPkgs.ceph;
+            groups = [ "sensuclient" ];
+          }
         ];
 
         flyingcircus.services.sensu-client.checks = {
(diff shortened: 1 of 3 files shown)


nixos/roles/ceph/mon.nix
```diff
@@ -201,6 +208,18 @@ in
...
+              Unexpected rbd images found in trash. We do not
+              regularly use trash so far, contact the Infra team if you see a
+              need to do so.'';
+            command = ''
+              if [ $(sudo ${cephPkgs.ceph}/bin/rbd trash ls | ${pkgs.coreutils}/bin/wc -l) -ne 0 ]; then
+                exit 1
+              fi
+            '';
+            interval = 600;
+          };
```
sysvinit, line 221: It would maybe be helpful if this script outputs an info message indicating what the current state is (either all okay, or "there's rbd trash when we don't expect that, you should go take a look" in the failure case). Putting the details into the notification isn't really helpful, as I don't think we expose that anywhere other than the check details view in uchiwa (which requires a bit of a clickpath to get to). We *do* see the script output though, as that's what the monitoring review UI and the uchiwa overview pages show us next to the check name.

It might also be worth breaking this out into a `pkgs.writeShellScriptBin` rather than inlining it in the sensu-client configuration file if it gets much more involved.

sysvinit: The only people who are likely to ever see this message are the infra team anyway under current processes, so "contact the infra team" is a bit redundant. But this should do its job otherwise.

sysvinit (approved):

PR #2517, opened 2026-08-06
Fix configuration of routed PUB on routers and refactor to remove dangerous idiom

sysvinit:
The change which introduced support for dynamic private customer L2 networks in #2471 changed the format of some internal networking data structures, removing the `routed` attrset key and replacing it with `linktype`. During development, we discovered that the KVM role would use the idiom `net.routed or false` to discover routed L3 networks in order to configure forwarding correctly – however, as this field was removed and the `or` operator masked the field access, so L3 networks stopped being configured correctly. (This problem was fixed in the KVM role before merging.)

This idiom is also present in the router role, which resulted in routed L3 networks breaking in DEV once the above PR was rolled out to the routers. This change updates the router role to use the `linktype` field, and fixes the synthesised data for the loopback interface under `fclib.network` so the usages of the `or` operator which were previously necessary can be removed.

PL-135250

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - The network shouldn't silently break due to changes in internal data structures.
  - Breaking changes should fail loudly.
  - Currently running on the dev routers, VMs with pub interfaces are remotely accessible with this change applied.
  - Builds on a KVM host with no diff against the running system configuration.

### DIFF
--- a/nixos/lib/network.nix
+++ b/nixos/lib/network.nix
@@ -38,7 +38,7 @@ let
         ])
       ) encInterfaces;
       invalidLinktypesCount = length (attrNames invalidLinktypes);
-      dynamicInterfaces = lib.filterAttrs (_: value: value.linktype or null == "dynamic") encInterfaces;
+      dynamicInterfaces = lib.filterAttrs (_: value: value.linktype == "dynamic") encInterfaces;
       dynamicCount = length (attrNames dynamicInterfaces);
 
       commonConditions = [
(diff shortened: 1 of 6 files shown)


leona-ya (dismissed):

leona-ya (approved):

leona-ya:
Again: Merging now to get this into today's release as a fix for the PR merged yesterday.

PR #2532, opened 2026-08-17
frr-wait-online.service: provide a timeout

osnyx:
As a oneshot service, frr-wait-timeout.service by default had no timeout. In case of a broken state where no bgp routes appear, switch-to-configuration blocked due to this infinetly running service. Adding a timeout provides a fail-stop behaviour instead of masking the actual failure indefinitely.

PL-135151

## Release process

- [x] PR has internal ticket

## Design notes

## Security implications

  - provide a fail-stop behaviour for cases of the frr-based BGP routes not appearing
  - failure is not masked anymore, but overall system switch now fails instead of hanging indefinitley => improved visibility
  - [x] automated tests eval and pass

### DIFF
--- a/nixos/platform/network.nix
+++ b/nixos/platform/network.nix
@@ -1365,6 +1365,11 @@ in
         serviceConfig = {
           Type = "oneshot";
           RemainAfterExit = true;
+          # Timeouts necessary to prevent switch-to-configuration from blocking.
+          # Allow a very generous amount of time for routes to appear…
+          TimeoutStart = config.systemd.services.frr.startLimitIntervalSec * 2;
+          #… but if they don't, fail for good because that is an unexpected result in need of manual investigation.
+          Restart = "no";
         };
       };
     })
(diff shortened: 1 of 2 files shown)


nixos/platform/network.nix
```diff
@@ -1365,6 +1365,11 @@ in
         serviceConfig = {
           Type = "oneshot";
           RemainAfterExit = true;
+          # Timeouts necessary to prevent switch-to-configuration from blocking.
+          # Allow a very generous amount of time for routes to appear…
+          TimeoutStart = config.systemd.services.frr.startLimitIntervalSec * 2;
+          #… but if they don't, fail for good because that is an unexpected result in need of manual investigation.
+          Restart = "no";
```
sysvinit, line 1372: I would have put the timeout in the script itself, given that it's already maintaining a counter which it never checks.

`Restart=no` is also the default behaviour.

osnyx: Specifying this as a timeout at the unit level has the advantage that the reason "Timeout" is semantically persisted in the unit status and journal data, instead of specifying just *some* non-zero exit code + text message.
And making the `Restart = "no";` explicit improves readability here IMHO.

I have no problem with changing this if you insist, but the current style indeed has a reason.

sysvinit: That's all fine, I just wasn't entirely sure about the method behind the madness at first glance :)

sysvinit (approved):

PR #2534, opened 2026-08-18
sensu-client: reduce keepalive thresholds for KVM hosts

osnyx:
sensu-client sends keepalive beacons to the server each ~20s, where the lack of receiving such a beacon is translated to a `keepalive` check failure after the defined thresholds are exceeded. From observation I can say that this is only evaluated each 30s. The default thresholds are 120s (warning) and 180s (critical).

To avoid false alerts, we need to take into account how machine maintenances suppress keepalive alerts: A directory async job adds a Sensu silence for that node. That job has the following worst-case timings:
request timeout: 30s
retry period: 60s
+ some additional jobqueue delay Setting the Critical threshold to 130s allows 1 timeout processing failure.

Our alerting pipeline that processes the keepalive failure signal adds some additional delays, dampening false alerts but also delaying true alerts:
- 20s keepalive beacon message interval
- 30s evaluation interval on sensu server
- 60s sensubpi.timer interval (fc.support) + script runtime => adding up to ~110s in worst case.

Reducing the Critical threshold to 130s allows us to maintain an expected alerting delay of <5min after a KVM host crash, even when accounting for processing delays.
The Warn state of keepalive is not processed anywhere in our alerting infrastructure, so setting it low is no issue.

PL-135552

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - availability: speed up visibility of crashed KVM hosts
  - detailed overview of involved timings and considerations of selected value
  - [x] tested on a dev KVM host
  - [x] verified that changed thresholds are picked up by sensu server

### DIFF
--- a/changelog.d/20260818_135533_PL-135552-keepalive-alerting-speedup_scriv.md
+++ b/changelog.d/20260818_135533_PL-135552-keepalive-alerting-speedup_scriv.md
@@ -1,0 +1,29 @@
+<!--
+
+A new changelog entry.
+
+Delete placeholder items that do not apply. Empty sections will be removed
+automatically during release.
+
+Leave the XX.XX as is: this is a placeholder and will be automatically filled
+correctly during the release and helps when backporting over multiple platform
+branches.
+
+-->
+
+### Impact
+
+<!-- Impact means "when this change is rolled out, there
+     might be interruptions/downtimes/required actions/... that
+     IMPACT THE RUNNING APPLICATION NEGATIVELY.
+
+     Having new features or changed is not an "impact". That's what
+     the main changelog (see below) is for.
+     -->
+
+
+### NixOS XX.XX platform
+
+- KVM hosts: reduce keepalive check's critical threshold (PL-135533)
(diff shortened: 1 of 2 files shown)


sysvinit (dismissed):

sysvinit (approved):

ctheune:
Clarification question: you mention 130s threshold, but it has an 80s threshold now (4 20 second intervals) plus 30second + 60 second = 170s worst case delay for alerting. It's both below 5 minutes, but I'm just not sure where the 130 in the PR description came from.

osnyx:
It's indeed not obvious, sorry: 110s (as detailed in the commit message) + 20s as a beacon interval = 130s. This is the additional expected worst case delay throughout our alerting pipeline *after* the critical timeout of 80s has passed. Overall these 80s+130s=210s are still well below 5min.

ctheune:
Ok, so my math somehow works differently, but thanks for the explanation. It's good enough, no need to rip it apart further.

PR #2539, opened 2026-08-21
fc.agent: modify so that switch can be used without connection to Hydra

leona-ya:
PL-133327

The motivator for this change is:
  (the new) fc.nixos.util.switch can now be called without a channel object or URL. When no channel is given, the existing channel on the system disk gets used. So, a normal switch without channel bump doesn't need to connect to hydra.flyingcircus.io at all.

This has a few side effects and lead to quite a few refactors, see commit messages for details. 

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - Tested normal system switch
  - Tested system switch while hydra IPs are blackholed
  - Tested maintenance switch


leona-ya:
The refactors are awful to review, I'm sorry. The commit messages should help with a more detailed description of the changes and their reasoning.

leona-ya:
This also needs a decision whether it should be backported.

tests/router/default.nix
```diff
@@ -30,8 +30,8 @@ import ../make-test-python.nix (
         virtualisation.memorySize = 2048;
 
         imports = [
-          <fc/nixos>
-          <fc/nixos/roles>
+          ../../nixos
+          ../../nixos/roles
```
sysvinit, line 34: This is a change I've seen multiple people making in the tests recently because the imports don't resolve when they run them by hand on Hydra. Is it just me who runs the tests with `nix-build -I channels tests/...` to override the import path?

leona-ya: I typically run the tests using `nix-build release -A tests/<...>` as this is also what hydra does. It seems odd to include `<fc>`, which is inherently not defined here, as it refers to itself. Hydra however also sets `<fc>` to the PR tarball

PR #2544, opened 2026-08-24
Fix FRR reloading on 26.05

sysvinit:
systemd version 259 changed the handling of sd_notify messages sent by supervised processes, which breaks FRR's runtime reload procedure. Reloads triggered through systemd result in FRR being killed and restarted, which can cause disruptions to network connectivity. NixOS 26.05 ships with systemd version 260, which includes this behaviour. Further details are in FRRouting/frr#20430.

This change pulls in the upstream FRR patch which fixes the reload process to be compatible with the new systemd behaviour in order to avoid this issue.

PL-135549

## Release process

or

- [x] PR has internal ticket

## Design notes

## Security implications

  - Integration and compatibility – reloads should be non-disruptive and not cause connectivity outages.
  - Tested on a lab machine.

### DIFF
--- a/tests/frr.nix
+++ b/tests/frr.nix
@@ -356,6 +356,9 @@ import ./make-test-python.nix (
               host1.succeed("ip route show 192.168.42.3 | grep -F fe80::5054:ff:fe12:203")
               host1.succeed("ip route show 192.168.42.3 | grep -F fe80::5054:ff:fe12:104")
 
+          with subtest("check that config reload works correctly"):
+              # https://github.com/FRRouting/frr/issues/20430
+              host1.succeed("systemctl reload frr")
         '';
       };
       evpn = {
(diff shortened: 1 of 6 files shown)


tests/frr.nix
```diff
@@ -356,6 +356,9 @@ import ./make-test-python.nix (
               host1.succeed("ip route show 192.168.42.3 | grep -F fe80::5054:ff:fe12:203")
               host1.succeed("ip route show 192.168.42.3 | grep -F fe80::5054:ff:fe12:104")
 
+          with subtest("check that config reload works correctly"):
+              # https://github.com/FRRouting/frr/issues/20430
+              host1.succeed("systemctl reload frr")
```
osnyx, line 361: I am not certain how reliable this smoke test is. It checks that the reload command succeeds, yes, but does not check for any actual config changes in frr.
According to your commit message frr used to be killed before this change, so I guess `systemctl reload` then failed due the service main process being gone, and the restart was just a result of the `Restart=` behaviour?

Good for now, but maybe not as reliable as the subtest description may read.

osnyx (approved):

tests/frr.nix
sysvinit: Before this change `systemctl reload frr.service` would fail immediately, and then after a timeout FRR would be killed. But fair point about the message, the point is testing the interface between systemd and FRR (which is where the regression is) rather than testing that FRR actually reloads config correctly (which I assume is sufficiently tested by upstream).

PR #2557, opened 2026-09-02
Refactor platform networking configuration for upstream changes in 26.05

sysvinit:
This change refactors the network configuration code in the platform to be compatible with the changes made to scripted networking by NixOS upstream in 26.05. Upstream has removed `network-setup.service` and now uses `network.target` for coordinating network configuration. Additionally, this change reimplements the blob of shell with the jq one-liner which we used for configuring physical ethernet links in Python. There is also a fix for an upstream bug in unit ordering affecting unmounting of NFS filesystems at shutdown.

This PR should not be merged directly, as these changes need to be merged simultaneously with the Nixpkgs bump which drops our previous patch set against scripted networking. See flyingcircusio/nixpkgs#1122.

PL-135549

## Release process

  internal
or

- [x] PR has internal ticket

## Design notes

## Security implications

  - tested on router, releasetest (nfs)

### DIFF
--- a/pkgs/fc/netdev.py
+++ b/pkgs/fc/netdev.py
@@ -1,0 +1,269 @@
+import argparse
+import contextlib
+import fcntl
+import json
+import signal
+import subprocess
+import sys
+from pathlib import Path
+
+from netaddr import EUI, AddrFormatError, mac_unix_expanded
+
+INTERFACE_RENAME_LOCKFILE = "/var/lock/interface-rename.lock"
+LOCK_TIMEOUT = 60
+
+RENAME_LIMIT = 60
+INTEL_DRIVER_NAMES = ["e1000", "e1000e", "igb", "ixgbe", "i40e"]
+
+
+def fatal(msg):
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+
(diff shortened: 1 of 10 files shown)


pkgs/fc/netdev.py
```diff
@@ -0,0 +1,269 @@
...
+    sys.exit(1)
+
+
+class LockTimeoutError(Exception):
+    pass
+
+
+@contextlib.contextmanager
+def interface_renaming_lock():
+    lockfile = Path(INTERFACE_RENAME_LOCKFILE)
```
leona-ya, line 30: We could instate the `Path` object directly at the declaration

pkgs/fc/netdev.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def alarm_handler(sig, frame):
+        raise LockTimeoutError()
+
+    print("Acquiring interface renaming lock ...")
+
+    lockfile.touch()
+    with open(lockfile, "r") as f:
```
leona-ya, line 38:
```suggestion
    with lockfile.open("r") as f:
```

leona-ya (approved):
We can do this later, does not affect the functionality. I checked for behavior changes.

PR #2566, opened 2026-09-08
ceph: fix disabling the osd scrub auto repair in Nautilus

osnyx:
Properly override the ceph setting version-gated in Ceph Nautilus, that disables autorepairs during scrub. Such repairs turned out to report spurious status messages in Nautilus and remain disabled in that older release.
The override settings attr path used to be wrong, resulting in 2 settings in different settings in the config file, one true and one false.

Discovered during PL-135552, but likely unrelated.

## Release process

  - no, internal

- [x] PR has internal ticket

## Design notes

## Security implications

  - fix an incorrect configuration name attr path in a version-specific override
  - verified the override result in the ceph.conf of the NixOS test
  - verified a slightly modified variant on a live server
  - noted as a motivation for a follow-up restructuring of the Ceph module settings logic to avoid future mistakes

### DIFF
--- a/nixos/roles/ceph/osd.nix
+++ b/nixos/roles/ceph/osd.nix
@@ -293,7 +293,7 @@ in
       # we need to disable autorepair for Nautilus because it has spurious status
       # display issues of generally indicating a `repairing` for deep-scrubbing PGs.
       # see PL-131662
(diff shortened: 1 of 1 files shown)


osnyx:
Fixup for https://github.com/flyingcircusio/fc-nixos/pull/2501

sysvinit (approved):

